PR #11, opened 2016-03-05
Add summary metric type

pnieuwenhuis:
I've added a summary metric type to report to Prometheus. Added a dependency to TDigest to calculate the percentiles. Also reports count and sum of each label.

Also includes tests for the summary metric type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,13 @@ var summary = new Client.summary('metric_name', 'metric_help');
 summary.observe(10);
 ```
 
+Utility to observe request durations
+```
+var end = summary.startTimer();
+xhrRequest(function(err, res) {
+	end(); // Observes the value to xhrRequests duration in seconds
+});
+```
 
 #### Labels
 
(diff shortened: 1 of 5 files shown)


siimon:
Awesome work! Thank you!

The implementation lgtm, there are some comments in summary.js pointing to histogram that needs to be changed.. Also could you document the utility functions in the readme aswell?

Thanks!

pnieuwenhuis:
I've made some additions to the README and altered the comments in `summary.js`.

siimon:
Perfect! Thanks!

ceecko:
Nice work. I was wondering if it's possible to define a duration for which an observation stays relevant for the summary and a number of buckets for rotation. Such as 15mins over 5 buckets.
See MaxAge and AgeBuckets setting in [Go client's SummaryOpts](https://godoc.org/github.com/prometheus/client_golang/prometheus#SummaryOpts)

PR #13, opened 2016-03-24
Allow resetting histogram and summary metrics

pnieuwenhuis:
Added function to reset the values on a histogram and summary metric. Also enabled all tests again, because only the push-gateway tests were executed.

### DIFF
--- a/lib/histogram.js
+++ b/lib/histogram.js
@@ -79,6 +79,10 @@ Histogram.prototype.get = function() {
 	};
 };
 
+Histogram.prototype.reset = function() {
+	reset.call(this);
+};
+
 /**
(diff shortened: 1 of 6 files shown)


siimon:
Thanks!

What's the use case for this? I've been looking around and as far as I can see none of the other prometheus libs support this

pnieuwenhuis:
We want to use it to reset the summaries every second. Some metrics don't get much new data at the moment, so it will result in straight lines, which isn't nice/useful and confusing. It is in line with Kamon (http://kamon.io/core/metrics/instruments/), which is doing the same and is our metrics provider for our services that are written in Scala.

PR #17, opened 2016-05-09
Eliminate dependency on the object-hash package

amp001:
...and replace with custom function in lib/util.js

Improves performance of metric updates by several orders of magnitude.


siimon:
Looks good to me! 👍  I'll publish this to npm as a patch version as soon as I can

PR #19, opened 2016-06-28
Fixed bug where counter labels were not correctly handled when passed…

andrewzk:
… as a parameter to inc()


siimon:
Thanks! Merged and published to npm

PR #23, opened 2016-08-11
Added getMetricsAsJSON method

vot:


siimon:
Maybe change name to getMetricsAsJSON instead? 

And a test for this method would be great aswell!

vot:
Hey siimon, updated and renamed the method + added two very simple tests.

siimon:
Looks good! Thanks :)

PR #25, opened 2016-08-31
Add the officially recommended metrics

SimenB:
Hi there!

All Prometheus clients are recommended to export some basic metrics https://prometheus.io/docs/instrumenting/writing_clientlibs/#standard-and-runtime-collectors.

Descriptions stolen from the official Go client, https://github.com/prometheus/client_golang/blob/master/prometheus/process_collector.go.

This PR adds all of them, although most will only work on Linux.

No tests added in case you're not interested in the addition.

Example file for testing:
```js
var prom = require('./');

prom.defaultMetrics([], 500);

setInterval(function () {
    console.log(prom.register.metrics());
}, 500);
```

Example Dockerfile for testing
```dockerfile
FROM node:6

COPY . .

CMD ["node", "testfile.js"]
```

Example output of running app:

```
# HELP process_cpu_seconds_total Total user and system CPU time spent in seconds.
# TYPE process_cpu_seconds_total gauge
process_cpu_seconds_total 33.73
# HELP process_start_time_seconds Start time of the process since unix epoch in seconds.
# TYPE process_start_time_seconds gauge
process_start_time_seconds 1472657378
# HELP process_resident_memory_bytes Resident memory size in bytes.
# TYPE process_resident_memory_bytes gauge
process_resident_memory_bytes 30420000
# HELP process_virtual_memory_bytes Virtual memory size in bytes.
# TYPE process_virtual_memory_bytes gauge
process_virtual_memory_bytes 1174712000
# HELP process_heap_bytes Process heap size in bytes.
# TYPE process_heap_bytes gauge
process_heap_bytes 1131760000
# HELP process_open_fds Number of open file descriptors.
# TYPE process_open_fds gauge
process_open_fds 13
# HELP process_max_fds Maximum number of open file descriptors.
# TYPE process_max_fds gauge
process_max_fds 198652
```

Could probably add some metrics regarding stuff like GC etc.

EDIT: In hindsight, I can see that I should just check `process.platform` for `linux` instead of try-catching... I'll make that change tomorrow along with GC metrics!

EDIT2: Here is the table with recommended metrics:

| Metric name                     | Help string                                            | Unit             |
| ------------------------------- | ------------------------------------------------------ | ---------------  |
| `process_cpu_seconds_total`     | Total user and system CPU time spent in seconds.       | seconds          |
| `process_open_fds`              | Number of open file descriptors.                       | file descriptors |
| `process_max_fds`               | Maximum number of open file descriptors.               | file descriptors |
| `process_virtual_memory_bytes`  | Virtual memory size in bytes.                          | bytes            |
| `process_resident_memory_bytes` | Resident memory size in bytes.                         | bytes            |
| `process_heap_bytes`            | Process heap size in bytes.                            | bytes            |
| `process_start_time_seconds`    | Start time of the process since unix epoch in seconds. | seconds          |

### DIFF
--- a/test/defaultMetrics.js
+++ b/test/defaultMetrics.js
@@ -1,0 +1,64 @@
+'use strict';
+
+describe('defaultMetrics', function() {
+	var expect = require('chai').expect;
+	var register = require('../index').register;
+	var defaultMetrics = require('../index').defaultMetrics;
+	var platform;
+	var cpuUsage;
+	var interval;
+
+	before(function () {
+		platform = process.platform;
+		cpuUsage = process.cpuUsage;
+
+		Object.defineProperty(process, 'platform', {
+			value: 'my-bogus-platform'
+		});
+
+		if(cpuUsage) {
+			Object.defineProperty(process, 'cpuUsage', {
+				value: function () {
+					return { user: 1000, system: 10 };
+				}
+			});
+		} else {
+			process.cpuUsage = function () {
+				return { user: 1000, system: 10 };
+			};
+		}
+
+		register.clear();
+	});
+
+	after(function () {
+		Object.defineProperty(process, 'platform', {
+			value: platform
+		});
+
+		if(cpuUsage) {
+			Object.defineProperty(process, 'cpuUsage', {
+				value: cpuUsage
+			});
+		} else {
+			delete process.cpuUsage;
+		}
+	});
+
+	afterEach(function() {
+		register.clear();
+		clearInterval(interval);
+	});
+
+	it('should add metrics to the registry', function() {
+		expect(register.getMetricsAsJSON()).to.have.length(0);
+		interval = defaultMetrics();
+		expect(register.getMetricsAsJSON()).to.have.length(3);
+	});
+
+	it('should allow blacklisting unwanted metrics', function() {
+		expect(register.getMetricsAsJSON()).to.have.length(0);
+		interval = defaultMetrics(['osMemoryHeap']);
+		expect(register.getMetricsAsJSON()).to.have.length(2);
+	});
+});
--- a/README.md
+++ b/README.md
@@ -12,6 +12,61 @@ See example folder for a sample usage. The library does not bundle any web frame
 
 All metric types has 2 mandatory parameters, name and help.
 
+#### Default metrics
+
+There are some default metrics recommended by Prometheus
+[itself](https://prometheus.io/docs/instrumenting/writing_clientlibs/#standard-and-runtime-collectors). These metrics are collected
+automatically for you when you do `require('prom-client')`.
+
+NOTE: Some of the metrics, concerning File Descriptors and Memory, are only available on Linux.
+
+The function returned from `defaultMetrics` takes 2 options, a blacklist of metrics to skip, and a timeout for how often the probe should
+be fired. By default all probes are launched every 10 seconds, but this can be modified like this:
+
+```js
+var client = require('prom-client');
+
+var defaultMetrics = client.defaultMetrics;
+
+// Skip `osMemoryHeap` probe, and probe every 5th second.
+defaultMetrics(['osMemoryHeap'], 5000);
+````
+
+You can get the full list of metrics by inspecting `client.defaultMetrics.metricsList`.
+
+`defaultMetrics` returns an identification when invoked, which is a reference to the `Timer` used to keep the probes going. This can be
+passed to `clearInterval` in order to stop all probes.
+
+NOTE: Existing intervals are automatically cleared when calling `defaultMetrics`.
+
+```js
+var client = require('prom-client');
+
+var defaultMetrics = client.defaultMetrics;
+
+var interval = defaultMetrics();
+
+// ... some time later
+
+clearInterval(interval);
+````
+
+NOTE: `unref` is called on the `interval` internally, so it will not keep your node process going indefinitely if it's the only thing
+keeping it from shutting down.
+
+##### Disabling default metrics
+
+To disable collecting the default metrics, you have to call the function and pass it to `clearInterval`.
+
+```js
+var client = require('prom-client');
+
+clearInterval(client.defaultMetrics());
+
+// Clear the register
+client.register.clear();
+```
+
 #### Counter
 
 Counters go up, and reset when the process restarts.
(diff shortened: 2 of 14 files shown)


siimon:
Interesting, thanks! 

I'm away for this week and half of next so won't have a chance to dive into the code. But all of your suggestions seems reasonable and should be implemented!

A few other thoughts, 

1. Why have a defaultMetrics? Shouldn't we include them in the metrics function, with a setting if they should be included or not?

2. Is it possible to include GC metrics without adding a 3rd party dependency? If not I'm hesitant to add it

3. Maybe we should add a option object where you can disable any of the defaultMetrics?

4. Adding some documentation to the readme would be nice as well!

Did my thoughts make any sense?

SimenB:
1. That'd mean there'd be no way to disabling actually getting the metrics, which is potentially a somewhat expensive operation (as it's I/O).
2. IDK. I remember there being added GC metrics to Node back in 2013 or something, but it seems to have been removed in later iterations... I'll dig around!
3. That I can add
4. Yup, I dropped this, along with tests, in case you weren't interested in this change 😄 

One last things; at work we have our own wrapper around this library adding a few more metrics (event loop lag, OS load, process handles and process requests), that are all something node exposes. Would you be interested in adding those as well?

siimon:
actually, when thinking about this point 1 and 4 are pretty much the same thing. Would it make any sense to add a configuration object into the metrics function where you can disable/enable what metrics you want? Then it's just a matter of selecting sensible defaults.

I like your ideas, we do the same thing. I don't see any downside to this, as long as we have a good api where it's easy to enable and disable any of them. Just keep in mind that I'm trying to keep the number of dependencies as low as possible!

SimenB:
There is no "start measuring" right now, so disabling them can't be done until you first query for them, and that'd mean keeping some state around. So I still prefer having a "start polling default metrics", and this function can take options disabling them or not.

All of those extras are dependency free, just using node's own APIs 😄 

I'm back at work now, so I'll spend some time on it.

siimon:
True, sounds great! :)

SimenB:
I opened a PR with Prometheus to document this a bit better btw. If you have some comments, that'd be appreciated!

prometheus/docs#529

SimenB:
Only missing tests now.

SimenB:
@siimon This is ready to go now 😄

SimenB:
It's a bit hard to get the tests to pass when it depends on both node version and os. Ideas?

SimenB:
@siimon Latest commit makes it pass, but I'm not happy with it... Ideas?

SimenB:
I dug into GC, and it got pulled from Node here: nodejs/node#174

And open issue about adding same sort of data back is here: nodejs/node#4496

And [this](https://www.npmjs.com/package/gc-stats) userland module seems to fit the bill. I can create a project exposing this along with registering it with prom-client, which can be linked in the readme here?

siimon:
A couple of ideas,

1. Update the travis ci with more node versions, maybe one test per major version or something?

2. Not sure if node allows it, but the platform variable is probably mutable, so you could change it in your test, then put it back when your test is done. Not that pretty either but it works.

SimenB:
1. Added separate PR for it, which I'll rebase upon (#28)
2. Good idea! I've done something similar with `process.version` in one of my own projects

SimenB:
I'll try to wrap up the tests at work tomorrow, but I came over prometheus/docs#506, and it touches upon the fact that default metrics should be registered by default, but able to be turned off.

Thoughts on that? IDK if it'd be considered a breaking change, but it's just a number anyways 😄

siimon:
Yeah, the best thing would be if users of this lib doesn't have to do anything, and as you said I think we should do a major bump so we don't introduce any weird duplicate metrics for those who are doing this by themselves.

jsha:
Thanks for working on this! I am planning to make use of it as soon as it lands.

I agree these metrics should be enabled by default, and agree that that's consistent with the Prometheus docs.

SimenB:
@siimon Should be good to go now 😄 

I suggest removing the lower case exports `client.counter` etc in this major as well 😄

test/defaultMetrics.js
```diff
@@ -0,0 +1,64 @@
...
+
+	before(function () {
+		platform = process.platform;
+		cpuUsage = process.cpuUsage;
+
+		Object.defineProperty(process, 'platform', {
+			value: 'my-bogus-platform'
+		});
+
+		if(cpuUsage) {
```
SimenB: Have to support both `node@<6.1` and `node@>=6.1`

SimenB:
I'll create a separate PR for some other useful metrics, such as event loop lag and process handles/requests. Would be nice to include that as well.

@jsha @siimon Any ideas if `memoryUsage` could be used better? WRT what prometheus expects. I updated OP with the table of data they suggest.

jsha:
I think `process_heap_bytes` is intended to be the total size of the heap, which I *believe* better matches `process.memoryUsage().heapTotal`. I think it would also be useful to report `.heapUsed`, which I don't think corresponds to any of the standard Prometheus metrics.

SimenB:
I did them as they are because of the description in the original word document: https://docs.google.com/document/d/1Q0MXWdwp1mdXCzNRak6bW5LLVylVRXhdi7_21Sg15xQ/edit

jsha:
The word document says:

> Heap size: process_heap_bytes. ??? DATA as reported from ps, this comes from VmData in /proc/self/status ???

But in `man ps` I don't see anything about a DATA section. The closest thing I can see in `procps-ng version 3.3.10` is:

```
drs         DRS       data resident set size, the amount of physical memory devoted to other than executable code.
```

If I run a `node` REPL, and run `ps ao args,drs | grep node` I get:

```
node                        917122
```

Or ~9MB. From the REPL itself I get:

```
> process.memoryUsage().heapTotal
10619424
```

Not sure why Node thinks it has more heap than ps does. It could be a distinction between resident size and non-resident. At any rate, I don't have strong opinions here. I'm happy to proceed with what you've got implemented for `process_heap_bytes`.

SimenB:
Ok, thanks for the analysis 😄 Should be GTG then, shouldn't be any problem to iterate on this later if we figure out we've done it wrong

siimon:
Great :)

One thing we talked about earlier was enabling this by default, if I understood you correctly, the prom docs mentioned they should be enabled by default?

If that is the case, my suggestion is to enable them right away, and then instead have an exported function that stops collecting if you want to do that in your service.

I won't have any problems having it like that if we make a major release which I think we should

SimenB:
It does collect automatically as soon as you require `index.html`. And yup, it's recommended in the docs to collect automatically.

And I agree it should be major bump.

siimon:
Bah, thanks, i'll do a test spin and stop writing dump comments then :)

siimon:
I've no further comments, @jsha do you have any input before merging?

jsha:
LGTM

PR #28, opened 2016-09-02
Test on all supported majors of node, and OSX

SimenB:


.travis.yml
```diff
@@ -1 +1,11 @@
 language: node_js
+node_js:
+  - node
```
SimenB: Ensures latest major. So will automatically test on 7 when it releases next month. At that time, 6 should be added to the matrix 😄

PR #29, opened 2016-09-02
Add appveyor

SimenB:
Not sure if you want this, but might be a good idea to also test on windows just to avoid breakage.

You'll have to activate it if you want it, but this config file should be an ok starting point if you want it 😄

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,23 @@
+init:
+  - git config --global core.autocrlf true
+
+environment:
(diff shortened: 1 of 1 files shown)


siimon:
Definitely want this! I'll look into enabling it..

SimenB:
Great! Remember to add the badge to the readme when you get it going 😄

siimon:
Done ;)

Issue #1, opened 2015-12-02
The label handling is broken


siimon:
Labels values should be dynamically handled in each metric and expose a "labels" function as guidelines describes

PR #30, opened 2016-09-02
Add post-timer labels (#1)

notkevin888:
One common use case I've run into while using Prometheus at Tinder is having a label for success/failure on a single metric, like the following:

```javascript
var promClient = require('prom-client');
var metric = promClient.metric('metric_name', 'metric_help', ['success_response']);

var timer = metric.startTimer();
someFunction((err, result) => {
  if (err) {
    timer({ success_response: 'ERROR' });
    console.log('uh-oh');
    // do error logic
  } else {
    timer({ success_response: 'SUCCESS' });
    console.log('yay!');
    // do success logic
  }
})
```

With the original code, it seemed like there wasn't a very clean way to handle this; the only way we had solved the problem before was by creating two timers (eg, a `successTimer` and an `errorTimer`) and only ever calling one of the timers later, but that seems both prone to error and not very memory-safe.

In addition, we might also want to do something with more than just 2 options, such as tracking average response times for every HTTP code we have in an application.  In this case, the multiple-timer approach is also not very scalable, since we'd have to add an additional timer for each new code we wanted to send.

This was the approach I came up with; let me know if there's a better way of handling these kinds of use cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,18 @@ gauge.set({ method: 'GET', statusCode: '200' }, 100); // 1st version, Set value 
 gauge.labels('GET', '200').set(100); // 2nd version, Same as above
 ```
 
+It is also possible to use timers with labels, both before and after the timer is created:
+```js
+var end = startTimer({ method: 'GET' }); // Set method to GET, we don't know statusCode yet
+xhrRequest(function(err, res) {
+	if (err) {
+		end({ statusCode: '500' }); // Sets value to xhrRequest duration in seconds with statusCode 500
+	} else {
+		end({ statusCode: '200' }); // Sets value to xhrRequest duration in seconds with statusCode 200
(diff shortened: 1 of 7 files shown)


siimon:
Great, thanks! Would you mind to add some tests for this and update the readme?

I won't be able to publish to npm until the middle of next week since I'm away from a computer - hope that's ok!

notkevin888:
Yeah, for sure :)  Adding those in now!

notkevin888:
Hey, hope you had a great weekend!  I know it's still the beginning of the week; just wondering what the timeline looks like for this pr making it to npm :)

siimon:
Published with version 3.5.0! Thanks for your help!

notkevin888:
Sweet, thank you!

PR #34, opened 2016-09-05
Extra metrics

SimenB:
Builds on top of #25. I'll rebase after that's merged (or you can just merge this).
See the last commit in this PR for the diff in the meantime.

Ideas for extra metrics very much welcome!
At work we show more of `memoryUsage` (but I think those are covered in #25, although just for Linux) and `os.loadAvg[0]`, but I'm not sure if those are actually useful?

### DIFF
--- a/lib/metrics/eventLoopLag.js
+++ b/lib/metrics/eventLoopLag.js
@@ -1,0 +1,20 @@
+'use strict';
+
+var Gauge = require('../gauge');
+
+function reportEventloopLag(start, gauge){
+    var delta = process.hrtime(start);
+    var nanosec = delta[0] * 1e9 + delta[1];
+    var ms = nanosec / 1e6;
+
+    gauge.set(Math.round(ms));
+}
+
+module.exports = function() {
+    var gauge = new Gauge('node_eventloop_lag_milliseconds', 'Lag of event loop in milliseconds.');
+
+    return function() {
+        var start = process.hrtime();
+        setImmediate(reportEventloopLag, start, gauge);
+    };
+};
--- a/lib/metrics/processRequests.js
+++ b/lib/metrics/processRequests.js
@@ -1,0 +1,16 @@
+'use strict';
+
+var Gauge = require('../gauge');
+
+module.exports = function() {    // Don't do anything if the function is removed in later nodes (exists in node@6)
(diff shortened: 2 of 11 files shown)


lib/metrics/eventLoopLag.js
```diff
@@ -0,0 +1,20 @@
...
+function reportEventloopLag(start, gauge){
+    var delta = process.hrtime(start);
+    var nanosec = delta[0] * 1e9 + delta[1];
+    var ms = nanosec / 1e6;
+
+    gauge.set(Math.round(ms));
+}
+
+module.exports = function() {
+    var gauge = new Gauge('node_eventloop_lag_milliseconds', 'Lag of event loop in milliseconds.');
```
SimenB: could report this in nanoseconds instead?

siimon:
Thanks, as always, a test for this would be great - i will probably mess stuff up sooner rather than later :)

lib/metrics/processRequests.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+var Gauge = require('../gauge');
+
+module.exports = function() {
+    var gauge = new Gauge('node_active_handles_total', 'Number of active requests.');
+
+    return function() {
+        gauge.set(process._getActiveRequests().length);
```
siimon: _getActiveRequests and _getActiveHandles are new stuff to me, tried googling it and didn't find much info in node documentation. Is it a good practice to depend on "private" functions?

SimenB: It's the current recommended way. https://github.com/nodejs/node/issues/1128#issuecomment-78409928

SimenB:
@siimon Should be GTG now.

Took way more time than it should because timers didn't work. Fix for that included in its own commit.

lib/metrics/processRequests.js
SimenB: Could add an `if` guard like for `process.cpuUsage` in case it disappears in later nodes.

siimon: That would be great! Just to as you said to safe guard us for the future

siimon:
Lets just add a safeguard and I'll merge right away!

SimenB:
@siimon Done!

siimon:
Merged and published to npm with a new major version!

PR #36, opened 2016-09-19
Don't prematurely register  processMaxFileDescriptors metrics

tcolgate:


tcolgate:
Without this change, Gauge('process_max_fds', is still present when disabling all default metrics.

tcolgate:
I think I misunderstood the bug.
processMaxFileDescriptors  gets registered after it's initial run, it's not present when you attempt to list all the metrics initially to prevent them being registered.

siimon:
Thanks! :)

PR #39, opened 2016-09-20
Enfore tabs as indentation

SimenB:
Noticed that there's horribly mixed spaces and tabs (most of the spaces come from my PR, but not all).

This enforces tabs using eslint, and adds an editorconfig to help IDEs use it automatically.

The second commit upgrades to latest major of eslint. Breaking is mostly dropping support for node older than 4, which is why I included a dep that checks for the node version

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+; http://editorconfig.org
+
+root = true
+
+[*]
+indent_style = tab
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[package.json]
+indent_size = 2
+indent_style = space
+
+[*.md]
+trim_trailing_whitespace = false
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,73 +1,92 @@
 {
   "env": {
-    "node": true
-  },
-  "globals": {
+	"node": true
   },
   "rules": {
-    "quotes": [
-      2,
-      "single"
-    ],
-    "eol-last": [
-      0
-    ],
-    "no-underscore-dangle": [
-      0
-    ],
-    "no-mixed-requires": [
-      0
-    ],
-    "key-spacing": [
-      0
-    ],
-    "new-cap": [
-      0
-    ],
-    "no-console": 2,
-		"no-debugger": 2,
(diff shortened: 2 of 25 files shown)


.eslintrc
```diff
@@ -2,9 +2,11 @@
   "env": {
     "node": true
   },
-  "globals": {
-  },
   "rules": {
+    "indent": [
```
SimenB: first new

.eslintrc
```diff
@@ -45,29 +47,43 @@
...
+          "else": {
+            "after": true,
+            "before": true
+          }
+        }
+      }
+    ],
     "space-unary-ops": 2,
-    "strict": [2, "global"],
+    "space-before-function-paren": [2, "never"],
```
SimenB: second new

package.json
```diff
@@ -12,7 +12,7 @@
   },
   "scripts": {
     "test": "npm run lint && npm run test-unit",
-    "lint": "eslint --ignore-pattern doc/ --ignore-path .gitignore .",
+    "lint": "node-version-gte-4 && eslint . || node-version-lt-4",
```
SimenB: `doc/` doesn't exist, and nothing in `.gitignore` contains js files

test/.eslintrc
```diff
@@ -1,15 +1,8 @@
 {
 	"env": {
-		"node": true,
```
SimenB: these were the same as the root one, config inherit up the tree

package.json
siimon: Yeah, that's old stuff and should be safe to remove!

siimon:
👍 on this, just noticed we had mixed spaces and tabs :s

SimenB:
All changes in source code in this PR was made using `eslint . --fix`, btw

.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+root = true
+
+[*]
+indent_style = tab
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[package.json]
```
SimenB: npm rewrites this file all the time, should just keep it as spaces

PR #41, opened 2016-09-20
Throw on duplicate metrics

SimenB:
Prometheus isn't able to parse it, no reason not to fail hard

### DIFF
--- a/test/summaryTest.js
+++ b/test/summaryTest.js
@@ -2,12 +2,17 @@
 
 describe('summary', function() {
 	var Summary = require('../index').Summary;
+	var register = require('../index').register;
 	var expect = require('chai').expect;
 	var sinon = require('sinon');
 	var instance;
 
 	beforeEach(function() {
 		instance = new Summary('summary_test', 'test');
+	});
+
+	afterEach(function() {
+		register.clear();
 	});
 
 	it('should add a value to the summary', function() {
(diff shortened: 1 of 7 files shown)


siimon:
Ah, didn't know Prometheus threw an error on duplicate names - thanks.

test/summaryTest.js
```diff
@@ -63,6 +68,7 @@ describe('summary', function() {
 	});
 
 	it('should correctly use calculate other percentiles when configured', function() {
+		register.clear();
```
siimon: Do you need to clear here if it's done in the afterEach? 

Same on line 95

SimenB:
Says something along the lines of `Unknown format`, IIRC

SimenB:
https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

Can't find anything about it erroring, but it did for us...

test/summaryTest.js
SimenB: Yeah, it creates a new one with the same name as in `beforeEach`

siimon:
Yeah I tried it locally and I got an error from the scrape so you are right..

test/summaryTest.js
siimon: wouldn't it be easier to change the name in the created Summary?

siimon: never mind, I changed my mind - it's better to clear else there is a risk we will by mistake have the same name in another test.

PR #42, opened 2016-09-20
Remove single metric

SimenB:
Builds on #41 

By having a single key, we can now easily remove a metric.

If you don't want #41, I can use the array as well (`removeMetricByName`). I prefer this way, though 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,15 @@ xhrRequest(function(err, res) {
 });
 ```
 
+#### Register
+
+You can get all metrics by running `register.metrics()`, which will output a string for prometheus to consume.
+
+##### Removing metrics
+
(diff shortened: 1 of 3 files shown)


siimon:
Not sure there is a use case for this? :) When would you ever want to do that?

SimenB:
For us, it's to easily opt-out of some default metrics, without messing with `defaultMetrics`

siimon:
Alright, makes sense!

Could you just add a test for this?

SimenB:
There is a test. Do you want another one?

siimon:
Sigh, my brain must be on vacation or something.. Thats great! Thanks!

I'll merge the other PR and then push a new major version to npm

PR #45, opened 2016-09-21
Remove all default metrics on initalization (#44)

siimon:
ping @SimenB, a better solution imo, any comments?

### DIFF
--- a/lib/defaultMetrics.js
+++ b/lib/defaultMetrics.js
@@ -38,11 +38,10 @@ module.exports = function startDefaultMetrics(disabledMetrics, interval) {
 			return disabledMetrics.indexOf(metric) < 0;
 		})
 		.map(function(metric) {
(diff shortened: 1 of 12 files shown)


lib/defaultMetrics.js
```diff
@@ -37,6 +38,10 @@ module.exports = function startDefaultMetrics(disabledMetrics, interval) {
 			return disabledMetrics.indexOf(metric) < 0;
 		})
 		.map(function(metric) {
+			metrics[metric].metricNames.forEach(function(name) {
```
SimenB: `metricNames.forEach(register.removeSingleMetric)`

SimenB: `metrics[metric]` in a variable?

siimon:
Yeah the readme instructions should still be valid

PR #46, opened 2016-09-22
Remove newline from process_max_fds value

tcolgate:
The value read from the file includes a new line, this is being passed
on tot he value, and printed in theoutput. It doesn't cause a problem
for scraping right now, but it could break things in the future.


SimenB:
Why not just pass it into `Number`?

PR #52, opened 2016-10-03
Use seconds as base unit for eventloop lag

SimenB:
Ref prometheus/docs#574

### DIFF
--- a/lib/metrics/eventLoopLag.js
+++ b/lib/metrics/eventLoopLag.js
@@ -2,18 +2,18 @@
 
 var Gauge = require('../gauge');
 
-var NODEJS_EVENTLOOP_LAG = 'nodejs_eventloop_lag_milliseconds';
+var NODEJS_EVENTLOOP_LAG = 'nodejs_eventloop_lag_seconds';
 
 function reportEventloopLag(start, gauge) {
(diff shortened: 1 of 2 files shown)


voxpelli:
While this seems like a good change this is actually also a breaking change and thus should not have been released in 5.0.3 but rather as a 6.0.0 as this will cause everyone who relies on that metric to have to update their dashboards, alerts etc and also coordinate the update amongst all of their applications to get a consistent metric across all them.

I know this will cause our event lag graph in our dashboard at @sydsvenskan to break the next time we release one of our applications.

Thanks for the great module though, has really helped us jumpstart our metrics effort in our Node-apps, and sorry for not commenting and pointing this out sooner!

voxpelli:
Failed to ping @siimon in the above

SimenB:
This was a breaking change, yes. Sorry! 

@voxpelli say hi to Hugo from me 😉

siimon:
Sorry about that, completely forgot that I had merged this without bumping the version at all.. :(

SimenB:
@siimon You should revert my commit, release a patch, den reapply it before a major release 😄

siimon:
Good idea, doing it as we speak.. Sorry for the confusion

PR #59, opened 2016-11-17
Typescript improvements

spadgos:
I've just started poking around with Typescript and found there were a few problems with the definitions provided by this package. I'm _very_ new to TS definitions, so all feedback is welcomed.

A summary of the changes:

- The existing index.d.ts file isn't published with npm, so consumers of the package can't use it.
- It appears as though `[number]` was used to signify an array of numbers, however that's a tuple containing one number. `number[]` is the correct syntax.
- some method overloads were defined in a way which is apparently not good. Following the instructions in the [dos and don'ts guide](http://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html), I added optional parameters where possible, and rearranged overloads to put the more specific overload first.
- some functions didn't declare their return type. If using these definitions in a project with `noImplicitAny`, then it doesn't compile. I've added `--noImplicitAny` to the `npm run compile-typescript` command so this should be caught here too.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -389,19 +389,19 @@ export class Pushgateway {
 	 * @param params Push parameters
 	 * @param callback Callback when request is complete
 	 */
-	pushAdd(params: Pushgateway.Parameters, callback: (error?: Error, httpResponse?: any, body?: any) => void)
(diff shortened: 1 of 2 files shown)


siimon:
Thanks!

As you might noticed it's new for me aswell :) All changes looks good!

spadgos:
Thanks for the quick turnaround!

PR #62, opened 2016-12-05
Export split user+system CPU time

leonerd:
This patch adds exporting logic to export the CPU usage in split user and system counters, as well as a combined total.

At matrix.org we used to use `prometheus-client`, but I'm currently in the process of replacing it with `prom-client` instead. Because the former does not provide any process-standard metrics, I had to write the collector code myself:
https://github.com/matrix-org/matrix-appservice-bridge/blob/master/lib/components/prometheusmetrics.js#L112

This code exported CPU time split between user and system, which is a property I'd like to preserve.


siimon:
Thanks!

I don't see any reason not to expose them individually as well. I'll merge and improve the tests abit..

Issue #63, opened 2016-12-07
Counters should be created by `.inc(0)`


leonerd:
It's a common pattern in metrics instrumentation to report counters across some set of named states as labels. It's also common to initialise them all to zeroes initially, so that the first increment is noticed by prometheus and on graphs. Otherwise, prometheus' `rate()` function fails to see the jump from not-existing to 1, and hence the first occurrence of every event never appears on a graph. This can matter enormously for rare events.

For example, I have some code that performs:

    var counter = new client.Counter(..., ["state"]);
    STATES.forEach((state) => counter.inc({state: state}, 0));

With the older `prometheus-client` library I migrated away from, this would initialise every counter to zero in output. Using `prom-client` this doesn't happen, and the first occurrence goes unnoticed.

siimon:
I agree, we don't handle inc with 0's in an optimal way - working around this seems easy enough with just inc with 1 when you initialise it right? Ofc then the value of the counter won't match the number of times the event occurred but the rate will be correct.

I'll have a look on it abit later

disjunction:
+1
I was using a Counter for cron.job running daily. Current implementation misses first event when using `increase(...)` in the query

PR #64, opened 2016-12-29
Detailed nodejs v8 heap space metrics

SkeLLLa:
Hi. In this PR I've added detailed v8 heap spaces metrics. V8 has 5 main spaces (named in brackets): 
* New-space (new-space): Most objects are allocated here. New-space is small and is designed to be garbage collected very quickly, independent of other spaces.
* Old-pointer-space (old-space): Contains most objects which may have pointers to other objects. Most objects are moved here after surviving in new-space for a while.
* Old-data-space (old-space): Contains objects which just contain raw data (no pointers to other objects). Strings, boxed numbers, and arrays of unboxed doubles are moved here after surviving in new-space for a while.
* Large-object-space (large-object-space): This space contains objects which are larger than the size limits of other spaces. Each object gets its own mmap'd region of memory. Large objects are never moved by the garbage collector.
* Code-space: Code objects, which contain JITed instructions, are allocated here. This is the only space with executable memory (although Codes may be allocated in large-object-space, and those are executable, too).
* Cell-space, property-cell-space and map-space (map-space): These spaces contain Cells, PropertyCells, and Maps, respectively. Each of these spaces contains objects which are all the same size and has some constraints on what kind of objects they point to, which simplifies collection.

This metrics can be very useful for detecting memory leaks and where can they happen. For example if new-space size is growing rapidly it means that there are a lot of new data is allocated and you should probably allocate less new variables, try to reuse some of them and so on.

Also you can visually determine how much objects from new-space have survived garbage collection and were moved to old-space.

This metrics are supported starting from `>=node-6.0.0`. If older version support is needed this module can be used: https://github.com/josefzamrzla/v8-heap-space-statistics. But for now I didn't add it, but this can be easily fixed.

### DIFF
--- a/lib/defaultMetrics.js
+++ b/lib/defaultMetrics.js
@@ -9,6 +9,7 @@ var eventLoopLag = require('./metrics/eventLoopLag');
 var processHandles = require('./metrics/processHandles');
 var processRequests = require('./metrics/processRequests');
 var heapSizeAndUsed = require('./metrics/heapSizeAndUsed');
+var heapSpacesSizeAndUsed = require('./metrics/heapSpacesSizeAndUsed');
 var register = require('./register');
 
 var metrics = {
@@ -20,7 +21,8 @@ var metrics = {
 	eventLoopLag: eventLoopLag,
 	processHandles: processHandles,
 	processRequests: processRequests,
-	heapSizeAndUsed: heapSizeAndUsed
+	heapSizeAndUsed: heapSizeAndUsed,
+  heapSpacesSizeAndUsed: heapSpacesSizeAndUsed
 };
 
 var existingInterval = null;
--- a/lib/defaultMetrics.js
+++ b/lib/defaultMetrics.js
@@ -22,7 +22,7 @@ var metrics = {
 	processHandles: processHandles,
 	processRequests: processRequests,
 	heapSizeAndUsed: heapSizeAndUsed,
-  heapSpacesSizeAndUsed: heapSpacesSizeAndUsed
+	heapSpacesSizeAndUsed: heapSpacesSizeAndUsed
 };
 
 var existingInterval = null;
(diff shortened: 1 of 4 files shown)


disjunction:
Hi, @SkeLLLa 
wouldn't  it make more sense to just extend the existing `nodejs_heap_size_(total/used)_bytes` with a label specifying the space, like `{space='old'}` ?

SkeLLLa:
@disjunction may be it's good idea, but what about `available` size? Is it ok in prometheus to have metric like `nodejs_heap_size_available_bytes{space='old'}` without "parent metrics"?

SkeLLLa:
I've refactored it to use labels instead of separate metrics + added a test for it.

disjunction:
> ... without "parent metrics"?

It's very common that some metrics make sence only with the label. You can get the "parent" by just applying `sum` in the query.

> ... description. Wouldn't it make users a little bit confused?

you can add "... by space" to the description. But it makes sense for me as it is.

BTW as a developer having a dependency on `prom-client` I would be happy if I could switch on/off this functionality, despite `prom-client` being a singleton. But it's just my personal opinion.

SkeLLLa:
@disjunction As for disabling you can use method described here: https://github.com/siimon/prom-client#default-metrics. I think it will work for this metrics as well.

siimon:
Thanks! This LGTM!

SkeLLLa:
@siimon, great. As I've wrote before if someone will need support of this metrics on node, I can add `v8-heap-space-statistics` module. I don't add it because it needs compilation and it's only needed for node version 4 and 5.
So if you know some elegant way to add this dependency without affecting node >=6.0.0 users (may be as peer dependency or with some postinstall script trick that will be `node-version-lt-6 && npm i v8-heap-space-statistics`), I can add it.

siimon:
Oh sorry, forgot to comment that,

I rather just leave it as it is, if you want detailed heap statistics you need to run node >= 6.0.0. It's not worth adding a dependency that's not needed in a newer version imo. 

Just fyi, I'll try to fix #63 before publishing to npm - will be a major bump so I want to get everything in there.. Hope that is ok!

SkeLLLa:
@siimon, ok, no problem.

PR #65, opened 2017-01-02
Passing authentication data through URI, when pushing to Pushgateway

cfiet:
Hi,

I've done a small adjustment in `prom-client`, that you might be interested in merging into main branch.

In my setup, I run Pushgateway through proxy that requires basic client authentication, so I've modified the way requests are being made, to provide auth configuration through Pushgateway URL.

It's just a basic support for the URL format: `https://<username>:<password>@example.org`

Actual implementation just uses the standard NodeJS `url` module to construct HTTP request parameters that are not stripped of username and password parts of URL provided to Pushgateway object constructor.

### DIFF
--- a/lib/pushgateway.js
+++ b/lib/pushgateway.js
@@ -57,8 +57,7 @@ function useGateway(method, job, groupings, callback) {
 	});
 
 	if(method !== 'DELETE') {
-		var content = register.metrics();
-		req.write(content);
+		req.write(register.metrics());
 	}
 	req.end();
 }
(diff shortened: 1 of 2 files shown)


siimon:
Thanks, a test for this would be great!

I've pushed a commit in master where I've removed support for legacy node versions - so rebase your branch and hopefully the build should pass.

cfiet:
There are now some tests checking, if Authentication header is attached to outgoing requests, if it was provided in URL.

BTW, thanks for making this library, it's super useful!

siimon:
Thanks! :)

PR #66, opened 2017-01-03
Fix jsdoc type names

SkeLLLa:
Hello again. It's a little fix for type names in jsdoc to correct code completion in some IDEs.

I prefer format: `Array.<string>`, but `string[]` is also correct, so you can choose what do you like :).

Also I've noticed that `Number` type in jsdoc is starting from capital letter. So may be it will be better sometime to adjust all jsdocs to one style (e.g. all types starting from capital letter or just Arrays, Functions, Symbosl, etc.). Also there is nice plugin [jsdoc-to-markdown](https://github.com/jsdoc2md/jsdoc-to-markdown) that can generate quiet good docs from code (example [here](https://github.com/75lb/handbrake-js#api-reference)). I can make another PR later if you want.

PS: I don't know why another commits from previous PR added here. Tried to fix this but without success. Just see Files chaned section for real diff.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,8 @@
 language: node_js
 node_js:
   - node
-  - 5
+  - 6
   - 4
-  - 0.12
-  - '0.10'
 os:
   - linux
   - osx
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,11 +3,9 @@ init:
 
 environment:
   matrix:
-  - nodejs_version: "0.10"
-  - nodejs_version: "0.12"
   - nodejs_version: 4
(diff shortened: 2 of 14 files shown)


siimon:
Thanks!

The change looks good to me, but just to be safe, please create a new branch from the current master repo. Would be bad if it rewrote history!

SkeLLLa:
@siimon NP! I'll remove my fork and recreate it from scratch. May be in this case it will not include all older commits into new PR.

SkeLLLa:
@siimon Here's PR with clean commit history: https://github.com/siimon/prom-client/pull/68.

Closing this one as non relevant.

PR #68, opened 2017-01-11
Fix array jsdoc and remove unused variables

SkeLLLa:
This is PR https://github.com/siimon/prom-client/pull/66 with clean history.

Also I've removed unused `createValue` variable from some files (it also probably should look like `var createValue = require('./util').setValue;`, but in any case it's not used in code, so I think removing it is a good idea).


siimon:
Thanks! :)

PR #69, opened 2017-01-26
Add timestamp field support

rstuven:
According to exposition format specification https://prometheus.io/docs/instrumenting/exposition_formats/

### DIFF
--- a/lib/counter.js
+++ b/lib/counter.js
@@ -5,6 +5,7 @@
 var register = require('./register');
 var type = 'counter';
 var isNumber = require('./util').isNumber;
+var isDate = require('./util').isDate;
 var getProperties = require('./util').getPropertiesFromObj;
 var hashObject = require('./util').hashObject;
 var validateLabels = require('./validation').validateLabel;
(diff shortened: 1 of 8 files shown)


rstuven:
TODO: Update docs

lib/counter.js
```diff
@@ -47,15 +48,16 @@ function Counter(name, help, labels) {
...
  * @param {Number} value - Value to increment, if omitted increment with 1
+ * @param {(Number|Date)} timestamp - Timestamp to set the counter to
  * @returns {void}
  */
-Counter.prototype.inc = function(labels, value) {
+Counter.prototype.inc = function(labels, value, timestamp) {
 	if(isNumber(labels) || !labels) {
-		return inc.call(this, null)(labels);
+		return inc.call(this, null)(labels, value);
 	}
```
siimon: Not sure why value is added here

lib/gauge.js
```diff
@@ -49,34 +50,37 @@ function Gauge(name, help, labels) {
...
  * @param {object} labels - Object with labels and their values
  * @param {Number} value - Value to set the gauge to, must be positive
+ * @param {(Number|Date)} timestamp - Timestamp to set the gauge to
  * @returns {void}
  */
-Gauge.prototype.set = function(labels, value) {
+Gauge.prototype.set = function(labels, value, timestamp) {
 	if(isNumber(labels)) {
-		return set.call(this, null)(labels);
+		return set.call(this, null)(labels, value);
```
siimon: Same here, I'm probably missing something :)

lib/counter.js
rstuven: If labels is a number, then the next argument (value) actually is a timestamp.

lib/gauge.js
rstuven: Same as above :)

rstuven:
> Have you verified this so Prometheus is able to scrape metrics with time stamps?

Yes, already verified.

lib/counter.js
siimon: Gotcha!

Might rename the variable then so it makes more sense, I can do that later on if you don't want to do it..

siimon:
Another thing, we need to update the typescript definitions as well, as in the other comment, if you don't want to do it I can do it after this PR is merged

SimenB:
Any news here? 😄

PR #70, opened 2017-02-05
Use hrtime for timing

smithamax:
Using `new Date()` for timing is susceptible to clock reset issues via
ntp as well as leap seconds, resulting in incorrect or even negative
values. `process.hrtime()` is the recommended method of timing things
in node because it uses monotonic time and is not susceptible to issues
around wall clock resets. It also has the added benefit of being higher
resolution, although that wont really effect histograms/summaries much.

https://nodejs.org/api/process.html#process_process_hrtime_time

I had to use the pre release of sinon 2.x to get the tests to work for
this change, sinon 1.x does not support fake hrtime. As this is
just a dev dependency, I hope this is alright.

### DIFF
--- a/lib/gauge.js
+++ b/lib/gauge.js
@@ -137,10 +137,10 @@ function setToCurrentTime(labels) {
 function startTimer(startLabels) {
 	var gauge = this;
 	return function() {
-		var start = new Date();
+		var start = process.hrtime();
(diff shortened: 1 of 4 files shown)


siimon:
Thanks, looks good to me! :)

PR #78, opened 2017-03-09
Added explanation of how to include labels _and_ configuration for hi…

goofballLogic:
…stograms


siimon:
Thanks! Summary is missing the same documentation. But I'll merge this one and then we can update the summary in another pr

PR #86, opened 2017-03-30
Fix shadowing of nodejs_heap_size_*_bytes by spaces size

SkeLLLa:
I've found out that heap space metrics with labels can shadow simple heap stats. So I think it's better to move them to other "namespace".

### DIFF
--- a/lib/metrics/heapSpacesSizeAndUsed.js
+++ b/lib/metrics/heapSpacesSizeAndUsed.js
@@ -19,7 +19,7 @@ var METRICS = [
 var NODEJS_HEAP_SIZE = {};
 
 METRICS.forEach(function(metricType) {
-	NODEJS_HEAP_SIZE[metricType] = 'nodejs_heap_size_' + metricType + '_bytes';
+	NODEJS_HEAP_SIZE[metricType] = 'nodejs_heap_space_size_' + metricType + '_bytes';
 });
 
 


siimon:
Not sure I follow you, can you give an example?

Renaming metrics like this is annoying since it needs to be a major release

SkeLLLa:
@siimon, probably I'm mistaken, but plain heap usage stats and heap usage stats separated by `space` label have the same name:  `nodejs_heap_size_total_bytes`. So when I'm calling `metrics.register.metrics();` I only see metrics with `space` labels, but not the general one, they are just missing.

As a workaround without changing metrics name I can suggest adding `space` label to usual heap metrics, but I don't know what should be the correct value for that, probably 'all'. Or another option is just to make users choose what do they need: either detailed heap space metrics (and sum all those space sizes to get "total") or general heap metrics.

siimon:
Gotcha! I think your PR is the best way to solve this, but I'm not sure it justifies a major release. Or what do you guys feel? Are major versions bumps annoying or should we just go forward?

SimenB:
This is a bugfix, IMO

SimenB:
Wait what, it shouldn't be possible to shadow, the client should throw.

SimenB:
See #92

SkeLLLa:
@SimenB to say the truth I think that generally it will be probably better not to throw exception in constructor, but return existing metric.

SimenB:
It should throw on duplicate metrics, not silently drop either the old (current behavior) or the new (your suggestion) metrics

siimon:
I agree on throws! The lib should do what you think does and not trying to be smart

siimon:
Merging this and #92, will be a major release tomorrow when I got access to a computer. Will also try and get #69 in that release

PR #90, opened 2017-03-31
Fix getSingleMetric type and add exported Metric type

SkeLLLa:
I'm sorry, yesterday I've made a PR with "outdated" typing fix (forgot to push correct fix into my fork). Obviously `getSingleMetric` function returns one of Counter, Gauge, Summary or Histogram. For those reasons I've also added export type called `Metric` which corresponds to one of that types. It is very handy when you're writing something in typescript.


siimon:
Ah, I missed that as well! Thanks!

PR #92, opened 2017-04-01
Throw if there are duplicate default metric names

SimenB:
See #86 

This currently errors out, because of what #86 is trying to fix. So merging first that then rebasing this should make it green.

I never considered the case where the included metrics could clash when adding the throw...
This change makes the first run add all, and only on subsequent invocations of the function will it clean out the old metric

### DIFF
--- a/lib/defaultMetrics.js
+++ b/lib/defaultMetrics.js
@@ -28,6 +28,9 @@ var metrics = {
 };
 
 var existingInterval = null;
+// This is used to ensure the program throws on duplicate metrics during first run
+// We might want to consider not supporting running the default metrics function more than once
+var init = true;
 
 module.exports = function startDefaultMetrics(disabledMetrics, interval) {
 	if(existingInterval !== null) {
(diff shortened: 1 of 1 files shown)


lib/defaultMetrics.js
```diff
@@ -28,6 +28,9 @@ var metrics = {
 };
 
 var existingInterval = null;
+// This is used to ensure the program throws on duplicate metrics during first run
+// We might want to consider not supporting running the default metrics function more than once
```
SimenB: https://github.com/prometheus/client_java/blob/master/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/DefaultExports.java

Java noops. I don't see the use case we want to support (blacklisting certain metrics) as valid. Also, we should probably move collecting the metrics behind and explicit method call. In which case we can keep blacklisting, although I don't think it's necessary.

SimenB: That's a (bigger) breaking change, though. But makes it more in line with the official clients

siimon:
I'd say keep it as it is until we find a good reason to break the API like that.

I'll merge both the PR's

SimenB:
Rebased this 😄

PR #97, opened 2017-04-19
Displaying a single metric as a Prometheus string

serge1peshcoff:
I've recently faced an issue: I needed to display 2 different metrics on 2 different endpoints (they would be fetched with two different intervals by Prometheus). `prom-client` doesn't provide a way to get a single metric as a Prometheus string. So I've implemented it.

I changed a `register.getSingleMetric()` to return a Prometheus string, added `register.getSingleMetricObject()` to get metric by name (instead of `register.getSingleMetric()`), provided an example of these changes and README entry for them and fixed tests to reflect these changes.

There would be 1 breaking change (the `register.getSingleMetric()` would return a string instead of an object), I think my naming is better (if you have other ideas how to name methods, please tell).

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,11 +203,11 @@ You can get all metrics by running `register.metrics()`, which will output a str
 
 ##### Geting a single metric for Prometheus displaying
 
-If you need to output a single metric for Prometheus, you can use `register.getSingleMetric(*name of metric*)`, it will output a string for Prometheus to consume.
+If you need to output a single metric for Prometheus, you can use `register.getSingleMetricAsString(*name of metric*)`, it will output a string for Prometheus to consume.
 
-##### Getting a single metric object
+##### Getting a single metric
 
-If you need to get a reference to a previously registered metric, you can use `register.getSingleMetricObject(*name of metric*)`.
+If you need to get a reference to a previously registered metric, you can use `register.getSingleMetric(*name of metric*)`.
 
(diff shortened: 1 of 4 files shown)


siimon:
I'd rather not make a breaking change for this, could probably be avoided with another name for that function.

What about getSingleMetricAsString?

serge1peshcoff:
I thought about naming it the other way, but for me `getSingleMetricAsString` sounds too complicated. And also `getSingleMetric` is ambigious anyway, it's unclear what this function returns, a string or an object.
Not sure about this one.

siimon:
Yeah, the naming might not be great, but not that bad imo that it's worth a breaking change.

serge1peshcoff:
I can rename it, no problem, you can refactor it later in a major upgrade if you'll want to

serge1peshcoff:
@siimon check the new commits, they shouldn't include breaking changes now

siimon:
Thanks :)

SkeLLLa:
@siimon we should update [typedef ](https://github.com/siimon/prom-client/blob/master/index.d.ts) aswell.

serge1peshcoff:
@SkeLLLa whoops, I didn't know about this one. I can update it tomorrow if you want

SkeLLLa:
@serge1peshcoff it would be great.

Issue #38, opened 2016-09-20
Optimise metric functions when used with the labels function


siimon:
There is a good possibility to improve the performance if you're using the labels function - instead of calculating the labels hash everytime we do a metric operation it could be done when you create the labels function. This enables the user of the library to cache the metric with the specific label and you won't have to hash the label object everytime.

This is already done in Counters, but should be done in all other metric types as well

jdmarshall:
The new storage code includes some of these observations.

PR #99, opened 2017-04-22
Add file with disabled default metrics

SkeLLLa:
I've added a file that contains all exports from `index.js` excluding default metrics. It will be useful in modules like `node-prometheus-gc-stats` when those metrics aren't needed. I think that enabling collection of default metrics upon require is not good for using client library in some custom modules, because it will create useless metrics in registry.

Probably later it will be better to allow users manually trigger start of default metrics (including creation of proper gauges, counters, etc. in register) by calling some method. But it will be definitely breaking change, so for now this PR should give an option to users easily use client without default metrics.

Also I can modify `index.js` and add require of `noDefaultMetrics` to avoid copy-paste. WDYT?

### DIFF
--- a/noDefaultMetrics.js
+++ b/noDefaultMetrics.js
@@ -1,0 +1,18 @@
+/**
+ * Prometheus client
+ * @module Prometheus client
+ */
+
+'use strict';
+
+exports.register = require('./lib/register');
+exports.contentType = require('./lib/register').contentType;
+
+exports.Counter = require('./lib/counter');
+exports.Gauge = require('./lib/gauge');
+exports.Histogram = require('./lib/histogram');
+exports.Summary = require('./lib/summary');
+exports.Pushgateway = require('./lib/pushgateway');
+
+exports.linearBuckets = require('./lib/bucketGenerators').linearBuckets;
+exports.exponentialBuckets = require('./lib/bucketGenerators').exponentialBuckets;
--- a/README.md
+++ b/README.md
@@ -70,6 +70,12 @@ clearInterval(client.defaultMetrics());
 client.register.clear();
 ```
 
+Also you can require file `noDefaultMetrics` instead. It doesn't contain default metrics:
+
+```js
+var client = require('prom-client/noDefaultMetrics');
+```
+
 #### Counter
 
 Counters go up, and reset when the process restarts.
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
   "files": [
     "lib/",
     "index.js",
-    "index.d.ts"
+    "index.d.ts",
+    "noDefaultMetrics.js"
   ],
   "directories": {
     "test": "test"


siimon:
Not sure I agree with you :)

I've not used the gc-stats module myself so I'm not sure why they are unnecessary?

Using prom-client inside a module is really not recommended, how do you handle multiple modules that requires prom-client since it heavily relies on as it should be used as a singleton? To collect all metrics you would need to merge different registers and that would be a mess.

I'm actually a bit curious on how other solves this issue, exposing metrics from modules?

SkeLLLa:
@siimon I use prom client in couple internal modules and it's fine, but I need to write some code like this:
```js
var client = require('prom-client');
clearInterval(client.defaultMetrics());
```
to avoid unnecessary stats collection. But even if I do this, default metrics are created in register and than they are exported in `client.register.metrics()`, even if they aren't needed. So to avoid this I used to either delete all those default metrics manually or get needed metrics via `getSingleMetric()` and concatenate result strings.

Also as you can see in [prometheus-gc-stats](https://github.com/SimenB/node-prometheus-gc-stats/blob/master/gc-metrics.js#L3) only one counter:
```js
require('prom-client/lib/counter')
```
and it's directly required. But in any case it's metrics will be shown in registry of main app because of nodejs module caching.

As for singleton - nodejs by default handles (module caching) this unless you have the same versions of prom-client. So if you have the same version of prom-client across your app and modules, you'll have the same registry and there will be no need to merge something. Another corner case is npm linking - if you have an app module that requires `prom-client` and main app also requires it, but you're linking the app module to your app it will have different registries, because path of those two `prom-client`s will be different.

SkeLLLa:
To make it clear you can look at this [example](https://gitlab.com/prometheus-nodejs/prometheus-plugin-gc-stats/blob/master/lib/index.js#L4). It's slightly rewritten https://github.com/simenb/node-prometheus-gc-stats module with some extra features. As you can see I should require each metric file separately in order to avoid creation of default metrics.

```js
const client = {
  Counter: require('prom-client/lib/counter'),
  Gauge: require('prom-client/lib/gauge'),
  Histogram: require('prom-client/lib/histogram'),
  Summary: require('prom-client/lib/summary'),
  register: require('prom-client/lib/register'),
};
```

So I thought that it would be great if prom-client could also provide ability to users to use it without default metrics.

voxpelli:
@SkeLLLa I'm +1 on what @siimon says. One should not assume that `node_modules` will have deduped the same version of a module. It isn't required and eg. some systems with lock files (like probably Yarn) can end up with situations where they are not – and that's perfectly valid.

If it's critical that many modules uses the same instance of a dependency, as is the case with this module, then it should be a `peerDependency` of those modules. Only that way one will ensure that all modules will make use of the same instance.

SkeLLLa:
@voxpelli this PR is not about peer dependencies and deduplication. It's about giving ability to users import/require prom-client without activating defaultMetrics collection.

voxpelli:
@SkeLLLa Yes, with the argument of:

> I think that enabling collection of default metrics upon require is not good for using client library in some custom modules

And the follow up reasoning of:

> So if you have the same version of prom-client across your app and modules, you'll have the same registry and there will be no need to merge something.

Which is wrong.

If I should weigh in on the topic of the PR itself rather than just the technicalities brought up to justify it, then I agree that I think it would be preferable to have default metrics be actively toggled – either through something like `client.addDefaultMetrics()` (non-backwards compatible) or through your suggested mechanic (which would be backwards compatible)

noDefaultMetrics.js
```diff
@@ -0,0 +1,18 @@
...
+exports.contentType = require('./lib/register').contentType;
+
+exports.Counter = require('./lib/counter');
+exports.Gauge = require('./lib/gauge');
+exports.Histogram = require('./lib/histogram');
+exports.Summary = require('./lib/summary');
+exports.Pushgateway = require('./lib/pushgateway');
+
+exports.linearBuckets = require('./lib/bucketGenerators').linearBuckets;
+exports.exponentialBuckets = require('./lib/bucketGenerators').exponentialBuckets;
```
voxpelli: This isn't very DRY – I would suggest updating `index.js` as well and make `index.js` be the exported stuff from this module, but with defaultMetrics added on top of it

siimon:
But my point is bigger than that - i don't think using prom-client inside a module is a good idea because of the issues mentioned earlier. Since prom-client is designed to be used from the service itself, it also makes sense to enable collection of metrics by default. 

What stops you from having it as a peerDependency in your internal apps? Just like gc-metrics is doing?

noDefaultMetrics.js
SkeLLLa: @voxpelli yeah, I've wrote it in first comment as an option, but it will be breaking change. As for me both options will be ok :)

SkeLLLa:
@siimon I use it as peer dependency, but even if it's used like that npm linking could break it, but it's rare case :). As for prom-client in npm modules it's tricky question, because if you have the same metrics like gc or, you can set proxy handler on `net` lib to track down number of connections, requests and so on. And this could be needed not only in number of your apps, so making separate module is good idea.

Also some of my modules could be used like a module (in monolithic apps, for example), or as a separate service. In such case to avoid a lot of `if` statements and to store all logic in one place it is much easier to have prom-client in module and main app. When it is running like a module it just pushes stats to main app prom register (using peer dependency). When it is running like a service it just adds http endpoint for Prometheus.

SkeLLLa:
@voxpelli 

> something like client.addDefaultMetrics() (non-backwards compatible)

I like this option more, because it will allow to solve https://github.com/siimon/prom-client/issues/38 in easier way, because we'll have separate function and we pass some label arguments to default metrics.

SimenB:
Huge +1 to an explicit call for default metrics. That's what the Java client does. I mentioned the possibility here: https://github.com/siimon/prom-client/pull/92#discussion_r109308159

siimon:
Alright, that's settled then! Closing this in favor of #101

SkeLLLa:
@siimon great!

PR #100, opened 2017-04-27
Non-global registries

eljasala:
Currently the library registers all metrics in the global registry without a way to override this behavior & there isn't support for other registries. There are couple of downsides with this that we came across (there might be more):
* It's problematic for modules to use this library in the current state. This + #99 should make it easier for modules to expose metrics to application and application could then decide which ones it wants to expose
* If you want to expose different set of metrics on different endpoints, you would need to hardcode the metrics you want to expose on both endpoints.

I do realize there is still at least one problem left with the modules using library directly: When merging different registries, it's possible that there are clashes on names and there isn't currently a documented interface for changing the name. `metric.name` is public, but since it's not really documented, I would be a bit hesitant to use it & it might cause other issues as the registry uses it internally.

This PR adds support for `Registry`, converts current `register` to it and makes registration to global registry optional. All changes should be backwards-compatible as far as I know.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,7 +4,7 @@
 /**
  * Container for all registered metrics
  */
-export interface Registry {
+export class Registry {
 	/**
 	 * Get string representation for all metrics
 	 */
@@ -63,7 +63,8 @@ interface labelValues {
 export interface CounterConfiguration {
 	name: string,
 	help: string,
-	labels: string []
+	labels: string [],
+	register: Registry[]
 }
 
 /**
@@ -80,9 +81,8 @@ export class Counter {
 	 * @param name The name of the metric
 	 * @param help Help description
 	 * @param labels Label keys
-	 * @param register Register to default registry
 	 */
-	constructor(name: string, help: string, labels?: string[], register?: boolean)
+	constructor(name: string, help: string, labels?: string[])
 
 	/**
 	 * Increment for given labels
@@ -123,7 +123,8 @@ export namespace Counter {
 export interface GaugeConfiguration{
 	name: string,
 	help: string,
-	labels: string[]
+	labels: string[],
+	register: Registry[]
 }
 
 /**
@@ -140,9 +141,8 @@ export class Gauge {
 	 * @param name The name of the metric
 	 * @param help Help description
 	 * @param labels Label keys
-	 * @param register Register to default registry
 	 */
-	constructor(name: string, help: string, labels?: string[], register?: boolean)
+	constructor(name: string, help: string, labels?: string[])
 
 	/**
 	 * Increment gauge for given labels
@@ -250,7 +250,8 @@ export interface HistogramConfiguration {
 	name: string,
 	help: string,
 	labels: string[],
-	buckets: number[]
+	buckets: number[],
+	register: Registry[]
 }
 
 /**
@@ -268,9 +269,8 @@ export class Histogram {
 	 * @param help Help description
 	 * @param labels Label keys
 	 * @param config Configuration object for Histograms
-	 * @param register Register to default registry
 	 */
-	constructor(name: string, help: string, labels?: string[], config?: Histogram.Config, register?: boolean)
+	constructor(name: string, help: string, labels?: string[], config?: Histogram.Config)
 	/**
 	 * @param name The name of metric
 	 * @param help Help description
@@ -334,7 +334,8 @@ export interface SummaryConfiguration{
 	name: string,
 	help: string,
 	labels: string[]
-	percentiles: number[]
+	percentiles: number[],
+	register: Registry[]
 }
 
 /**
@@ -352,9 +353,8 @@ export class Summary {
 	 * @param help Help description
 	 * @param labels Label keys
 	 * @param config Configuration object
-	 * @param register Register to default registry
 	 */
-	constructor(name: string, help: string, labels?: string[], config?: Summary.Config, register?: boolean)
+	constructor(name: string, help: string, labels?: string[], config?: Summary.Config)
 	/**
 	 * @param name The name of the metric
 	 * @param help Help description
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,7 +4,7 @@
 /**
  * Container for all registered metrics
  */
-export interface register {
+export interface Registry {
 	/**
 	 * Get string representation for all metrics
 	 */
@@ -37,7 +37,7 @@ export interface register {
 /**
  * The register that contains all metrics
  */
-export const register: register
+export const register: Registry
 
 /**
 * General metric type
@@ -68,8 +68,9 @@ export class Counter {
 	 * @param name The name of the metric
 	 * @param help Help description
 	 * @param labels Label keys
+	 * @param register Register to default registry
 	 */
-	constructor(name: string, help: string, labels?: string[])
+	constructor(name: string, help: string, labels?: string[], register?: boolean)
 
 	/**
 	 * Increment for given labels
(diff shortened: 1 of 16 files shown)


index.d.ts
```diff
@@ -4,7 +4,7 @@
 /**
  * Container for all registered metrics
  */
-export interface register {
+export interface Registry {
```
SkeLLLa: In `index.js` `Registry` is added as extra property:
```js 
exports.Registry = require('./lib/registry');
```
However in this file old `register` is removed and replaced with `Registry`.

eljasala: That part should be fine, but I guess this should actually be `export class Registry`. The original `register` constant that's still in index can still be found from line 40, but the type has been changed to this new class.

SkeLLLa: Ok, thanks. Probably later I'll have some time to check it with some projects in written typescript to be sure that all is fine.

siimon:
Thanks! :)

A few comments, 

1. Wouldn't it be neater to send in a registry to the metric instead of a boolean? Then the users of the module doesn't need to register each metric. Why would you even create a metric and not add it to a register?

2. Not really a comment for this PR, but it's starting to get really messy with all the input parameters in each metrics constructor. What do you guys think about creating an object for this instead? Would be alot easier to add new functionality.. 

3. Could you do something about the naming of the variables in each metric? Registry vs register gets tossed around pretty much and is confusing.

I'll try to add some more code review comments in a bit

lib/counter.js
```diff
@@ -35,13 +36,21 @@ function Counter(name, help, labels) {
 	if(!validateLabelNames(labels)) {
 		throw new Error('Invalid label name');
 	}
+
+	if(register === undefined) {
+		register = true;
```
siimon: I don't like that we mutate the register property here, create a new variable inside the scope instead

lib/pushgateway.js
```diff
@@ -5,7 +5,11 @@ var http = require('http');
 var https = require('https');
 var register = require('./register');
 
-function Pushgateway(gatewayUrl, options) {
+function Pushgateway(gatewayUrl, options, registry) {
+	if(!registry) {
+		registry = register;
```
siimon: Same here, don't mutate input variables to the function - will bite us in the end

eljasala:
1. That was my original plan, but decided against it for couple of reasons:

* If user wants to register metric to two different registries, they would pass one of them as the parameter and then manually register it to the other one. This could be fixed if the parameter was an array, but it becomes a bit messy.
* When used in a library, it's likely better to expose the raw metric instances and let user decide where they want to register them. Otherwise it gets somewhat complicated to expose them on different endpoints.
* The only time metrics use registry is when they register themselves. To me that, that feels like a bit unnecessary side-effect. If we let user handle it, metrics are no longer dependent on any particular registry implementation and the dependency becomes one-way (registry requires metrics to implement certain interface, but that no longer applies the other way around).

2. That would likely be better. Would the current new parameter then be it or should we overload the first one?

3. Any suggestions? `globalRegistry` for the "register" instance and `registryInstance` for currently referred one?

As for comments, I can fix them, though we probably should decide on the 2nd point first (and if we will do that on this PR or afterwards).

siimon:
Responses: 

1. I see your points, but if want to optimize for the most common use case and make the library as easy to use as possible. I think providing registries as an array would work? But another suggestion would be if we do point 2, there could be a registry property and a don't add to register property. 

2. I think we need to have both versions running at some time, to make life easier for everyone using the lib to migrate. But maybe this PR is only supported in the new one? I could hopefully take a stab at this in the coming days.

3. Sounds good!

SimenB:
I'm very +1 on more elegant ways to register metrics from libs. 😄

There was some discussion between myself and and a co-worker internally about how to avoid having to do peer deps all over to ensure stuff went into the correct registry. I'll quote them here as they seem relevant.

>The problem with prom-client being a singleton is that it can only be one version of it in a project. But, there is no need for the prom-client to be a singleton. If the design of the module was different they would have separated the data objects (the collected metrics) from the module itself.
>The data object can then be stored on global scope in node (not in the module) and since the data object probably changes less (highly possible almost never), its possible to have multiple versions of the module pushing data objects to the global scope.
>A better approach is to operate with an internal separation of data object, collector and dispatcher and use a socket between the collector and dispatcher which only sends data objects. This will make the prom-client work in clusters also.
>Both approaches makes it possible to have multiple versions of the prom-client in the same project.

siimon:
All preperations for this to be implemented should be fixed @eljasala

SimenB:
What about never registering metrics to the registry by default? So instead of introducing a new option, we just add a `register` method to each metric. If called with no argument, it registers it to the default registry, if not, to the registry provided as an argument. This is what the java client does.

Of course, the java client only exposes a builder pattern, and you have to call `create` or `register` (which internally calls `create`) to get an actual metric instance and not the builder, so it's easier to enforce people not doing the wrong thing.

siimon:
For me it's about what use case we should optimize for - modules or for use within the application itself. If you use prom-client within a module, you would like to control which register it is registered to, within apps you don't really care about that. I feel that it's important that the api is easy to use and that we avoid unnecessary code that would always be needed and looked the same within the common path.

But the question again is, what is the common path? Is it more common to use prom-client within a module or an app? I've always believed it to be within apps but I might be wrong.

I'm no fan of adding the 'builder pattern' for just registers, then we should rebuild creation for all metrics to follow that pattern and that doesn't feel very javascriptish. One idea I had, which I added to the 'default' configuration object is that you can provide the registers you want in the configuration object, and defaulting to the global one. So if you don't want to register it to the global register, you would create the metric like this

```javascript
  new Counter({ name: 'metric', help: 'help', registers: [] });
```

Then you won't need a separate parameter for just that. But as I said, if you guys think it's more common to use prom-client inside modules, the default could be changed to an empty array instead. Thoughts?

eljasala:
I'm okay with passing array as it's pretty easy to not register to global registry in that case.

I will try to get the modifications (merge object configuration & use array for registring) done today.

SimenB:
>If you use prom-client within a module, you would like to control which register it is registered to, within apps you don't really care about that.

Depending on what you mean with "you"; in my experience that's not the case. If "you" means author of module then you want to make it easy to plug it into the register of the consumers choice, you don't care _what_ registry that is.

I think, in general, it should be as easy as possible to register metrics with the default register. I don't think the use case of another register is so common that it should trump usability for the common use case (although we should of course make it _possible_).

My "issue" with the current API is that it's harder than it should be to register metrics with the correct registry as you have to manually make sure everything is deduped properly.

>I'm no fan of adding the 'builder pattern' for just registers, then we should rebuild creation for all metrics to follow that pattern and that doesn't feel very javascriptish.

I agree we'd have to make all APIs adhere to that pattern, and I also agree we shouldn't do it 😄 Just bringing it up as an example of how the Java client enforces explicit registering.

eljasala:
All changes should be in the PR now.

lib/counter.js
```diff
@@ -23,22 +23,24 @@ var getLabels = require('./util').getLabels;
...
 	var config;
 	if(isObject(name)) {
 		config = extend({
 			labelNames: [],
-			registers: [ register ]
 		}, name);
+		if(name.registers !== undefined) {
+			config.registers = name.registers;
+		} else {
+			config.registers = [ globalRegistry ];
```
eljasala: When looking into util-extend's code, it wasn't completely clear if it will try to merge arrays or replace them. With closer look, it seems to be replacing them in which case we can use it. I will do the changes.

eljasala:
Requested changes done.

siimon:
Thanks! :)

Issue #83, opened 2017-03-20
Way to provide labels for default metrics?


goofballLogic:
I can't find a way (from reading the code and readme) to add labels to the default metrics. This is important with regard to production deployment so that e.g. process id can be associated without forcing this data to be a grouping.

This is particularly important when needing to override the "instance" label (for pushgateway instances)

SimenB:
This is not currently possible, no. I like the idea though, but how do you imagine the API would look like?

siimon:
I agree, as long as we keep the API simple I don't mind this, but I think that will be the hardest part. 

I guess you could send in a function that returns an object with key values which is the label key and values. But I guess some validation is needed, since it could only be one level deep and not contain any functions etc..

SkeLLLa:
I also need some kind of such functionality. @siimon probably it could solve the problem if there will be some way to set "default" labels for all metrics. E.g. setting them in some kind of constructor:
```js
var pc = require('prom-client')({
  defaultLabels: {
    hostname: 'host1.example.com',
    zone: 'us-east'
  }
})
```
and all those labels should be passed to both default metrics constructors and custom. So:
```js
new Gauge(PROCESS_HEAP , 'Process heap size in bytes.')
```
should become:
```js
new Gauge(PROCESS_HEAP , 'Process heap size in bytes.', Object.keys(defaultLabels))
```
and custom metrics:
```js
var customLabels = ['label1', 'label2'];
new Gauge('custom_metric' , 'description', Object.keys(defaultLabels).concat(customLabels))
```

bspradling8:
+1 for this request.

SkeLLLa:
I've found some kind of workaround for me. As I use consul for registering my services I was able to add default labels through counul tags and parse them with promise consul meta information, but anyway if someone don't have such options there should be some way to do this through this lib.

SkeLLLa:
@goofballLogic, @bspradling8 I've made some "progress" according to this issue. The result you could find here: https://gitlab.com/prometheus-nodejs. There are separate plugins that could be initialized separately (see readmes). During initialization you could override metric name, description, add custom labels or disable some metrics. 

I'm currently in progress of porting all existing metrics into such separate plugins. Also there will be some new plugins like universal http/tcp stats plugin. Some metric names have change, so pay attention to it or just use override option to make them look like current. Also I plan to create some meta package or just helper that could start all those plugins with one or two function calls, but for now they should be started separately.

SimenB:
This is done in #137.

PR #101, opened 2017-04-28
Move collection of default metrics inside an explicit function call

SimenB:
/cc @SkeLLLa

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,16 +23,16 @@ NOTE: Some of the metrics, concerning File Descriptors and Memory, are only avai
 In addition, some Node-specific metrics are included, such as event loop lag, active handles and Node.js version. See what metrics there are in
 [lib/metrics](lib/metrics).
 
-The function returned from `collectDefaultMetrics` takes 2 options, a blacklist of metrics to skip, and a timeout for how often the probe should
-be fired. By default all probes are launched every 10 seconds, but this can be modified like this:
+The function returned from `collectDefaultMetrics` takes 1 option, a timeout for how often the probe should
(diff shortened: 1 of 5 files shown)


SkeLLLa:
BTW, after merging this it will be easier to implement this: https://github.com/siimon/prom-client/issues/83

It can be done through refactoring default metrics files to have 2 or 3 methods: `start`, `stop` and `clear` (optionally). And `start` method could take metric parameters that are needed to be overridden or added. Something like I've done here (it's a bit rewritten SimenB's gc-stats plugin) : https://gitlab.com/prometheus-nodejs/prometheus-plugin-gc-stats/blob/master/lib/index.js#L80. If every default plugin can take params on start, then "global" start function for all default metrics could also take them and send them to all default metrics `start` method.

SimenB:
Should we drop blacklisting of metrics at the same time? And do we need to be able to stop collection?

siimon:
I'd say yes on both, at least on blacklisting. Stopping could be desired in some cases

SimenB:
When would you want to stop?

siimon:
One I can think of now is if you want to stop the service in a graceful way, then it would make sense to stop

SimenB:
Fair enough 😄

Should I update this PR with removal of blacklist, or separate?

siimon:
Update this PR :)

SimenB:
@siimon Updated

siimon:
LGTM! I'll merge this but I don't want to publish to npm until #100 is completed aswell.. Sounds good?

SimenB:
Yup, sounds good!

PR #102, opened 2017-04-28
Use object instead of separate params as constructor parameters

siimon:
As discussed in #100, it's been a mess to add new functionality to the metrics because of parameter bloat in the constructor for each metric.

This PR tries to address this, replacing the old 'one parameter for each parameter' with an object that can hold all parameters needed for the constructor. This will make it easier to add new functionality in the future.

This PR should be backwards compatible, but the old one should be deprecated from now on.

@SimenB @SkeLLLa @eljasala could anyone of you have a look?

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Counters go up, and reset when the process restarts.
 
 ```js
 var client = require('prom-client');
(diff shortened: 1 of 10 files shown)


SimenB:
LGTM!

PR #103, opened 2017-04-29
Add timestamp to default metrics

SimenB:
/cc @rstuven


siimon:
Is there any reason why not all of the default metrics have timestamps added?

SimenB:
Very good question, my search failed me, and I didn't double check. Everyone should have it now! 😄

PR #107, opened 2017-05-18
Merge registries

siimon:
Makes it possible to merge multiple registries - handy when you want to expose modules registers within your app for example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## Unreleased
 ### Added
+- Ability to merge registries
+
 ### Changed
 
 ## [9.0.0] - 2017-05-06
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ You can prevent this by setting last parameter when creating the metric to `fals
 Using non-global registries requires creating Registry instance and adding it inside `registers` inside the configuration object. Alternatively
 you can pass an empty `registers` array and register it manually.
 
-Each registry has a merge function that enables you to expose multiple registries on the same endpoint.
+Each registry has a merge function that enables you to expose multiple registries on the same endpoint. If the same metric name exists in both registries, an error will be thrown.
 
 ```js
 var client = require('prom-client');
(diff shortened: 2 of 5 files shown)


siimon:
@eljasala @SimenB does this looks sane?

SimenB:
Code LGTM 😄 A couple of qs, though.

Should this return a new merged registry, or just combine them into an object? When is this merging supposed to be done?

siimon:
My thoughts were to always merge on each scrape request - but as it's
written now it doesn't really matter.

What would the object contain if it wasn't a register?

D34THWINGS:
We need this to merge registries of clustered apps !

SimenB:
If you're waiting on me, I have now objections. My questions got answered

eljasala:
Sorry for delay, missed the initial email.

1) Should the documented/tested way of calling merge be through Registry.merge? As merge is done between given parameters and the instance of which merge operation is called isn't used at all, it would likely be more natural to call `Registry.merge` than through global instance (`register.merge`)
2) Would it make sense to check to only throw an error if there is conflicting metrics? I think currently it would throw an error even when merging two registries which have same metric instance registered.

siimon:
> We need this to merge registries of clustered apps !

I'm afraid it won't solve all your problems, all metric names within each node must be unique across the cluster or else it will throw errors. I think it would get really advanced if we were to merge the values as well. Would probably be alot of edge cases.

@eljasala I'm fine with 1, was torn about them when implementing it so I'll change that. Not sure what you mean with 2, as I see it the only way to uniquely identify a metric is by it's name. How would you detect a conflict if you were not using the name?

eljasala:
On the 2nd point, I mean that instead of just checking if there's a metric with same name, you could also check if the metric is actually same one (`_metrics[metricFn.name] && _metrics[metricFn.name] !== metricsFn`). If it is, you don't need to abort the merge.

If you want to keep full backwards compatibility, you would need to do the check on merge before attempting to call registerMetric (or adding new option on registerMetric to allow it to perform the expanded check). Though admittedly it's a bit doubtful that someone relies on the fact that they will get an error when attempting to add same instance multiple times.

siimon:
Gotcha :) I'll add it to the current check, think we don't need to care about backwards compatibility in this case! :)

Issue #96, opened 2017-04-13
Optionally serve metrics in protocol buffer format?


snarlysodboxer:
Thank for the great library!

Would you be interested in a pull request related to serving metrics to prometheus in the protocol buffer format?

My thought was to create a new function in `lib/register.js` named something like `getMetricsAsProtobuf()` which returns a buffer containing the 32-bit varint-encoded record length-delimited protocol buffer messages of type `io.prometheus.client.MetricFamily` which prometheus expects. This buffer could be returned in the body of an HTTP response using something like Express or Koa.

There would be one added dependency, [protobufjs](https://www.npmjs.com/package/protobufjs).

The only issue is, you guys internally store the metrics in a format that is not compliant with prometheus' [metrics.proto](https://github.com/prometheus/client_model/blob/proto3/metrics.proto) messages, so either A), a conversion function is required, or B), we consider updating prom-client to internally store metrics as `MetricFamily` compliant objects. (E.G. for `metric.type`, we store `0` instead of `counter`, etc.)

Here's the conversion function we're using, it would be nice not to have to do this. It returns an array of metric objects from which we create protobuf messages using protobufjs's generated code's `MetricFamily.fromObject(metric)` function. (Note: support for Summary and Histogram is unfinished.)

```node
// Convert prom-client metric objects to `io.prometheus.client.MetricFamily` compliant objects
function convertMetrics() {
    let convertedMetrics = []
    for (let metricFamily of prom.getMetricsAsJSON()) {
        let newMetricFamily = {}
        newMetricFamily.name = metricFamily.name
        newMetricFamily.help = metricFamily.help
        let newMetrics = []
        for (let metric of metricFamily.values) {
            let newMetric = {}
            let newLabels = []
            Object.keys(metric.labels).forEach(function (key) {
                if (metric.labels[key].toString() != null)
                    newLabels.push({ "name": key, "value": metric.labels[key].toString() });
            });
            newMetric.label = newLabels;
            switch (metricFamily.type) {
                case "counter":
                    newMetricFamily.type = 0;
                    newMetric.counter = { "value": metric.value };
                    break;
                case "gauge":
                    newMetricFamily.type = 1;
                    newMetric.gauge = { "value": metric.value };
                    break;
                case "summary":
                    newMetricFamily.type = 2;
                    // TODO
                    break;
                case "histogram":
                    newMetricFamily.type = 4;
                    // TODO
                    break;
            }
[...]

PR #108, opened 2017-05-18
Support exposing metrics as Protocol Buffer messages

snarlysodboxer:
As discussed in [#96](https://github.com/siimon/prom-client/issues/96). This pull request is not finished or ready for merge, but I wanted to start the conversation and get feedback.

I've attempted to follow your styling, but some decisions had to be made. Let me know if you want things named, placed, or to work differently. I'm open to any and all feedback, I'm still getting the hang of Node. Also, you may or may not like my additions to `README.md`, I'm happy to remove them if you like.

Note, as my [comment here](https://github.com/snarlysodboxer/prom-client/blob/master/lib/registry.js#L5) states, I'm unsure if this could be a problem nor how to address if so, but it's technically possible for the `MetricFamily` variable to be accessed before it receives the results of the `protobuf.load` function. Any advice would be appreciated.

This commit adds [protobuf.js](https://github.com/dcodeIO/protobuf.js) as a dependency. It enables exposure of
the collected metrics in the protobuf format Prometheus expects.
Counter and Gauge are working, Histogram and Summary have yet to be
started. All tests pass.

NOTE: If you try to run the `example/server.js`, you'll get an error because histogram is unfinished, simply comment out the histogram portions of `server.js` and it will work.

### DIFF
--- a/example/server.js
+++ b/example/server.js
@@ -2,16 +2,14 @@
 
 var express = require('express');
 var server = express();
-var register = require('../lib/register');
+var client = require('../index');
+var register = client.register;
 
-var Histogram = require('../').Histogram;
-var h = new Histogram({ name: 'test_histogram', help: 'Example of a histogram', labelNames: [ 'code' ] });
+var h = new client.Histogram({ name: 'test_histogram', help: 'Example of a histogram', labelNames: [ 'code' ] });
 
-var Counter = require('../').Counter;
-var c = new Counter({ name: 'test_counter', help: 'Example of a counter', labelNames: [ 'code' ] });
+var c = new client.Counter({ name: 'test_counter', help: 'Example of a counter', labelNames: [ 'code' ] });
 
-var Gauge = require('../').Gauge;
-var g = new Gauge({ name: 'test_gauge', help: 'Example of a gauge', labelNames: [ 'method', 'code' ] });
+var g = new client.Gauge({ name: 'test_gauge', help: 'Example of a gauge', labelNames: [ 'method', 'code' ] });
 
 setTimeout(function() {
 	h.labels('200').observe(Math.random());
@@ -45,6 +43,12 @@ server.get('/metrics', function(req, res) {
 server.get('/metrics/counter', function(req, res) {
 	res.set('Content-Type', register.contentType);
 	res.end(register.getSingleMetricAsString('test_counter'));
+});
+
+server.get('/metrics/protobuf', function(req, res) {
+	client.collectDefaultMetrics();
+	res.set('Content-Type', register.contentTypeProto);
+	res.end(register.metricsProtobuf());
 });
 
 console.log('Server listening to 3000, metrics exposed on /metrics endpoint'); //eslint-disable-line no-console
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -1,5 +1,16 @@
 'use strict';
 
+var protobuf = require('protobufjs');
+var path = require('path');
+// TODO consider the possibility this gets loaded too late
+var MetricFamily;
+protobuf.load(path.resolve( __dirname, '../metrics.proto'), function(err, root) {
+	if(err !== null) {
+		throw err;
+	}
+	MetricFamily = root.lookupType('io.prometheus.client.MetricFamily');
+});
+
 function escapeString(str) {
 	return str.replace(/\n/g, '\\n').replace(/\\(?!n)/g, '\\\\');
 }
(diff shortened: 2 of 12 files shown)


test/registerTest.js
```diff
@@ -110,6 +120,30 @@ describe('register', function() {
 		});
 	});
 
+	describe('should output metrics as Protobuf', function() {
+		// If length delimiting is incorrect, it could work for one metric and
+		// 	 not multiple, or vise versa - Test both
+		var metric;
+		beforeEach(function() {
```
SimenB: clear registry in `afterEach` so that execution order does not matter

test/registerTest.js
```diff
@@ -1,5 +1,15 @@
 'use strict';
 
+var protobuf = require('protobufjs');
+var MetricFamily;
+protobuf.load('metrics.proto', function(err, root) {
```
SimenB: this should be in a top level `describe`'s `before` to avoid timing issues

metrics.proto
```diff
@@ -0,0 +1,83 @@
+// FROM https://github.com/prometheus/client_model/blob/proto3/metrics.proto
```
SimenB: This isn't merged, do we want it instead of what's in master? https://github.com/prometheus/client_model/pull/14

lib/registry.js
```diff
@@ -1,5 +1,16 @@
 'use strict';
 
+var protobuf = require('protobufjs');
+var path = require('path');
+// TODO consider the possibility this gets loaded too late
```
SimenB: Can generate at build time?

https://github.com/dcodeIO/ProtoBuf.js/#using-generated-static-code

example/server.js
```diff
@@ -47,5 +45,11 @@ server.get('/metrics/counter', function(req, res) {
 	res.end(register.getSingleMetricAsString('test_counter'));
 });
 
+server.get('/metrics/protobuf', function(req, res) {
+	client.collectDefaultMetrics();
```
SimenB: Put this it the top level

SimenB:
Nice start, looking forward to it 😄

Left a few nits, but overall looks good

test/registerTest.js
snarlysodboxer: will do. i was thinking maybe the top level `afterEach` took care of this, but wasn't sure.

test/registerTest.js
snarlysodboxer: will do.

metrics.proto
snarlysodboxer: good question. i'm not sure, i [requested feedback here](https://github.com/prometheus/client_model/issues/13#issuecomment-302881167).

lib/registry.js
snarlysodboxer: this is what i was doing in our code before contributing it, but i figured you didn't want the generated code in the repo.

i'm assuming you mean npm build time. now that you say, that makes the most sense. great idea, thanks!

example/server.js
snarlysodboxer: @SimenB i'm assuming you mean the `client.collectDefaultMetrics();` part, not the `server.get('/metrics/protobuf', function(req, res) {` part.

will do.

SimenB: Yup 🙂

lib/registry.js
SimenB: Yes, as a prepublish step

test/registerTest.js
SimenB: If you build it before running tests that might be cleaner. Your choice of course

siimon:
I've not had the time to look at the code yet, but I read the comment by Björn in https://github.com/prometheus/client_model/issues/13#issuecomment-302881167 and if I understood him correctly they're not sure to have protobuf support in 2.0? If that is correct, that makes me wonder if it's wise to implement this in prom-client? What are your thoughts?

SimenB:
+1 on waiting for the dust to settle on whether or not they will re-add protobuf. He seemed against removing it, but still.

snarlysodboxer:
I've submitted an [issue here](https://github.com/prometheus/prometheus/issues/2788) asking if they plan to drop the Protobuf Exposition Format. It's hard to imagine they would, we'll see what happens.

SkeLLLa:
Probably it would be better and faster if you compile `.proto` file into commonjs module via pbjs and use it instead of loading and parsing`.proto` file? I think simple require of js module is better than async loading of proto file.

example/server.js
marcbachmann: I guess this should be 
```js
server.get('/metrics/protobuf', client.collectDefaultMetrics)
```
or
```js
server.get('/metrics/protobuf', function (req, res) {
  client.collectDefaultMetrics(req, res)
}
```

SimenB:
Closing, see https://github.com/prometheus/docs/pull/972.

Happy to reopen if it gets reintroduced

Bruno-DaSilva:
can this be reopened? Reintroduced! https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md#protobuf-format

PR #109, opened 2017-05-20
Correct typedef for constructor object

SimenB:


siimon:
Bah, I made the same fixes in #107, but I'll merge this and revert the commit on that branch :) 

Thanks!

PR #110, opened 2017-05-20
Include name of unknown label in error

SimenB:
I had a typo when using the object form of labels. Would be nice if it the error included info about what label I used, and which were available

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -1,15 +1,19 @@
 'use strict';
 
+var util = require('util');
+
+// These are from https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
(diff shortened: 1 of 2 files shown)


lib/validation.js
```diff
@@ -1,15 +1,19 @@
 'use strict';
 
+var util = require('util');
+
+// These are from https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
+var metricRegexp = /^[a-zA-Z_:][a-zA-Z0-9_:]*$/;
+var labelRegexp = /^[a-zA-Z_][a-zA-Z0-9_]*$/;
```
SimenB: Probably JITed away, but why not help V8 out a bit 🙂

siimon:
LGTM! Thanks!

PR #120, opened 2017-06-18
Format code with prettier

SimenB:
I love prettier. 😄 Makes it so easy to have consistent styling, and it allows us to remove some eslint rules as well

### DIFF
--- a/lib/pushgateway.js
+++ b/lib/pushgateway.js
@@ -6,7 +6,7 @@ var https = require('https');
 var register = require('./register');
 
 function Pushgateway(gatewayUrl, options, registry) {
-	if(!registry) {
+	if (!registry) {
 		registry = register;
 	}
 	this.registry = registry;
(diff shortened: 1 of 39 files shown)


lib/pushgateway.js
```diff
@@ -63,19 +70,26 @@ function useGateway(method, job, groupings, callback) {
...
 function generateGroupings(groupings) {
-	if(!groupings) {
+	if (!groupings) {
 		return '';
 	}
-	return Object.keys(groupings).map(function(key) {
-		return ['/', encodeURIComponent(key), '/', encodeURIComponent(groupings[key])].join('');
-	}).join('');
+	return Object.keys(groupings)
+		.map(function(key) {
```
siimon: Not so much about Prettier, but we could use templated strings instead of Array.join

lib/registry.js
```diff
@@ -48,14 +47,19 @@ var registry = (function() {
...
 			.map(this.getMetricAsPrometheusString, this)
 			.join('\n');
 	};
 
 	Registry.prototype.registerMetric = function(metricFn) {
-		if(_metrics[metricFn.name] && _metrics[metricFn.name] !== metricFn) {
-			throw new Error('A metric with the name ' + metricFn.name + ' has already been registered.');
+		if (_metrics[metricFn.name] && _metrics[metricFn.name] !== metricFn) {
+			throw new Error(
+				'A metric with the name ' +
```
siimon: Same here, templated strings is better

lib/validation.js
```diff
@@ -22,8 +22,13 @@ exports.validateLabelName = function(names) {
 
 exports.validateLabel = function validateLabel(savedLabels, labels) {
 	Object.keys(labels).forEach(function(label) {
-		if(savedLabels.indexOf(label) === -1) {
-			throw new Error('Added label "' + label + '" is not included in initial labelset: ' + util.inspect(savedLabels));
+		if (savedLabels.indexOf(label) === -1) {
+			throw new Error(
```
siimon: Templated string 🙂

SimenB:
I'll send a separate PR for template strings 🙂

PR #117, opened 2017-06-17
Node 4 syntax

SimenB:
Commits should probably be looked at individually. Can land stuff separately if you want?

What do you think about adding prettier, btw? Can simplify the eslint config as lots of things would be removed, and we'll probably end up with higher consistency on the code.

PR #121, opened 2017-06-18
Remove util-extend

SimenB:
Piecemeal #117


SimenB:
`util-extend` is just a shallow copy, same as `Object-assign`. https://github.com/isaacs/util-extend/blob/master/extend.js

PR #124, opened 2017-06-18
Use object shorthand

SimenB:
Piecemeal #117


siimon:
Do you never sleep? ☺️

SimenB:
Sleep is for the weak! 😄

I had trouble sleeping. Not to happy about it now, though, as I'm on my way to work pretty sleepy 😆

PR #125, opened 2017-06-18
Convert prototype to classes

SimenB:
Piecemeal #117

This diff should be viewed without whitespace changes, as the indentation has changed.

https://github.com/siimon/prom-client/pull/125/files?w=1

### DIFF
--- a/lib/counter.js
+++ b/lib/counter.js
@@ -2,7 +2,7 @@
  * Counter metric
  */
 'use strict';
-const globalRegistry = require('./register');
+const globalRegistry = require('./registry').globalRegistry;
 const type = 'counter';
 const isNumber = require('./util').isNumber;
 const isDate = require('./util').isDate;
--- a/lib/counter.js
+++ b/lib/counter.js
@@ -15,93 +15,93 @@ const isObject = require('./util').isObject;
 
 const getLabels = require('./util').getLabels;
 
-/**
- * Counter
- * @param {string} name - Name of the metric
- * @param {string} help - Help description for the metric
- * @param {Array.<string>} labels - Labels
- * @constructor
- */
-function Counter(name, help, labels) {
-	let config;
-	if (isObject(name)) {
-		config = Object.assign(
-			{
-				labelNames: [],
+class Counter {
+	/**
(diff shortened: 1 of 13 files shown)


lib/register.js
```diff
@@ -1,26 +1,4 @@
 'use strict';
```
SimenB: this file can probably be deleted, and a global registry be created directly in `registry.js`

lib/counter.js
```diff
@@ -16,91 +16,92 @@ var extend = require('util-extend');
...
 
-	this.name = name;
-	this.name = config.name;
-	this.hashMap = {};
+		if(!validateLabelNames(config.labelNames)) {
+			throw new Error('Invalid label name');
+		}
 
-	this.labelNames = (config.labelNames || []);
+		this.name = name;
```
siimon: Could remove the first assignment of name, even though it exists in current version🙂

lib/register.js
siimon: I don't completely remember, but I think there was something with backwards compatibility which was why we left it as it is, can be wrong though

lib/counter.js
SimenB: Nice spot, removed 🙂

lib/register.js
SimenB: I think we might want to do a major bump on these changes either way, just to be safe.

Before, it only broke on pushgateway, which is not too usual to use, and it only brok on something polyfillable. Now we break on syntax. WDYT?

siimon: Not sure tbh, but maybe you are right.. It might be nice to do a major release as well to minimize the effects of any regressions since there are a lot of changes.

So let's remove this file also as we don't care about backwards compatibility!

SimenB:
@siimon This is now green. Ok to merge?

lib/register.js
SimenB: Great! Done 🙂

siimon:
Great! Could you do a major release with these changes?

SimenB:
Sure. Will try to get in a couple of more changes first 🙂

PR #126, opened 2017-06-18
Make tests pass in node 8

SimenB:
This makes the tests slower (as it just sets up an actual express server instead of mocking out the requests), but at least they pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "eslint-plugin-prettier": "^2.1.2",
     "express": "^4.13.3",
     "jest": "^20.0.4",
-    "nock": "^9.0.11",
     "prettier": "^1.4.4",
     "sinon": "^2.0.0",
     "typescript": "^2.0.3"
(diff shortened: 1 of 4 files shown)


siimon:
A thought, if jest PR is merged, is it possible to use mock out the http module there and use it instead?

SimenB:
I haven't tried mocking out http before. It'll be different kinds of tests, testing out calling the http module instead of the actual network request. But that's ok, I guess?

siimon:
Yeah, but with tests used in nock doesn't make any requests either. I think asserting on the http module is fine. But I'm not sure you can catch http headers.

SimenB:
It's part of the options sent in to `http(s).request(options, cb)`, so just asserting on `options` should be fine. We lose coverage on the cb though, unless we want to mess with eventemitters etc

siimon:
I guess that's ok for now until we find a better solution

SimenB:
@siimon updated with mocking out http

PR #128, opened 2017-06-18
Use jest for tests

SimenB:
I much prefer jest these days, so thought I'd take a crack at converting the tests. This was basically just running `jest-codemods`, so no hard feelings if you don't want it 🙂

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+coverage/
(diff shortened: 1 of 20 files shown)


siimon:
I don't mind, used Jest at work a couple of years ago but then it was rather immature, but I think it's better now. Also gives us better chance to mock out stuff

PR #131, opened 2017-06-19
Add precommit hook to format code

SimenB:
I like doing this. That way, you don't really have to formatting the code as you go. Just commit, and it'll be reformatted for you.

If you really want to commit regardless, you can do `--no-verify` (which it will tell you to do if it fails, so no googling necessary).

This commit also formats the d.ts file, cuz why not 🙂


siimon:
Why not? 🙂

PR #132, opened 2017-06-19
Make merge static in ts typings

SimenB:
Missing from #107 I guess.

@SkeLLLa is this correct syntax?


SkeLLLa:
@SimenB yep, looks fine.

PR #134, opened 2017-06-24
Print deprecation warnings on deprecated constructors

SimenB:
Since the next is a major anyways, might get this is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,11 @@
 # Changelog
-All noteable changes to this project will be documented in this file.
+All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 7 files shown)


siimon:
I'm not sure we should do that already, want folks to have some more time to migrate since it will literally break all metrics. Should we add a deprecation warning in this release instead?

SimenB:
Yeah, that makes sense

SimenB:
@siimon Updated.

SimenB:
One thing to note is that the java client support having 2 strings provided directly (name and help). It's only when you add labels, custom buckets/percentiles etc that it forces you to be explicit. Should we support the same thing?

If passed 2 arguments, check that they're both strings, and use as name and help. Otherwise print deprecation warning (if not an object).

siimon:
I rather just have a configuration object to keep the api simple, what do you think?

PR #135, opened 2017-06-24
Allow registering default metrics to other registries

SimenB:
Maybe take an object to `collectDefaultMetrics`? I'm going to tackle custom labels for the default metrics now, yet another option and it gets quite hairy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - Print deprecation warning when metrics are constructed using non-objects
 ### Added
 - Ability to set default labels by registry
+- Allow passing in `registry` as second argument to `collectDefaultMetrics` to use that instead of the default registry
 ### Changed
 - Convert code base to ES2015 code (node 4)
   - add engines field to package.json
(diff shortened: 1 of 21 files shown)


siimon:
To get consistent, maybe there should only be one argument with a configuration object?

siimon:
But maybe support both versions in this release and add a deprecation warning?

siimon:
Sigh, just read the comment you wrote about configuration object so nvm about my first comment..

SimenB:
I'll change it to an object and support both variants 🙂

SimenB:
Updated 🙂

SimenB:
@siimon I rebased this, and the comments were lost... If you comment on the PR instead of the commit, it should be ok.

Only missing this, and I'll cut a v10.

SimenB:
Git reflog to the rescue: https://github.com/siimon/prom-client/commit/5f7a633649721d24984c6c74f2facf2af94fbe7f

SimenB:
@siimon anything else before release? Or should I go ahead with 10?

siimon:
Nope, not that I know of!

SimenB:
Awesome.

https://github.com/siimon/prom-client/releases/tag/v10.0.0

siimon:
Great work!

PR #136, opened 2017-06-26
Upgrade eslint, and activate recommended config

SimenB:
eslint@4.1.0 supports glob based config, which allows us to drop the extra eslintrc file.

I also added `eslint:recommended` because, well, it's recommended.


.eslintrc
```diff
@@ -1,62 +1,60 @@
...
-	"no-shadow-restricted-names": 2,
-	"no-unused-expressions": 2,
-	"no-unused-vars": 2,
-	"no-use-before-define": 0,
-	"no-new": 0,
-	"dot-notation": [
-	  2,
-	  {
-		"allowKeywords": true
+	"no-underscore-dangle": "off",
```
SimenB: why are some turned off?

siimon: Im guessing it's just a copy and paste from another project I used, so no reason really :)

PR #137, opened 2017-06-26
Add support for default labels

nwest1:
I believe this is a pretty good solution for #83 

It's simple, doesn't break any APIs, allows you to configure static labels per registry, and defaulted values can be overridden.

Let me know if this wasn't what you were thinking about this, or if there are any styling considerations I should be changing.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ This will output metrics in the following way:
 process_resident_memory_bytes{serviceName="api-v1"} 33853440 1498510040309
 ```
 
-Default metrics will be overridden if there is a name conflict.
+Default labels will be overridden if there is a name conflict.
 
 `register.clear()` will clear default labels.
 
--- a/README.md
+++ b/README.md
@@ -195,8 +195,8 @@ xhrRequest(function(err, res) {
 
 Static labels may be applied to every metric emitted by a registry:
 ```js
-var client = require('prom-client');
-var defaultLabels = { serviceName: "api-v1"};
+const client = require('prom-client');
+const defaultLabels = { serviceName: "api-v1" };
 client.register.setDefaultLabels(defaultLabels);
 ```
 
--- a/README.md
+++ b/README.md
@@ -207,7 +207,9 @@ This will output metrics in the following way:
 process_resident_memory_bytes{serviceName="api-v1"} 33853440 1498510040309
 ```
 
-Note: `register.clear()` will clear default labels.
+Default metrics will be overridden if there is a name conflict.
+
+`register.clear()` will clear default labels.
 
 #### Timestamps
 
--- a/README.md
+++ b/README.md
@@ -191,6 +191,24 @@ xhrRequest(function(err, res) {
 });
 ```
 
+##### Default Labels (segmented by registry)
+
+Static labels may be applied to every metric emitted by a registry:
+```js
+var client = require('prom-client');
+var defaultLabels = { serviceName: "api-v1"};
+client.register.setDefaultLabels(defaultLabels);
+```
+
+This will output metrics in the following way:
+```
+# HELP process_resident_memory_bytes Resident memory size in bytes.
+# TYPE process_resident_memory_bytes gauge
+process_resident_memory_bytes{serviceName="api-v1"} 33853440 1498510040309
+```
+
+Note: `register.clear()` will clear default labels.
+
 #### Timestamps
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -191,6 +191,26 @@ xhrRequest(function(err, res) {
...
+```
+
+This will output metrics in the following way:
+```
+# HELP process_resident_memory_bytes Resident memory size in bytes.
+# TYPE process_resident_memory_bytes gauge
+process_resident_memory_bytes{serviceName="api-v1"} 33853440 1498510040309
+```
+
+Default metrics will be overridden if there is a name conflict.
```
SimenB: "labels", not "metrics", right?

lib/registry.js
```diff
@@ -27,10 +28,12 @@ class Registry {
 		const type = ['#', 'TYPE', name, item.type].join(' ');
 
 		const values = (item.values || []).reduce((valAcc, val) => {
-			const labels = Object.keys(val.labels || {}).map(
+			let labels = Object.keys(val.labels || {}).map(
 				key => `${key}="${escapeLabelValue(val.labels[key])}"`
 			);
 
+			labels = Object.assign(this._defaultLabels, labels);
```
SimenB: not sure `Object.assign` makes sense here, they are arrays not objects. should be `this._defaultLables.concat(labels)`. Unsure how to handle dupes, though? I can see from the tests that it works, but it feels weird to me

SimenB: if you keep `_defaultLabels` as an object internally, then do `Object.assign` before the `map`?

nwest1:
Thanks for the feedback! Glad you like it.

How should the `static merge` function work? I'm not sure what should be intended there.

nwest1:
Not sure if I did the type def'n correctly. Let me know if there's anything else 😄

SimenB:
Looks good!

If you rebase on master the build should be green again

README.md
```diff
@@ -191,6 +191,26 @@ xhrRequest(function(err, res) {
 });
 ```
 
+##### Default Labels (segmented by registry)
+
+Static labels may be applied to every metric emitted by a registry:
+```js
+var client = require('prom-client');
+var defaultLabels = { serviceName: "api-v1"};
```
SimenB: this has inconsistent spacing

SimenB: and can you use `const`?

nwest1:
I'm still getting local npm test failures - 
```
/Users/z002b1g/work/prom-client/lib/metrics/osMemoryHeap.js
  28:29  error  Replace `⏎↹` with `·`  prettier/prettier
  30:2   error  Delete `↹`             prettier/prettier
  31:1   error  Delete `↹`             prettier/prettier

/Users/z002b1g/work/prom-client/lib/pushgateway.js
  44:24  error  Replace `⏎↹↹gatewayUrlParsed.pathname·&&·` with `·gatewayUrlParsed.pathname·&&⏎↹↹`  prettier/prettier
  46:3   error  Delete `↹`                                                                          prettier/prettier
  47:1   error  Delete `↹`
```

SimenB:
Have you ran `npm install` after rebase?

nwest1:
Oops! That did it. Thanks.

siimon:
I like the idea tying it together with the registry, that makes sense I think.

One thought I had earlier was to instead of providing an object which is the default labels, you could instead provide a function that in turn returns an object which is the default labels. That would make it more flexible and more powerful, however the api might be a bit more complex.

SimenB:
What's the use case for a function? When would you want to use it?

nwest1:
Functional options are a great pattern to allow for expanding an API without breaking it. It would require some rework for sure.

This is for golang, but I think the pattern still applies to js:
https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

SimenB:
I have no strong feeling either way. Doing `register.defaultMetrics({ foo: 'bar' })` or `register.defaultMetrics(() => { foo: 'bar' })` doesn't matter to me. Maybe support both?

While I get the general argument "it's more extensible", I don't get what it would give default metrics. Would it receive the metric as an argument? Still, I don't see what you could really do with it.

nwest1:
I'll let you guys decide how you'd like to do this. 
I just need something that works for the edge case I'm working with right now 😄

siimon:
Well, it's probably bait YAGNI, so lets keep it as it is and if we find another use case where a function is better we could probably live with both versions for awhile..

Issue #118, opened 2017-06-17
Un-incremented counters don't have value 0


paulborza:
When I navigate to my metrics endpoint, some counters don't have values because they were never incremented. Such a counter is the `unknown_failure_total` which gets incremented on unhandled exceptions.

I have Prometheus hooked up to Grafana and it keeps alerting me that there's missing data for some counters. I don't want to suppress this alert because missing data could also mean that Grafana is not able to reach my Prometheus server.

The `HELP` and `TYPE` statements are printed even for un-incremented counters, but their zero value is not. Should a 0 value be printed for counters which have yet to be incremented?

```
# HELP postgres_upsert_failure_total postgres_upsert_failure_total
# TYPE postgres_upsert_failure_total counter
postgres_upsert_failure_total{environment="production"} 2

# HELP unknown_failure_total cli_unknown_failure_total
# TYPE unknown_failure_total counter
```

Thanks!

siimon:
Not sure what the "correct" behaviour is in this case, you could also argue that it shouldn't have any value at all until you have set any.

A workaround in your case is to inc with 0 just after you created it

paulborza:
Thanks for the workaround suggestion. Here are my thoughts:

1. According to the [official Prometheus documentation about writing client libraries](https://prometheus.io/docs/instrumenting/writing_clientlibs/): "Counters MUST start at 0." So when a new counter is initialized, its value has to be zero. And it is zero, because otherwise the `inc` method wouldn't make sense unless a starting value is offered.
2. Its odd to see `HELP` and `TYPE` metadata displayed but the counter value not.

I think we should always print the value of counters, even if they're zero. A missing counter is != than a counter which was never initialized. What do you think?

SimenB:
I agree with @paulborza. It also matches what the java client does.

```java
import io.prometheus.client.CollectorRegistry;
import io.prometheus.client.Counter;
import io.prometheus.client.exporter.common.TextFormat;

import java.io.StringWriter;
import java.util.Collections;

public class App {
    static Counter c = Counter.build("name", "help").register();

    public static void main(String[] args) throws Exception {
        StringWriter writer = new StringWriter();
        TextFormat.write004(writer, CollectorRegistry.defaultRegistry.filteredMetricFamilySamples(Collections.<String>emptySet()));
        System.out.println(writer.toString());
    }
}
```

Prints

```
# HELP name help
# TYPE name counter
name 0.0
```

SimenB:
[...]

PR #139, opened 2017-07-04
Init metrics to zero

siimon:
Fixing #118 

Added fixes for Counters, Gauges and Histograms.. Summary I'm not sure what to do about tbh.. After a quick look it's not an easy fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 ## [Unreleased]
 ### Breaking
 ### Added
+- Metrics should be initialized to 0 when there are no labels
 ### Changed
 
 ## [10.0.0] - 2017-07-04
--- a/lib/counter.js
+++ b/lib/counter.js
@@ -63,7 +63,7 @@ class Counter {
 		}
 
 		this.name = config.name;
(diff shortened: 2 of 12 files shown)


siimon:
Great! Thanks! :)

SimenB:
Ok, should be done now.

Had to revert most of the test changes when considering labels

siimon:
LGTM!

SimenB:
@dm3 @paulborza does this look correct? You can check the snapshot test for output with and without labels.

paulborza:
@SimenB why initialize the metrics only when there are no labels? The metrics should be initialized to 0 across all labels.

siimon:
How would you do that since you don't know the values for the labels? You only know the label names when you create the metric

paulborza:
@siimon You're right; my bad. The output looks good for metrics with no labels.
Argh, I just wish there was a way Prometheus would support some sort of wildcard syntax to initialize metrics across all dimensions/labels; but apparently it doesn't.

Thanks for fixing this! :)

paulborza:
Once merged, can you also publish the new version to the registry? Thanks

SimenB:
>Once merged, can you also publish the new version to the registry? Thanks

Of course 🙂

PR #146, opened 2017-07-30
Compatibility with pushgateway 0.4

RubenSlabbert:
As of pushgateway 0.4, timestamps are no longer accepted when pushing.

See: https://github.com/prometheus/pushgateway/pull/119

### DIFF
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -10,6 +10,10 @@ function escapeLabelValue(str) {
 	return escapeString(str).replace(/"/g, '\\"');
 }
 
+const defaultMetricsOpts = {
+	timestamps: true
+};
+
 class Registry {
 	constructor() {
 		this._metrics = {};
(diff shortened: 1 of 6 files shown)


siimon:
Maybe a silly question, but why do you have timestamps in your metrics in the first place when you know it's not supported by the pushgateway?

RubenSlabbert:
As far as I could gather, the default metrics collection automatically adds timestamps to guages and counters, which will then automatically be pushed to the gateway.

siimon:
Gotcha!

Add some test cases and update the README + Changelog :)

lib/registry.js
```diff
@@ -48,9 +57,9 @@ class Registry {
 		return acc;
 	}
 
-	metrics() {
+	metrics(opts) {
 		return this.getMetricsAsArray()
-			.map(this.getMetricAsPrometheusString, this)
+			.map(metric => this.getMetricAsPrometheusString.bind(this)(metric, opts))
```
SimenB: Don't use bind, just call it normally. The arrow function ensures correct `this`

RubenSlabbert:
All the requested changes have been implemented. Let me know if you'd like more tests.

RubenSlabbert:
Works perfectly for me with [pushgateway 0.4](https://github.com/prometheus/pushgateway/releases/tag/v0.4.0).

siimon:
Great! Thanks! :)

SimenB:
Published as 10.0.3.

Thanks!

Issue #82, opened 2017-03-20
Guidance for clustered processes


goofballLogic:
There is currently no guidance provided for clustered processes (default mechanism for running an app on multi-core servers). https://nodejs.org/api/cluster.html#cluster_cluster

By default, a clustered process, operating in round-robin fashion, will only serve metrics local to the process which handled a particular scrape request from Prometheus. This makes default metrics like "active requests" meaningless.

Possible solutions include:

1. Recommend that users "push" metrics rather than the default pull mechanism. If this is the solution, the documentation should do a better job of showing how to set up a regular push of the default metrics via the push gateway
2. Provide a mechanism for collecting metrics from multiple child processes - possibly by feeding them back to the master process, or e.g. via socket to a dedicated process which can handle all requests for metrics.

siimon:
I don't think it's up to prom-client to give any recommendations to this since it depends on how, where, type of application etc. I rather give no recommendation than giving a recommendation that only fits for certain conditions.

However, if someone builds a library on top of prom-client that acts like a master process and collect metrics from other node processes, I'll happily link to it!

disjunction:
@goofballLogic i use prometheus in the environment like you describe and have no problems with it. You have all the aggregation functions in prometheus for that. In your case just use sum for the "active requests".

goofballLogic:
Using "sum" does help resolve this problem because you are only ever receiving partial metrics. If you have e.g. 4 workers on a 4-core machine and you are Counting requests on each worker, and you expose a /metrics endpoint, each call to /metrics will only retrieve the metrics for 25% of requests processed. There's no way to make 4 calls to get all four sets of metrics and then sum them.

goofballLogic:
@siimon we're talking about the default way to run a node.js application in production here. Initially I followed the directions in the readme which work using a single process e.g. in development. Once we deployed to staging it became apparent that the default mechanism can't work for the deployment configuration recommended by node.js.

Would you be opposed to me at least documenting the problem and outlining how to work around it using the Pushgateway?

disjunction:
@goofballLogic sorry, i misunderstood with workers. As an idea - sum should work as long as you provide a unique additional label for each worker. But i agree it becomes an ugly workaround then.

siimon:
@goofballLogic im not sure about that - as I said it depends on certain things, not only how you built it, i.e how your infrastructure looks like etc.

[...]

Issue #80, opened 2017-03-16
Aggregate metrics in multiprocess mode


tommiv:
So we use pm2 to run nodes "cluster" in fork mode to fully utilize server cpu cores. This means that every node.js instance can use prom-client to generate the metrics of its own process, but we need one more step to aggregate metrics from every node to a single metric. Is there some solutions with prom-client or I need an external aggregator for this (is there any in the wild then?). Sorry it this is a wrong place to ask this question.

siimon:
Prom-client does not support that, and I don't think it ever will. 

LBest would probably be as you said, to build something on top of prom-client that aggregates and exposes them. But it has some challenges, maybe you would like to know stats for a specific node? That data will be lost when aggregated etc!

tommiv:
Thanks for your answer, I will try some workarounds.
> That data will be lost when aggregated
this is true, but we probably ok with this. Since we can't collect data from each process with standard pull model, we need some trade-offs.

saqibarfeen:
With the help of [this](https://gist.github.com/yekver/34c9d41c1c4ea478151574ea539e9953) and [this](https://github.com/slanatech/swaggerstats.io/blob/master/docs/guide/prometheus.md)
I was able to achieve exactly what you are asking.  See my [gist](https://gist.github.com/saqibarfeen/8bc729ade1c1b17fad764d924e64af32)
(in case someone else reads this)

PR #147, opened 2017-07-30
Cluster support

zbjornson:
Hello! This is an opening gambit for supporting aggregated metrics for use with node's `cluster` module. It's not ready for merge; looking for feedback first as to whether or not this would go in prom-client or a separate module that wraps prom-client.

It's only lightly tested, and there are no unit tests yet.

I made the changes completely isolated in a separate file that is not loaded by default. From reading issues #82 and #80, I wasn't sure if a PR that provides this functionality would be accepted, so I wanted the changes to work if the functionality was provided as a separate Node.js module. However, that means that [the `metric.aggregate` stuff](https://github.com/siimon/prom-client/compare/master...zbjornson:cluster?expand=1#diff-0faa53fc02580d5de2ebb484c41d691cR98) is a bit ugly. If this is something you want in prom-client, I could definitely add the aggregation mode to the metrics classes.

Ref #82 
Ref #80 

/cc @goofballLogic and @tommiv in case you're interested

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,8 @@
   "plugins": ["prettier"],
   "extends": "eslint:recommended",
   "env": {
-	"node": true
+	"node": true,
+	"es6": true
   },
   "parserOptions": {
 	"ecmaVersion": 2015
@@ -55,6 +56,12 @@
 		"no-shadow": "off",
 		"no-unused-expressions": "off"
 	  }
+	},
+	{
+		"files": ["example/**/*.js"],
+		"rules": {
+			"no-console": "off"
+		}
 	}
   ]
 }
--- a/README.md
+++ b/README.md
@@ -4,7 +4,17 @@ A prometheus client for node.js that supports histogram, summaries, gauges and c
 
 ### Usage
 
-See example folder for a sample usage. The library does not bundle any web framework, to expose the metrics just return the metrics() function in the registry.
+See example folder for a sample usage. The library does not bundle any web framework, to expose the metrics just return the `metrics()` function in the registry.
+
+#### Usage with Node.js's `cluster` module
+
+Node.js's `cluster` module spawns multiple processes and hands off socket connections to those workers. Returning metrics from a worker's local registry will only reveal that individual worker's metrics, which is generally undesirable. To solve this, you can aggregate all of the workers' metrics in the master process. See `example/cluster.js` for an example.
+
+Default metrics use sensible aggregation methods. Custom metrics are summed across workers by default. To use a different aggregation method, set the `aggregator` property in the metric config to one of 'sum', 'first', 'min', 'max', 'average' or 'omit'. (See `lib/metrics/version.js` for an example.)
+
+If you need to expose metrics about an individual worker, you can include a value that is unique to the worker (such as the worker ID or process ID) in a label. (See `example/server.js` for an example using `worker_${cluster.worker.id}` as a label value.)
+
+Metrics are aggregated from the global registry by default. To use a different registry, call `client.AggregatorRegistry.setRegistries(registryOrArrayOfRegistries)` from the worker processes.
 
 ### API
 
@@ -248,7 +258,7 @@ You can prevent this by setting last parameter when creating the metric to `fals
 Using non-global registries requires creating Registry instance and adding it inside `registers` inside the configuration object. Alternatively
 you can pass an empty `registers` array and register it manually.
 
-Registry has a merge function that enables you to expose multiple registries on the same endpoint. If the same metric name exists in both registries, an error will be thrown.
+Registry has a `merge` function that enables you to expose multiple registries on the same endpoint. If the same metric name exists in both registries, an error will be thrown.
 
 ```js
 const client = require('prom-client');
@@ -259,6 +269,15 @@ registry.registerMetric(histogram);
 counter.inc();
 
 const mergedRegistries = client.Registry.merge([registry, client.register]);
+```
+
+If you want to use multiple or non-default registries with the Node.js `cluster` module, you will need to set the registry/registries to aggregate from:
+
+```js
+const AggregatorRegistry = client.AggregatorRegistry;
+AggregatorRegistry.setRegistries(registry);
+// or for multiple registries:
+AggregatorRegistry.setRegistries([registry1, registry2]);
 ```
 
 #### Register
@@ -279,6 +298,22 @@ If you need to get a reference to a previously registered metric, you can use `r
 
 You can remove all metrics by calling `register.clear()`. You can also remove a single metric by calling
 `register.removeSingleMetric(*name of metric*)`.
+
+##### Cluster metrics
+
+You can get aggregated metrics for all workers in a node.js cluster with `register.clusterMetrics()`. This method both returns a promise and accepts a callback, both of which resolve with a metrics string suitable for Prometheus to consume.
+
+```js
+register.clusterMetrics()
+	.then(metrics => { /* ... */ })
+	.catch(err => { /* ... */ });
+
+// - or -
+
+register.clusterMetrics((err, metrics) => {
+	// ...
+});
+```
 
 #### Pushgateway
 
--- a/example/cluster.js
+++ b/example/cluster.js
@@ -1,0 +1,28 @@
+'use strict';
+
+const cluster = require('cluster');
+const express = require('express');
+const metricsServer = express();
+const AggregatorRegistry = require('../').AggregatorRegistry;
+const aggregatorRegistry = new AggregatorRegistry();
+
+if (cluster.isMaster) {
+	for (let i = 0; i < 4; i++) {
+		cluster.fork();
+	}
+
+	metricsServer.get('/cluster_metrics', (req, res) => {
+		aggregatorRegistry.clusterMetrics((err, metrics) => {
+			if (err) console.log(err);
+			res.set('Content-Type', aggregatorRegistry.contentType);
+			res.send(metrics);
+		});
+	});
+
+	metricsServer.listen(3001);
+	console.log(
+		'Cluster metrics server listening to 3001, metrics exposed on /cluster_metrics'
+	);
+} else {
+	require('./server.js');
+}
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -1,0 +1,172 @@
+'use strict';
+
+/**
+ * Extends the Registry class with a `clusterMetrics` method that returns
+ * aggregated metrics for all workers.
+ *
+ * In cluster workers, listens for and responds to requests for metrics by the
+ * cluster master.
+ */
+
+const cluster = require('cluster');
+const Registry = require('./registry');
+const util = require('./util');
+const aggregators = require('./metricAggregators').aggregators;
+
+const GET_METRICS_REQ = 'prom-client:getMetricsReq';
+const GET_METRICS_RES = 'prom-client:getMetricsRes';
+
+let registries = [Registry.globalRegistry];
+let requestCtr = 0; // Concurrency control
+const requests = new Map(); // Pending requests for workers' local metrics.
+
+class AggregatorRegistry extends Registry {
+	/**
+	 * Gets aggregated metrics for all workers. The optional callback and
+	 * returned Promise resolve with the same value; either may be used.
+	 * @param {Function?} callback (err, metrics) => any
+	 * @return {Promise<string>} Promise that resolves with the aggregated
+	 *   metrics.
+	 */
+	clusterMetrics(callback) {
+		const requestId = requestCtr++;
+
+		callback = callback || function() {};
+
+		return new Promise((resolve, reject) => {
+			const nWorkers = Object.keys(cluster.workers).length;
+
+			if (nWorkers === 0) {
+				return process.nextTick(() => {
+					callback(null, '');
+					resolve('');
+				});
+			}
+
+			const request = {
+				responses: [],
+				pending: nWorkers,
+				callback,
+				resolve,
+				reject,
+				errorTimeout: setTimeout(() => {
+					request.failed = true;
+					const err = new Error('Operation timed out.');
+					request.callback(err);
+					reject(err);
+				}, 5000),
+				failed: false
+			};
+			requests.set(requestId, request);
+
+			const message = {
+				type: GET_METRICS_REQ,
+				requestId
+			};
+			for (const id in cluster.workers) cluster.workers[id].send(message);
+		});
+	}
+
+	/**
+	 * Creates a new Registry instance from an array of metrics that were
+	 * created by `registry.getMetricsAsJSON()`. Metrics are aggregated using
+	 * the method specified by their `aggregator` property, or by summation if
+	 * `aggregator` is undefined.
+	 * @param {Array} metricsArr Array of metrics, each of which created by
+	 *   `registry.getMetricsAsJSON()`.
+	 * @return {Registry} aggregated registry.
+	 */
+	static aggregate(metricsArr) {
+		const aggregatedRegistry = new Registry();
+		const metricsByName = new util.Grouper();
+
+		// Gather by name
+		metricsArr.forEach(metrics => {
+			metrics.forEach(metric => {
+				metricsByName.add(metric.name, metric);
+			});
+		});
+
+		// Aggregate gathered metrics.
+		metricsByName.forEach(metrics => {
+			const aggregatorName = metrics[0].aggregator;
+			const aggregatorFn = aggregators[aggregatorName];
+			if (typeof aggregatorFn !== 'function') {
+				throw new Error(`'${aggregatorName}' is not a defined aggregator.`);
+			}
+			const aggregatedMetric = aggregatorFn(metrics);
+			// NB: The 'omit' aggregator returns undefined.
+			if (aggregatedMetric) {
+				const aggregatedMetricWrapper = Object.assign(
+					{
+						get: () => aggregatedMetric
+					},
+					aggregatedMetric
+				);
+				aggregatedRegistry.registerMetric(aggregatedMetricWrapper);
+			}
+		});
+
+		return aggregatedRegistry;
+	}
+
+	/**
+	 * Sets the registry or registries to be aggregated. Call from workers to
+	 * use a registry/registries other than the default global registry.
+	 * @param {Array<Registry>|Registry} regs Registry or registries to be
+	 *   aggregated.
+	 * @return {void}
+	 */
+	static setRegistries(regs) {
+		if (!Array.isArray(regs)) regs = [regs];
+		regs.forEach(reg => {
+			if (!(reg instanceof Registry)) {
+				throw new TypeError(`Expected Registry, got ${typeof reg}`);
+			}
+		});
+		registries = regs;
+	}
+}
+
+if (cluster.isMaster) {
+	// Listen for worker responses to requests for local metrics
+	cluster.on('message', (worker, message) => {
+		if (arguments.length === 2) {
+			// pre-Node.js v6.0
+			message = worker;
+			worker = undefined;
+		}
+
+		if (message.type === GET_METRICS_RES) {
+			const request = requests.get(message.requestId);
+			message.metrics.forEach(registry => request.responses.push(registry));
+			request.pending--;
+
+			if (request.pending === 0) {
+				// finalize
+				requests.delete(message.requestId);
+				clearTimeout(request.errorTimeout);
+
+				if (request.failed) return; // Callback already run with Error.
+
+				const registry = AggregatorRegistry.aggregate(request.responses);
+				const promString = registry.metrics();
+				request.callback(null, promString);
+				request.resolve(promString);
+			}
+		}
+	});
+} else if (cluster.isWorker) {
+	// Respond to master's requests for worker's local metrics.
+	process.on('message', message => {
+		if (message.type === GET_METRICS_REQ) {
+			process.send({
+				type: GET_METRICS_RES,
+				requestId: message.requestId,
+				metrics: registries.map(r => r.getMetricsAsJSON())
+			});
+		}
+	});
+}
+
+module.exports = AggregatorRegistry;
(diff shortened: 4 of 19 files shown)


tommiv:
We actually implemented aggregated metrics via https://github.com/prometheus/statsd_exporter and "push" model statsd driver for node, but it'll be nice to eliminate an additional entity. Thank you.

siimon:
Thanks for your PR :)

Im not against adding cluster support, the thing I am worried about is supporting it in the long term since I lack the experience and time to get up to speed with clustering node processes.

However, but I understand that it's something that probably should be in prom-client since a lot of people apparently cluster out the node processes, so I don't see any reason to not merge it when it looks ready.

A thought, would it be interesting to apply some kind of label to the metric to indicate which node in the cluster the metric is from?

tommiv:
> A thought, would it be interesting to apply some kind of label to the metric to indicate which node in the cluster the metric is from?
This is definetely a good idea, we used similar tag in all our setups.

zbjornson:
Thanks for the feedback. I'll work on cleaning it up. I'm happy to help with maintenance of this; I'm more familiar with the cluster module than I ever wanted to be, and I think most prod node.js servers use it.

As far as metric labels: This acts as an aggregator, so all worker processes appear as one. This seemed to be the preferred approach for the Python (gunicorn) and Rails clients; the concern discussed there was that exposing all workers separately would massively increase the number of time series, and percentage metrics like CPU and memory usage would be fractions of an unknown denominator.

Nonetheless, the example file this PR shows setting a metric label to a worker's ID, and the result is that it doesn't get aggregated, so you *can* keep a metric separate if you want to.

Does that seem reasonable? @tommiv satisfy your use case?

tommiv:
> Does that seem reasonable? @tommiv satisfy your use case?
I thought we're talking about a physical node address (or VM), not the cluster node id. But now I see I misread the previous message. We don't split up metrics to the cluster nodes, it makes no sense indeed.

example/cluster.js
```diff
@@ -0,0 +1,29 @@
...
+
+if (cluster.isMaster) {
+	for (let i = 0; i < 4; i++) {
+		cluster.fork();
+	}
+
+	metricsServer.get('/cluster_metrics', (req, res) => {
+		promCluster.clusterMetrics((err, metrics) => {
+			//eslint-disable-next-line no-console
+			if (err) console.log(err);
```
SimenB: Throw it?

lib/cluster.js
```diff
@@ -0,0 +1,211 @@
...
+}
+
+if (cluster.isWorker) {
+	// Respond to master's requests for worker's local metrics.
+	process.on('message', message => {
+		if (message.type === GET_METRICS_REQ) {
+			process.send({
+				type: GET_METRICS_RES,
+				requestId: message.requestId,
+				metrics: Registry.globalRegistry.getMetricsAsJSON()
```
SimenB: Registry should probably not be hard-coded

lib/cluster.js
```diff
@@ -0,0 +1,211 @@
...
+
+	/**
+	 * Creates a new Registry instance from an array of metrics that were
+	 * created by `registry.getMetricsAsJSON()`. Metrics are aggregated using
+	 * their `aggregate` method, or by summation if that is undefined.
+	 * @param {Array} metricsArr Array of metrics, each of which created by
+	 *     `registry.getMetricsAsJSON()`.
+	 * @return {Registry} aggregated registry.
+	 */
+	Registry.aggregate = function(metricsArr) {
```
SimenB: We have a `merge`, but that just includes different metrics and doesn't handle dupes?

If so, this function should be added next to merge, not live here

lib/cluster.js
```diff
@@ -0,0 +1,211 @@
...
+		// Not always defined: some are platform-specific.
+		if (metric) metric.aggregate = metricMergeFns[metricName];
+	}
+
+	/**
+	 * Gets aggregated metrics for all workers.
+	 * @param {Function} callback (err, metrics) => any
+	 * @return {undefined} undefined
+	 */
+	Registry.prototype.clusterMetrics = function(callback) {
```
SimenB: Don't modify the prototype. Either add directly or use inheritance

lib/cluster.js
```diff
@@ -0,0 +1,211 @@
...
+		v => v.reduce((p, c) => p + c.value, 0) / v.length
+	);
+	// const MIN = AggregatorFactory(v => v.reduce((p, c) => Math.min(p + c.value), 0));
+	// const MAX = AggregatorFactory(v => v.reduce((p, c) => Math.max(p + c.value), 0));
+
+	// Define merge strategies for all default metrics. User metrics may
+	// define an `aggregate` method, or will be summed by default.
+	// TODO These are here for the sake of isolation, but obviously would be
+	// nicer if they were defined in the metrics/ files.
+	// TODO Might be nice to have multiple aggregates for some, like the
```
SimenB: We already export them, right?

lib/cluster.js
```diff
@@ -0,0 +1,211 @@
...
+				metric.values.forEach(value => {
+					const key = util.hashObject(value.labels);
+					if (byLabels.has(key)) {
+						byLabels.get(key).push(value);
+					} else {
+						byLabels.set(key, [value]);
+					}
+				});
+			});
+			byLabels.forEach(values => {
```
SimenB: `result.values = byLabels.map` etc

lib/cluster.js
```diff
@@ -0,0 +1,211 @@
...
+			}
+		}
+	});
+
+	/**
+	 * Returns a new function that applies the `aggregatorFn` to the values.
+	 * @param {Function} aggregatorFn function to apply to values.
+	 * @return {Function} aggregator function
+	 */
+	const AggregatorFactory = function(aggregatorFn) {
```
SimenB: Does this factory have to be within the `if`? Reviewing on mobile, so not too easy seeing if you close over anything

README.md
```diff
@@ -6,6 +6,14 @@ A prometheus client for node.js that supports histogram, summaries, gauges and c
 
 See example folder for a sample usage. The library does not bundle any web framework, to expose the metrics just return the metrics() function in the registry.
 
+#### Usage with Node.js's `cluster` module
+
+Node.js's `cluster` module spawns multiple processes and hands off socket connections to those workers. Returning metrics from a worker's local registry will only reveal that worker's metrics, which is generally undesirable.
+
+To solve this, you can aggregate all of the workers' metrics in the master process. See example folder for a sample usage. By default, custom metrics are summed across workers, and default metrics use sensible defaults. To use a different aggregation method, define an `aggregate()` function for the metric's name in the global registry in the *master* process. (See the `AggregatorFactory` in `lib/cluster.js` for details.)
+
+Note: You must use the default global registry in cluster mode.
```
SimenB: Why?

example/cluster.js
SimenB: Meh, it's an example

example/cluster.js
```diff
@@ -0,0 +1,29 @@
...
+			res.set('Content-Type', promCluster.contentType);
+			res.send(metrics);
+		});
+	});
+
+	metricsServer.listen(3001);
+	//eslint-disable-next-line no-console
```
SimenB: Can just add a eslintrc in examples/ which disables this rule

.eslintrc
```diff
@@ -2,7 +2,8 @@
   "plugins": ["prettier"],
   "extends": "eslint:recommended",
   "env": {
-	"node": true
+	"node": true,
+	"es6": true
```
SimenB: For `Map`?

zbjornson: Yep! Based on the other new language features you're already using, I don't think this changes the oldest supported version of node.js.

zbjornson:
Thanks very much for the review. Will work this weekend on making it nice and removing the parts that were awkward as a result of trying to keep it isolated.

README.md
zbjornson: In the current arrangement, the cluster master is the one that fulfills the metrics request by querying all of the workers and aggregating their local metrics. The workers need to know *which* registry to send their local metrics from, so the global registry is easy. I'll try to figure out a nice way to support other registries.

lib/cluster.js
vidhill: He's using a Map, I don't thing you can call .map directly, 
he could convert it to an Array using Array.from an go from there.

zbjornson:
Most comments resolved and most tests added. Need a few more tests, and there are still some parts of this that I don't love and that I want to think about better ways of doing, so please hang on another day or so. (Namely: (1) The global registry still has to be used. (2) The way aggregators are looked up is sort of ugly.)

README.md
```diff
@@ -278,6 +288,10 @@ If you need to get a reference to a previously registered metric, you can use `r
 You can remove all metrics by calling `register.clear()`. You can also remove a single metric by calling
 `register.removeSingleMetric(*name of metric*)`.
 
+##### Cluster metrics
+
+You can get aggregated metrics for all worker processes in a cluster by running `register.clusterMetrics((err, metrics) => {...})`, which will be called with an error, or with the aggregated metrics string for prometheus to consume.
```
SimenB: @siimon do we want callback or promise api?

lib/cluster.js
```diff
@@ -0,0 +1,151 @@
...
+				);
+				aggregatedRegistry.registerMetric(aggregatedMetricWrapper);
+			}
+		});
+
+		return aggregatedRegistry;
+	}
+}
+
+module.exports = AggregatorRegistry;
```
SimenB: Can you move the export to the bottom?

README.md
SimenB: Since cluster api is very much callback based, keeping the cb might be natural

lib/metricAggregators.js
```diff
@@ -0,0 +1,82 @@
...
+		byLabels.forEach(values => {
+			if (values.length === 0) return;
+			const valObj = {
+				value: aggregatorFn(values),
+				labels: values[0].labels
+			};
+			if (values[0].metricName) {
+				valObj.metricName = values[0].metricName;
+			}
+			// TODO how do we aggregate timestamps? Average? Omit?
```
SimenB: either keep the max or omit, I think. I'm not sure what makes the most sense...

README.md
zbjornson: Easy enough to do both: return a promise and accept a callback.

siimon: Sorry, haven't had the time to look at this in detail. Having callbacks is a must have imo, but I'm fine with both aswell.

lib/histogram.js
```diff
@@ -64,6 +65,7 @@ class Histogram {
 
 		this.name = config.name;
 		this.help = config.help;
+		this.aggregate = config.aggregator || AGGREGATORS.SUM;
```
siimon: Couldn't you set this in the object assign step instead? Will be easier to see all the defaults if they are set on the same place

lib/metricAggregators.js
siimon: Yeah not sure either, I'm more leaning towards omitting, since I rather provide no data at all than misguiding data

lib/metricAggregators.js
```diff
@@ -0,0 +1,82 @@
...
+			result.values.push(valObj);
+		});
+		return result;
+	};
+};
+
+/**
+ * Functions that can be used to aggregate metrics from multiple registries.
+ */
+exports.AGGREGATORS = {
```
siimon: Im thinking these might belong in a separate file? Not sure about the uppercase function names either

.eslintrc
SimenB: We support node 4 and up, and `Map` has been in node since 0.12, so we're good 🙂

.eslintrc
```diff
@@ -55,6 +56,12 @@
 		"no-shadow": "off",
 		"no-unused-expressions": "off"
 	  }
+	},
+	{
+		"files": ["example/**/*.js"],
```
SimenB: 👍

README.md
SimenB: I prefer Promise these days because of `async-await`, which is in node 8 which will be LTS in a couple of months. Makes for cleaner code.

That said, I don't feel strongly about it

lib/cluster.js
```diff
@@ -0,0 +1,151 @@
...
+const AGGREGATORS = require('./metricAggregators').AGGREGATORS;
+
+const GET_METRICS_REQ = 'prom-client:getMetricsReq';
+const GET_METRICS_RES = 'prom-client:getMetricsRes';
+
+// Default metrics that don't use summation:
+const DEFAULT_METRIC_AGGREGATORS = {
+	nodejs_version_info: AGGREGATORS.FIRST,
+	process_start_time_seconds: AGGREGATORS.OMIT,
+	nodejs_eventloop_lag_seconds: AGGREGATORS.AVERAGE
```
SimenB: feels sorta icky to hard-code these. what about custom metrics? Can you filter out all non-sum metrics by checking the metric's type at aggregation-time?

SimenB: after doing `Registry.globalRegistry.getSingleMetric(metricName);` further down, figure out strategy. If `type` is not available, either do an `instanceof` or add a `type` function to the metrics

lib/cluster.js
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * In cluster workers, listens for and responds to requests for metrics by the
+ * cluster master.
+ */
+
+const cluster = require('cluster');
+const Registry = require('./registry');
+const AGGREGATORS = require('./metricAggregators').AGGREGATORS;
+
+const GET_METRICS_REQ = 'prom-client:getMetricsReq';
```
SimenB: Can this be a `Symbol` instead of a string? Not sure how serialization works here

lib/metricAggregators.js
```diff
@@ -0,0 +1,82 @@
...
+		if (metrics.length === 0) return;
+		const result = {
+			help: metrics[0].help,
+			name: metrics[0].name,
+			type: metrics[0].type,
+			values: []
+		};
+		// Gather metrics by labels.
+		const byLabels = new Map();
+		metrics.forEach(metric => {
```
SimenB: this part seems duplicated from `aggregate`. Can it be put in `utils`? (I realize the diff is `.name` vs `.labels`)

lib/metricAggregators.js
SimenB: I think they fit fine in here. I also find uppercase odd, though 🙂

lib/metrics/eventLoopLag.js
```diff
@@ -16,7 +17,8 @@ module.exports = registry => {
 	const gauge = new Gauge({
 		name: NODEJS_EVENTLOOP_LAG,
 		help: 'Lag of event loop in seconds.',
-		registers: registry ? [registry] : undefined
+		registers: registry ? [registry] : undefined,
+		aggregator: AGGREGATORS.AVERAGE
```
SimenB: Should the `Gauge` class itself set this? So all get it for free? Or is there some use-case where it makes sense to not do so?

SimenB: I suppose there are some special-cases (like `nodejs_version_info`), but can a sensible default be set?

lib/histogram.js
zbjornson: Originally I had it there, but then it has to be repeated where the 2nd constructor form (formal arguments) is handled:

```js
if (isObject(name)) {
  // once here
} else {
  config = {
    // again here
  }
}
```

lib/cluster.js
zbjornson: (Re: first comment) Yeah, this is my highest priority to clean up. These aggregation methods definitely shouldn't be defined twice; right now they're here and in `lib/metrics/xxx.js`. The factory functions exported from `lib/metrics/` don't make it easy to retrieve the metric's `aggregate`/`aggregator` value, so I had been playing with calling those factory functions with a mocked registry, but chaos ensued. To be continued...

lib/metrics/eventLoopLag.js
zbjornson: Everything defaults to `sum` via code in the `Gauge`/`Histogram`/`Summary`/`Counter` classes ([example](https://github.com/siimon/prom-client/pull/147/files#diff-565b1b7426b6a8327b446f794fbd5d5fR73)), but there are three metrics with non-default methods:
* `nodejs_version_info` uses the first value (all values will be the same)
* `process_starttime_seconds` is omitted (all values will be slightly different -- open to suggestions)
* `nodejs_eventloop_lag` averages

lib/metricAggregators.js
zbjornson: `AGGREGATORS.sum` or `AGGREGATORS.Sum` or `aggregators.sum`?

lib/cluster.js
```diff
@@ -0,0 +1,151 @@
...
+			request.pending--;
+
+			if (request.pending === 0) {
+				// finalize
+				requests.delete(message.requestId);
+				clearTimeout(request.errorTimeout);
+
+				if (request.failed) return; // Callback already run with Error.
+
+				const registry = Registry.aggregate(request.responses);
```
japrescott: this needs to refer to `AggregatorRegistry`.

example/cluster.js
```diff
@@ -0,0 +1,27 @@
...
+const metricsServer = express();
+const AggregatorRegistry = require('../').AggregatorRegistry;
+
+if (cluster.isMaster) {
+	for (let i = 0; i < 4; i++) {
+		cluster.fork();
+	}
+
+	metricsServer.get('/cluster_metrics', (req, res) => {
+		AggregatorRegistry.clusterMetrics((err, metrics) => {
```
japrescott: `clusterMetrics` is not a static method and thus can only be called on an Instance - of which there is none in your example code

D34THWINGS:
Can't wait to see this working, good job guys ! ❤️

zbjornson:
Sorry for the delay folks. Will be pushing an update this weekend.

lib/cluster.js
zbjornson: Not as far as I know, without using `Symbol.for(string)` (which I don't think is any better than just using the string).

lib/cluster.js
```diff
@@ -0,0 +1,146 @@
...
+	});
+} else if (cluster.isWorker) {
+	// Respond to master's requests for worker's local metrics.
+	process.on('message', message => {
+		if (message.type === GET_METRICS_REQ) {
+			process.send({
+				type: GET_METRICS_RES,
+				requestId: message.requestId,
+				// TODO see if we can support the non-global registry also.
+				metrics: Registry.globalRegistry.getMetricsAsJSON()
```
zbjornson: This is the last remaining TODO. The worker processes need to know *which* registry to send metrics from for aggregation, so using the `globalRegistry` was easy.

Options I can think of:
1. We send metrics from *all* registries. This seems like it would create a mess.
2. We provide a way for users to specify which registry to use. Something like `Registry.aggregationSource = new Registry()`, then the above line becomes `Registry.aggregationSource .getMetricsAsJSON()`.

I don't know what the use cases are for multiple registries, so I'm not sure what makes sense to do.

zbjornson:
Alright, I think I've addressed all the comments, including adding a Promise interface in addition to the callback interface. All the ugly code around storing and looking up metric aggregator functions is much nicer now.

Thanks everyone for the continuing feedback!

lib/cluster.js
SimenB: Yeah, just the string is fine then 🙂

lib/cluster.js
SimenB: 1. We have no way of knowing about every registry
1. I think we should do this. Unsure about the API, though. Maybe a `AggregatorRegistry.setRegistries`?

lib/cluster.js
```diff
@@ -0,0 +1,146 @@
...
+		const metricsByName = new util.Grouper();
+
+		// Gather by name
+		metricsArr.forEach(metrics => {
+			metrics.forEach(metric => {
+				metricsByName.add(metric.name, metric);
+			});
+		});
+
+		// Aggregate gathered metrics. Default to summation.
```
SimenB: is that default needed if all the metric types has a default?

SimenB: is that default needed if all the metric types has a default?

lib/util.js
```diff
@@ -89,3 +89,21 @@ exports.printDeprecationCollectDefaultMetricsNumber = timeout => {
 		`prom-client - A number to defaultMetrics is deprecated, please use \`collectDefaultMetrics({ timeout: ${timeout} })\`.`
 	);
 };
+
+class Grouper extends Map {
```
SimenB: 👍

test/aggregatorsTest.js
```diff
@@ -0,0 +1,91 @@
+'use strict';
+
+xdescribe('aggregators', () => {
```
SimenB: remove the `x`

siimon:
I agree, looks really good! Another small detail, update the changelog :)

lib/metricAggregators.js
```diff
@@ -0,0 +1,80 @@
...
+			name: metrics[0].name,
+			type: metrics[0].type,
+			values: []
+		};
+		// Gather metrics by labels.
+		const byLabels = new util.Grouper();
+		metrics.forEach(metric => {
+			metric.values.forEach(value => {
+				const key = util.hashObject(value.labels);
+				byLabels.add(key, value);
```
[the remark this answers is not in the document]
zbjornson: Thanks for pointing that out, will fix and add a test for it.

zbjornson:
All remaining bits addressed.

Note that the test output says *"Warning: Possible EventEmitter memory leak detected. 11 message listeners added. Use emitter.setMaxListeners() to increase limit"*. This is not a real leak (doesn't happen outside of jest); see https://github.com/facebook/jest/issues/4413. If you want I can add a jest config file and a setup file containing `emitter.setMaxListeners(0)`.

Thank you!

siimon:
I agree! Great work and it's highly appreciated! I'll merge and publish a new minor!

And thanks @SimenB et al for great reviews!

zbjornson:
Thanks everyone!

PR #148, opened 2017-07-30
Fixed typing of DefaultMetricsCollectorConfiguration in definitions file

zhenwenc:
Fixed mismatch type definition. I have checked that the corresponding test case is using `register` instead of `registry`.


siimon:
Thanks 👍

PR #149, opened 2017-08-18
Include invalid values in the error messages

tothandras:
`Error` only expects one argument, the `message`, which means that the invalid values were lost in the second argument.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 ## [Unreleased]
 ### Breaking
 ### Added
+- Include invalid values in the error messages
+
 ### Changed
 
 ## [10.0.3] - 2017-08-07
--- a/lib/counter.js
+++ b/lib/counter.js
@@ -2,9 +2,10 @@
  * Counter metric
  */
 'use strict';
+
+const util = require('util');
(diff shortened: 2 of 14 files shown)


tothandras:
@SimenB Thanks for the review, I'll update the PR.

tothandras:
@SimenB is it a feature that `util.isNumber` returns `true` for `3ms` for example, but when adding it to a summary it's not the parsed value that is used, but the string?

SimenB:
That's a bug. `parseFloat('3ms')` returns `3`, which should be used

siimon:
I'd say that is a bug as well and I think the best thing would be to throw an error.

The PR looks good, so fix the issues @SimenB pointed out and it's ready to be merged!

tothandras:
@SimenB updated

SimenB:
Too bad `toThrowErrorMatchingSnapshot` doesn't includes the type of error, just message. No problem, though :)

SimenB:
@siimon thoughts?

siimon:
LGTM! 👍

SimenB:
@tothandras could you update the changelog as well? Then we'll get it released :smile:

tothandras:
@SimenB like this?

SimenB:
Perfect! I'll merge and release within the hour.

SimenB:
A bit slow, but released as 10.0.4 now. Thanks!

PR #153, opened 2017-09-22
Bugfix - cannot get cluster info in node version < 6

riggerthegeek:
The `cluster.on('message')` function doesn't have the worker as the first argument in node < 6. To do this, the official way is to check the number of arguments received. However, because this is using a fat arrow function (`() => {}`), the variable `arguments` is hoisted outside the scope so the actual arguments
has a length of 5.

If you run the old version of the code in [Babel](http://bit.ly/2wbApcb) this shows the problem:
```javascript
cluster.on('message', (worker, message) => {
  if (arguments.length === 2) {
      // pre-Node.js v6.0
      message = worker;
      worker = undefined;
  }
});
```

This gets compiled to:
```javascript
var _arguments = arguments;
cluster.on('message', function (worker, message) {
    if (_arguments.length === 2) {
        // pre-Node.js v6.0
        message = worker;
        worker = undefined;
    }
});
```

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -130,7 +130,7 @@ class AggregatorRegistry extends Registry {
 
 if (cluster.isMaster) {
 	// Listen for worker responses to requests for local metrics
-	cluster.on('message', (worker, message) => {
+	cluster.on('message', function(worker, message) {
 		if (arguments.length === 2) {
 			// pre-Node.js v6.0
 			message = worker;


riggerthegeek:
By the by, I'd normally use `(...args) => {}`, but this isn't available in Node < 6

zbjornson:
Oh wow, I didn't know `arguments` is hoisted in arrow functions. Thanks for explaining!

riggerthegeek:
Yeah, it's a bit of pain when you don't know it. In most cases, you should use ES6 destructing `(...args)` but this was only introduced in Node 6

siimon:
Thanks! 👍

riggerthegeek:
No probs. Any chance you could do an npm release this week as this bug is a blocker at work please?

siimon:
Ofcourse, I'm just waiting for #154

SimenB:
Out in 10.1.1

Issue #145, opened 2017-07-27
Clearing metrics data without removing the metric


mwho:
Every time my metric is updated, I'd like to clear all existing data before doing the update. This is because the existing data might be stale, and I want to make sure that the data associated with the metric is fresh.

It seems like the only way to do this is something like the following - removing the metric and then creating it again.

```
function updateMyGauge(x,y,z,val) {
    register.removeSingleMetric('myMetric');
    myGauge = new promClient.Gauge({name: 'myMetric',
                                           help: 'myMetric_help',
                                           labelNames: ['label1',
                                                        'label2',
                                                        'label3']});
     myGauge.set({label1: x, label2: y, label3: z}, val)
}
```

Is there a better way? Otherwise, it'd be nice to have a method `register.clearMetricData` that clears all labels to value mappings but keeps the metric definition itself intact.

siimon:
I've not used timestamps in prometheus myself, but that might be a way forward

SimenB:
Using timestamps should work. I think it then has a 5 minute window before being discarded. Is that enough?

mwho:
The sense I'm getting from reading the Prometheus forums (i.e. https://github.com/prometheus/client_golang/issues/187, https://github.com/prometheus/collectd_exporter/issues/24) is that specifying timestamps for metrics is discouraged. The Prometheus philosophy seems to be that the client side endpoint should always provide truthful and up to date data at time of scrape.

SimenB:
This was fixed in #157

PR #157, opened 2017-10-12
reset Metrics

donnut:
I needed a way to reset metrics, without removing the metric instances, see also #145. 
This solution is not completely tested, but I would like to know if this is a right way to go.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,9 @@ counter.inc(); // Inc with 1
 counter.inc(10); // Inc with 10
 ```
 
-A counter can be reset manually.
+A counter can be reset manually. This removes the label-values combinations and
+initializes to 0.
+
 ```js
 counter.reset();
 counter.inc(); // Inc with 1 starting from 0
@@ -123,7 +125,8 @@ gauge.inc(10); // Inc with 10
 gauge.dec(); // Dec with 1
 gauge.dec(10); // Dec with 10
 ```
-A gauge can be reset manually.
+A gauge can be reset manually. This removes the label-values combinations and
+initializes to 0.
 
 ```js
 gauge.reset();
--- a/README.md
+++ b/README.md
@@ -104,6 +104,12 @@ counter.inc(); // Inc with 1
 counter.inc(10); // Inc with 10
 ```
 
+A counter can be reset manually.
+```js
+counter.reset();
+counter.inc(); // Inc with 1 starting from 0
+```
+
 #### Gauge
 
 Gauges are similar to Counters but Gauges value can be decreased.
@@ -116,6 +122,12 @@ gauge.inc(); // Inc with 1
 gauge.inc(10); // Inc with 10
 gauge.dec(); // Dec with 1
 gauge.dec(10); // Dec with 10
+```
+A gauge can be reset manually.
+
+```js
+gauge.reset();
+gauge.inc(); // Inc with 1 starting from 0
 ```
 
 There are some utilities for common use cases:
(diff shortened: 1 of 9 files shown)


test/registerTest.js
```diff
@@ -204,6 +204,49 @@ describe('register', () => {
 		expect(output).toEqual(metric);
 	});
 
+	it('should allow resetting a counter', () => {
```
SimenB: the tests should be in their respective as well

test/registerTest.js
```diff
@@ -204,6 +204,49 @@ describe('register', () => {
...
+			labelNames: ['serial', 'active']
+		});
+		register.registerMetric(metric);
+
+		metric.inc({ serial: '12345', active: 'yes' }, 12);
+		expect(metric.get().values[0].value).toEqual(12);
+		expect(metric.get().values[0].labels.serial).toEqual('12345');
+		expect(metric.get().values[0].labels.active).toEqual('yes');
+
+		register.resetMetrics('serial');
```
SimenB: `'serial'`? Doesn't seem like the method takes an argument?

lib/registry.js
```diff
@@ -101,6 +101,12 @@ class Registry {
 		this._defaultLabels = labels;
 	}
 
+	resetMetrics() {
+		for (const metric in this._metrics) {
```
SimenB: `this._metrics.forEach(...`

test/registerTest.js
SimenB: `registerTest` can have a single tests with all metric types for the coverage 🙂

lib/counter.js
```diff
@@ -116,6 +124,14 @@ class Counter {
 	}
 }
 
+const reset = function() {
+	this.hashMap = {};
```
siimon: Might be a good idea to call reset in the constructor instead of having duplicate code

lib/gauge.js
```diff
@@ -226,6 +234,14 @@ function set(labels) {
 	};
 }
 
+function reset() {
+	this.hashMap = {};
+
```
siimon: Same as in the counter, call this function from the constructor

test/counterTest.js
```diff
@@ -240,4 +240,42 @@ describe('counter', () => {
 			expect(instance.get().values[0].timestamp).toEqual(undefined);
 		});
 	});
+	describe('counter reset', () => {
+		xit('should reset labelless counter', () => {
```
siimon: Why is this test disabled?

test/gaugeTest.js
```diff
@@ -375,6 +375,44 @@ describe('gauge', () => {
 			});
 		});
 	});
+	describe('gauge reset', () => {
+		xit('should reset labelless gauge', () => {
```
siimon: Disabled test again, enable or remove :)

README.md
```diff
@@ -104,6 +104,12 @@ counter.inc(); // Inc with 1
 counter.inc(10); // Inc with 10
 ```
 
+A counter can be reset manually.
```
siimon: Maybe we could add what reset means, ie removing all label-value combinations and initialize to 0

donnut: Good point. I'll extend the text.

lib/counter.js
donnut: Ok

test/counterTest.js
donnut: By accident.

test/gaugeTest.js
donnut: Same thing

donnut:
@SimenB I'm not that familair with `Histogram` and `Summary` yet, but I was under the impression that the current `reset` for both are already implemented. I added a test in registryTest that call their reset methods.

SimenB:
Oh, would you look at that, haha! It's not documented though, would you mind adding it to the readme? 🙂

siimon:
Thanks! 👍

SimenB:
Released as 10.2.0

PR #160, opened 2017-10-19
Update the Registry typedef with `contentType`

r-k-b:
bringing the type definitions inline with [the code](https://github.com/siimon/prom-client/blob/ee4c09322f6ffa3785b18a24cca6be4516f40140/lib/registry.js#L110)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -21,6 +21,11 @@ export class Registry {
 	metrics(opts?: MetricsOpts): string;
 
 	/**
+	 * The content-type that the metrics should be served with
+	 */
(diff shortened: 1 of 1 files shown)


r-k-b:
nvm, didn't see it was already in there.

SimenB:
Is it somehow wrong? Asking since my guess is that you came with the fix because your IDE didn't tell you about the property?

r-k-b:
Exactly! No it's not wrong, I failed to see prom-client in `npm outdated`, so I thought I had the latest version. 🤦‍♂️

PR #162, opened 2017-10-27
Only resolve/reject `clusterMetrics` promise if no callback is provided

zbjornson:
Small behavior change: Don't resolve/reject the promise returned by `clusterMetrics` if a callback is provided to that function. This prevents `UnhandledPromiseRejectionWarnings` from occurring if you're using callbacks.


SimenB:
Published in 10.2.1

PR #163, opened 2017-10-31
fixed invalid memory report on linux

yurijmikhalevich:
You have to multiply memory reported by the `/proc/self/status` by 1024, not 1000; because it's in KiB.

Proof:
```
$ ps -o vsz -p 15301
   VSZ
 13820
$ grep VmSize /proc/15301/status
VmSize:    13820 kB
$ man ps | grep vsz             
       %z     vsz      VSZ
                             vsz and rss.
       vsize       VSZ       see vsz.  (alias vsz).
       vsz         VSZ       virtual memory size of the process in KiB (1024-byte
```

Replace 15301 with the PID of your choice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 ## [Unreleased]
 ### Breaking
 ### Changed
+- Fixed invalid process_virtual_memory_bytes reported under linux
 ### Added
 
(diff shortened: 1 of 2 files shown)


SimenB:
Could you update the changelog?

yurijmikhalevich:
@SimenB done.

SimenB:
Released in 10.2.2

yurijmikhalevich:
Thanks!

PR #170, opened 2018-02-27
Cluster metrics aggregation ignores default labels

doochik:
registry.getMetricsAsJSON should merge metric labels and registry default labels

### DIFF
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -82,7 +82,21 @@ class Registry {
 	}
 
 	getMetricsAsJSON() {
-		return this.getMetricsAsArray().map(metric => metric.get());
+		return this.getMetricsAsArray().map(metric => {
+			const item = metric.get();
+			if (!item.values) {
+				return item;
+			}
+
+			item.values = item.values.map(val =>
+				// Avoid mutation and merge metric labels with registry default labels
+				Object.assign({}, val, {
+					labels: Object.assign({}, this._defaultLabels, val.labels)
+				})
+			);
+
+			return item;
+		});
 	}
 
 	removeSingleMetric(name) {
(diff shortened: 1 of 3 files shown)


lib/registry.js
```diff
@@ -82,7 +82,14 @@ class Registry {
 	}
 
 	getMetricsAsJSON() {
-		return this.getMetricsAsArray().map(metric => metric.get());
+		return this.getMetricsAsArray().map(metric => {
+			const item = metric.get();
+			(item.values || []).forEach(val => {
+				// Merge metric labels with registry default labels
+				val.labels = Object.assign({}, this._defaultLabels, val.labels);
```
SimenB: this mutates, which is not ok

SimenB:
Could you add a test for this?

doochik:
@SimenB add test for `getMetricsAsJSON` and fix mutation

lib/registry.js
```diff
@@ -82,7 +82,21 @@ class Registry {
...
 
 	getMetricsAsJSON() {
-		return this.getMetricsAsArray().map(metric => metric.get());
+		return this.getMetricsAsArray().map(metric => {
+			const item = metric.get();
+			if (!item.values) {
+				return item;
+			}
+
+			item.values = item.values.map(val =>
```
SimenB: this still mutates `item.values`. Should probably do a `JSON.parse(JSON.stringify(item))` or something

doochik: metric.get returns new object always, array.map returns new array. So there is no mutation here.

If you insist i'll change this :)

SimenB: Ah, `get` returns a new object. Then this is fine, thanks!

SimenB:
Could you update the changelog?

doochik:
done

siimon:
lgtm :)

doochik:
Can you release new version with this patch?

PR #171, opened 2018-02-28
Use base units (seconds) for current time in gauge

### LINKED ISSUE #167: Timestamps should be in seconds, eg in setToCurrentTime methods
The [implementation of setToCurrentTime](https://github.com/siimon/prom-client/blob/ee4c09322f6ffa3785b18a24cca6be4516f40140/lib/gauge.js#L170-L179) for gauges uses the value of `Date.now()` which is in milliseconds. The suggested conventions for client libraries is to use base units for all recorded metrics.

ThomWright:
I think this is the only instance where this is incorrect.

Breaking change

Fixes #167

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Breaking
 
+* Fixed `gauge.setToCurrentTime()` to use seconds instead of milliseconds
+  * This conforms to Prometheus [best practices](https://prometheus.io/docs/practices/naming/#base-units)
+
 ### Changed
 
 ### Added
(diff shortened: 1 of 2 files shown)


SimenB:
Mind updating the changelog as well? 🙂

ThomWright:
Done!

ThomWright:
Any chance this could get merged/released soon?

SimenB:
We could use this to drop node 4 support as it's EOL in one month, but nothing other than that I think.

siimon:
Yeah sounds like a good plan! I'll fix that!

kamalmarhubi:
What does dropping node 4 support mean? That it's not a semver breaking change if prom-client begins using features not present in node 4? Do you anticipate any changes like that?

siimon:
It might (most likely) work for a long time with node 4 too. But we want to make it obvious where we dropped support for a specific version.

Issue #181, opened 2018-03-26
AggregatorRegistry assumes all workers will have metrics setup


orestis:
Hello,

thanks for prom-client, it's a huge time saver for us!

I have run into an issue with the cluster aggregator support. It seems that unless *all* the workers of a node cluster are setup with prom-client, the aggregator won't work.

The issue seems to in https://github.com/siimon/prom-client/blob/bb09c6dea7941f54aa7a62fcfa5f052af9d00d5a/lib/cluster.js#L41-L74

We have workers in our node cluster that do not import prom-client for various reasons, hence I need a way to instruct prom-client to not bother contacting them.

I'm happy to provide a PR for this, but wanted to present possible solutions first:

1. Instead of rejecting the promise when a timeout happens, silently ignore the timeout. 
2. Same as (1) but report non-responsive workers as a separate metric for monitoring.
3. Instead of sending the `GET_METRICS_REQ` to *all* the workers, maintain a list of workers that have setup the listeners and send only to those.

Thoughts?

SimenB:
@zbjornson thoughts?

zbjornson:
Interesting situation...

Re: your possible solutions:

1. This would silently create wrong metric values where you were expecting workers to respond. E.g. if you were summing CPU or memory usage, you'd see artificial dips that are actually because a worker didn't report.

2. Seems like a hassle to monitor this auxiliary metric...

3. Possibly, depending on how it's done. If you did this so that on boot the workers message the master to indicate that they opt-in to metrics, users of this lib would have to ensure that they construct the `AggregatorRegistry` *before* forking workers. The opposite would be if there was a discovery phase (default to, say, one minute?) during which the master polls workers to see if they've setup listeners, and if they never issue a response during the discovery phase, don't poll them in the future.

I'll add another option:

4. (1) but opt-in, something like `registry.clusterMetrics({allowPartialAggregation: true})`.

orestis:
I've done a quick implemention for (3) here: https://github.com/siimon/prom-client/pull/182 -- not ready to merge yet but just to foster discussion.

I didn't think of the ordering issues (create registry *then* fork vs opposite). I'll think about it.

orestis:
I'm trying to think how can this be done in a way that the consumer of the library has control over the initialisation process, so thinking a bit aloud here:

[...]

Issue #180, opened 2018-03-26
defaultMetrics should be collected synchronously


brian-brazil:
As noted in https://prometheus.io/docs/instrumenting/writing_exporters/#scheduling collection should be synchronous with a scrape, and not get updated in a background thread. The defaultMetrics should be called at a scrape, not with a setInterval.

SimenB:
I didn't know about that, thanks for pointing it out!

Changing should be easy enough, we just have to either make `.metrics()` return a promise or add a separate method to update the registry whilst keeping `metrics()` synchronous. The update itself still has to be async since we read from `/proc` (sync IO is a big no-no in Node because of the single threaded model).

mirague:
Could the synchronous approach not be an addition (backwards compatibility), with being able to disable the background polling?

siimon:
Why would it be a breaking change? Sure the response will be different for each time you make a request but I find it hard to classify that as a breaking change.

If it is a breaking change I still think it should just be implemented and not get unnecessary technical debt.

sam-github:
It looks pretty breaking to me. Typical use looks like:
```
  var prom = require('prom-client')
  var register = prom.register;
  prom.collectDefaultMetrics({timestamps: false});
  app.get('/metrics', (req, res) => {
    res.set('Content-Type', register.contentType);
    res.end(register.metrics());
  });
```

If `.metrics()` kicks off an async read of /proc/whatever, then what is it going to return? The state as-of the last time `.metrics()` was called? I guess that's possible, but not fabulous.

Some options I can think of, all of which have downsides:
1. Always return current known data from `.metrics()`, but kick off a recollect -- so data is always a little bit stale. Then again, with current interval based approach, metrics are always a bit stale anyhow ATM, depending on how the prometheus scrape interval ends up interacting with the internal scrape interval.
2. Add a new async API `.collect((metrics) => res.end(metrics))` so that collection is async -- I like this API, but it does leave all current users with stale metrics, and it might also cause cruft in that there will be extra code to support both async and sync APIs
3. Other options?

(.collect could be called .metrics and be sync or async depending on whether it receives a cb)

I'm not sure how accepting of a semver-major you all are. Making `.metrics()` async and cleaning out the internals is best for maintainers, but it requires all users to update their code.

zbjornson:
I'm in favor of <span>#</span>2 also. Rather only provide one API and have it be the Right Way. 🚀

sam-github:
So, did a bit of poking around

[...]

Issue #178, opened 2018-03-26
Service discovery is the responsibility of Prometheus


brian-brazil:
https://github.com/siimon/prom-client#default-labels-segmented-by-registry indicates that this library has functionality to allow setting target labels. This is an anti-pattern in Prometheus as that's the role of service discovery/relabelling, and a client library should not facilitate this anti-pattern. This functionality should be removed.

SimenB:
It came from #83 (feature request) and #137 (PR), cc @goofballLogic @nwest1 thoughts?

nwest1:
Thanks for the mention - 

Bottom-line, this simplified the code necessary to emit metrics. If it was removed, it would mean more complexity and repetition in our own code, and no (easy) ability to label default metrics in a way we would expect. 

Due to timelines/prioritization this was a feature we found useful - especially when not having access to modify prometheus config. 

Would a note about the recommended approach be sufficient here without a full removal?

brian-brazil:
> Would a note about the recommended approach be sufficient here without a full removal?

This is such a common misunderstanding of the Prometheus architecture that it should be removed. You should switch to relabelling for this.

benny-medflyt:
I would like to point out that there is a legitimate use case for using `prom-client`'s default labels feature:

When using node cluster node, it can be used to apply a "worker" label with the worker id. This cannot be done outside of `prom-client` using regular prometheus relabeling.

All of the official prometheus client libraries are for languages that are multi-threaded (rather than multi-process), so they don't need this feature.

forresthopkinsa:
Is there support for the [multi-target pattern](https://prometheus.io/docs/guides/multi-target-exporter/) in this library? It doesn't seem like there's any way to pass parameters *into* exporters with this client, and that can be a blocker for certain setups

jdmarshall:
We had to roll our own for OTEL to handle PM2 instance-IDs to avoid court clobbering. I agree that process per core languages struggle with disambiguation.  #687 speeds this code up quite a bit. I’m happy to double down on this feature. 

If the committee thinks that default labels are an antipattern then the committee is wrong. Committees often are (Design By Committee). They can get stuffed. Prometheus and OTEL are both actively hostile to NodeJS, Bun. Ruby, Python and they didn’t need to be. But here we are doing our best.

jdmarshall:
Achtually...  The current page on Prometheus naming and labeling does not mention any of the concerns listed in this bug report.  

https://prometheus.io/docs/practices/naming/

[...]

Issue #177, opened 2018-03-26
Only custom collectors should be able to expose timestamps


brian-brazil:
https://github.com/siimon/prom-client#timestamps indicates that direct instrumentation can specify timestamps, and other code indicates that a /metrics will always have timestamps.

This makes no sense semantically, and breaks ingestion, timestamps, and staleness on the Prometheus end. All ability to set timestamps with direct instrumentation should be removed, and there should be no timestamps on /metrics in normal circumstances.

brian-brazil:
It looks like this client does not support custom collectors as it doesn't support collectors (it's all metric based), so as it stands there's no need for any timestamp-related functionality.

SimenB:
Oh, I didn't know that!

What's the role of collectors? Is it something we should support?

brian-brazil:
This is a bug report, this issue is actively causing problems.

> What's the role of collectors? Is it something we should support?

They're mainly needed for writing exporters, though they can also come up in applications sometimes. It's up to you if you want to support them.

SimenB:
What issues is it causing? It's possible to strip the timestamp from the textual representation (see https://github.com/siimon/prom-client#register)

Not to say I'm against removing them, but "actively causing problems" should at least be possible to work around until we remove it.

brian-brazil:
It's causing duplicate timestamp errors at the least. It's also going to be causing quite a few problems when trying to use affected metrics via PromQL due to staleness handling.

SimenB:
#333

PR #182, opened 2018-03-26
ensure that only registered cluster workers are asked to report metrics

orestis:
This is a PR related to #181, as start of discussion. 

Tests don't pass yet as I used spaces instead of tabs :)

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -27,36 +27,34 @@ class AggregatorRegistry extends Registry {
 	constructor() {
 		super();
 		addListeners();
-    this.aliveWorkers = new Set();
-    let that = this;
+		this.aliveWorkers = new Set();
+		const that = this;
 		cluster.on('message', function(worker, message) {
 			if (arguments.length === 2) {
 				// pre-Node.js v6.0
 				message = worker;
 				worker = undefined;
 			}
-      if (message.type === REG_METRICS_WORKER) {
-        let workerId = message.workerId;
-        console.log("metrics registering workerId", workerId);
-        that.aliveWorkers.add(workerId);
-        console.log(that.aliveWorkers);
-      }
-    });
-  }
+			if (message.type === REG_METRICS_WORKER) {
+				const workerId = message.workerId;
+				that.aliveWorkers.add(workerId);
+			}
+		});
+	}
 
 	/**
 	 * Gets aggregated metrics for all workers. The optional callback and
 	 * returned Promise resolve with the same value; either may be used.
 	 * @param {Function?} callback (err, metrics) => any
 	 * @return {Promise<string>} Promise that resolves with the aggregated
-	 *   metrics.
+	 *	 metrics.
 	 */
 	clusterMetrics(callback) {
 		const requestId = requestCtr++;
-    let that = this;
+		const that = this;
 
 		return new Promise((resolve, reject) => {
-      const nWorkers = that.aliveWorkers.size;
+			const nWorkers = that.aliveWorkers.size;
 
 			function done(err, result) {
 				// Don't resolve/reject the promise if a callback is provided
@@ -74,7 +72,7 @@ class AggregatorRegistry extends Registry {
 
 			const request = {
 				responses: [],
-        workerCount: nWorkers,
+				workerCount: nWorkers,
 				pending: nWorkers,
 				done,
 				errorTimeout: setTimeout(() => {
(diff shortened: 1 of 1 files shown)


lib/cluster.js
```diff
@@ -25,7 +27,22 @@ class AggregatorRegistry extends Registry {
...
+				message = worker;
+				worker = undefined;
+			}
+      if (message.type === REG_METRICS_WORKER) {
+        let workerId = message.workerId;
+        console.log("metrics registering workerId", workerId);
+        that.aliveWorkers.add(workerId);
+        console.log(that.aliveWorkers);
+      }
+    });
```
zbjornson: Instead of adding another `cluster.on("message")` listener, can you do this in `addListeners`?

lib/cluster.js
```diff
@@ -71,7 +90,18 @@ class AggregatorRegistry extends Registry {
...
+        let worker = cluster.workers[id];
+        if (worker === undefined || worker.isDead() || !worker.isConnected()) {
+          failedWorkers.add(id);
+          request.pending --;
+          request.workerCount --;
+        } else {
+          worker.send(message)
+        }
+      }
+      that.aliveWorkers = new Set([...that.aliveWorkers].filter(x => !failedWorkers.has(x)));
```
zbjornson: Can you just do `this.aliveWorkers.remove(...)` in the loop, instead of making a new `Set` instance each time?

zbjornson (changes requested):
The ordering issue (`new AggregatorRegistry()` vs. `cluster.fork()`) is a breaking change in this PR's current state. It might be sufficient for the worker to retry sending the init message until the master acknowledges, for up to some amount of time. Sort of adds a lot of code...

Didn't comment on the cleanup stuff (stray console.log, linter stuff).

lib/cluster.js
orestis: I'll have to check the semantics of mutating the Set while iterating, but yes will change.

lib/cluster.js
orestis: To do that I might have to move the aliveListeners to be a module-level variable. Would that be ok?

lib/cluster.js
```diff
@@ -10,20 +10,31 @@
...
 const requests = new Map(); // Pending requests for workers' local metrics.
 
 class AggregatorRegistry extends Registry {
-	constructor() {
+	/**
+     Create an AggregatorRegistry instance. Accepts an optional `options` object.
+
+     Options are:
+        coordinated If false (default), request metrics from all cluster workers. If true, request metrics only from workers that have required prom-client.
+	 * @param {object?} options object
```
zbjornson: Is the only reason this *needs* to be an option because setting it to `true` makes the order of forking vs. `new AggregatorRegistry()` matter? I'd rather go for an implementation that isn't sensitive to that, e.g. the worker repeatedly attempts to register with the master until the master acks the registration.

Having a heterogeneous pool of workers (not all of them setting up prom-client) is unusual, but I think the behavior achieved when this is `true` is what should happen by default.

lib/cluster.js
```diff
@@ -167,9 +193,19 @@ function addListeners() {
 					if (request.failed) return; // Callback already run with Error.
 
 					const registry = AggregatorRegistry.aggregate(request.responses);
+					const g = new Gauge({
+						name: 'nodejs_prom_client_cluster_workers',
+						help: 'Number of connected cluster workers reporting to prometheus',
+						registers: [registry]
+					});
+					g.set(request.workerCount);
```
zbjornson: Might want to move this new metric to a separate PR. (If @siimon and @SimenB are okay with it, it's a simple change that could land before this PR is ironed out.)

orestis: Noted, however this is very specific to the coordinated option - perhaps it should be exposed only when coordinated is true. Perhaps a similar metric could be added to the default metrics collected that counts *all* the cluster workers?

lib/cluster.js
orestis: Yes, the only reason to make it an option is that it would break backwards compatibility. 

Given that a metrics client should be as unobtrusive as possible, I would be wary of adding a prolonged discovery phase. I would certainly prefer in my consuming codebase to keep things simple and accept that ordering matters. 

In the case of the homogenous cluster, the previous  implementation cleanly sidesteps a lot of the issues with “garbage collecting” the workers and nobody else had complained so far :)

SimenB: FWIW, we will be breaking backwards compatibility pretty hard soon ish (see #177, #178 and #180), so don't let semver stop you from writing the code you'd like to write :D

zbjornson:
(Sorry for the delay, busy week. Will try to send more feedback later today.)

SimenB:
Any news here?

zbjornson:
Sorry to drop the ball.

I'm sorta hesitant to add the complexity to support an unusual use case (heterogeneous workers in a cluster), but this approach works.

My preference would be a version that isn't sensitive to the order of `new AggregatorRegistry()` vs. `cluster.fork()` in the client code, but since that only matters if you opt-in to `coordinated` and it's an advanced usage scenario, those advanced users could be expected to understand/deal with that requirement.

PR also still needs some formatting/linting issues addressed.

jdmarshall:
See #789

PR #184, opened 2018-04-06
Making global registry into singleton

pontusarfwedson:
This PR makes the 'prom-client.registry' call always give the user the same instance of a global registry. This allows the user to in completely different files do 
```
const defaultRegistry = client.register;
```
to get the default registry. If that registry then is passed into a metric definition as 
```
const errorVec = new client.Counter({
  name: 'failed__total',
  help: 'Number of times failed',
  registers: [defaultRegistry]
});
```
then it will be added to the registry regardless in which file the `require` has been performed. Previously every `client.register` created a new by running `new Registry() `. 
This allows for more similar usage as the golang lib which where the client can be imported and if you run `MustRegister(metric)` anywhere then it will go onto the default registry.

### DIFF
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -138,5 +138,23 @@ class Registry {
 	}
 }
 
+var GlobalRegistry = (function () {
+    var reg;
+
+    function createInstance() {
+        var reg = new Registry()
+        return reg;
+    }
+
+    return {
+        getInstance: function () {
+            if (!reg) {
+                reg = createInstance();
+            }
+            return reg;
+        }
+    };
+})();
+
 module.exports = Registry;
-module.exports.globalRegistry = new Registry();
+module.exports.globalRegistry = GlobalRegistry.getInstance()
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -138,21 +138,21 @@ class Registry {
 	}
 }
 
-var GlobalRegistry = (function () {
-    let reg;
+const GlobalRegistry = (function() {
+	let reg;
 
-    function createInstance() {
-        return new Registry();
-    }
+	function createInstance() {
+		return new Registry();
+	}
 
-    return {
-        getInstance: function () {
-            if (!reg) {
-                reg = createInstance();
-            }
-            return reg;
-        }
-    };
+	return {
+		getInstance() {
+			if (!reg) {
+				reg = createInstance();
+			}
+			return reg;
+		}
+	};
 })();
 
 module.exports = Registry;
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -139,11 +139,10 @@ class Registry {
 }
 
 var GlobalRegistry = (function () {
-    var reg;
+    let reg;
 
     function createInstance() {
-        var reg = new Registry()
-        return reg;
+        return new Registry();
     }
 
     return {
@@ -157,4 +156,4 @@ var GlobalRegistry = (function () {
 })();
 
 module.exports = Registry;
-module.exports.globalRegistry = GlobalRegistry.getInstance()
+module.exports.globalRegistry = GlobalRegistry.getInstance();


SimenB:
Because of module caching, it's already a singleton, so this PR shouldn't make any difference. Mind adding a test, or at least some example code, showing that this is necessary?

pontusarfwedson:
Hi SimenB! Just attempted to showcase it with an example but realized that it actually works the way you describe. However, there is a use case in which I appear to encounter the issue. If I have a server which defines some routes where we among others have a metrics route by doing
```
const collectDefaultMetrics = client.collectDefaultMetrics;
const defaultRegistry = client.register
// Probe every 5th second.
collectDefaultMetrics({ timeout: 5000 });
```
and 
```
path: '/sys/metrics',
    config: {
        handler: async function (request, h) {

      return h.response(defaultRegistry.metrics()).header('Content-Type', defaultRegistry.contentType).code(200)
    }
```
If we then import another module made by ourselves to get some additional functionality and that module also implements
```
const defaultRegistry = client.register;
```
as well as 
```
const errorVec = new client.Counter({
  name: 'failed__total',
  help: 'Number of times failed',
  registers: [defaultRegistry]
});
```
then that counter will not be present in our `sys/metrics` endpoint. This makes some sense but shouldn't there be a way to have all the functionality in your service report metrics to the same endpoint? However, I believe the issue is that the server and the module each have their own`package.json` and each run their own instance of the imported modules. So if that is the case, then I wonder how my desired functionality can be achieved? Without having to define my default registry in my module and then propagating it upwards. I want every module to be able to run `const defaultReg = require('prom-client').register` and that should make the metrics defined towards that register all appear on our metrics endpoint

pontusarfwedson:
Alright, so that gave me
```
<service>@1.0.0 <local-path-to-service>
├─┬ <package>@1.0.0 -> <local-path-to-module>
│ └── prom-client@11.0.0
└── prom-client@11.0.0
```
and after `npm dedupe` I unfortunately when running `npm ls prom-client` got
```
<service>@1.0.0 <local-path-to-service>
├─┬ <package>@1.0.0 -> <local-path-to-module>
│ └── UNMET DEPENDANCY prom-client@11.0.0
└── prom-client@11.0.0
npm ERR! missing: prom-client@^11.0.0, required by <package>@1.0.0
```
which breaks my build.

SimenB:
try `npm dedupe && npm prune && npm install`. npms has some issues when deduping which seems to resolve themselves when you run a second install

pontusarfwedson:
Those 3 commands take me back to the same state as before them :) so they are still un-deduped

SimenB:
Is it behind a symlink? If so it won't dedupe properly

pontusarfwedson:
Nope, regular dir :(

SimenB:
Why do you have `-> <local-path-to-module>` in the `ls` output then?

pontusarfwedson:
Because its a "module" locally on my computer? :) 
But regardless, it seems like we cannot talk to the same registry if we initialize the prom-client from two different modules, despite making dedupe work.

siimon:
I'd say sharing a registry between 2 modules is an anti pattern and should be avoided at all costs.. Then you have dependencies between those 2 and it could easily become a mess!

I would instead let both of the modules have their own registry, and the in the end when you want to expose them, you merge them together.

pontusarfwedson:
Alright! So If we have a service that has a set of metrics exposed to `sys/metrics` which requires a package/module which in turn also has some metrics, then there's no good way of combining them? I would say that the golang lib is maybe the most prominent of the prometheus libs and in that use-case, if you have `import prometheus ...` in any of the packages and then run prometheus.MustRegister(metric) in any of the packages, then all those metrics will go on the default registry. Perhaps your design was not intended to work this way, but why wouldn't this be desirable? Note that I am not to experienced with node :)

siimon:
Different religions I guess :) In my mind every module is self contained and has no side effects on others.. 

Yes, you got my suggested approach right :) What is the problem with adding another registry to the merge function?

SimenB:
Sorta related to #186.

>Because its a "module" locally on my computer? :)

Then it's linked, and not really deduped, which I think is why it doesn't work properly.

That said, I think we can have the discussion for how to best combine metrics from different modules in #186, and close this PR as the registry is already a singleton, and making this change won't help the case of a non-deduped `prom-client`.

PR #185, opened 2018-04-07
Support for browser use

pbadenski:
Exposes separate index.js for node & browser. Browser version doesn't expose default metrics (which are mostly if not all node specific) & AggregatorRegistry.


siimon:
Why do you want to use Prometheus in a browser?

pbadenski:
I work on an SPA for pricing financial instruments - a lot of calculation happens in the user's browser. We extensively use client-side metrics to troubleshoot performance and memory problems. We currently use InfluxDB and thinking of switching to Prometheus for consistency with the backend.

SimenB:
How would the metrics end up in the prometheus server?

I suppose you'd have to POST the metrics to your app server regularly. I wonder if it would be better to build a module running in the browser which serialises to something that `prom-client` can parse and add to its own metrics instead of using `prom-client`'s stuff in the browser directly

pbadenski:
Idea is to push directly from the browser to pushgateway.

siimon:
I agree on @SimenB's suggestion, including `prom-client` inside a browser is in my opinion an anti-pattern. 

I'm also not happy to have that dependency, if you want that, I think it fits better in it's own node module. So I'm closing this PR!

PR #193, opened 2018-05-11
Fix aggregator method being lost when aggregating several registries

Ratismal:
When aggregating registries, the `aggregator` property of the metrics is lost. This PR aims to repopulate it with the aggregator property of the first metric, for better consistency.

This will prevent the `AggregatorRegistry#aggregate` function from breaking when attempting to aggregate an already-aggregated registry.


zbjornson:
Can you update the tests though, please?

SimenB:
Should we have a new test covering "attempting to aggregate an already-aggregated registry"?

SimenB:
Also, please update the changelog

zbjornson:
Sorry for the delay. This looks good to me. @siimon / @SimenB?

PR #198, opened 2018-06-25
Add optional prefix to default metrics

DanForys:
Adds the ability to provide a `prefix` string to the default metrics.

This helps in our use-case, where we have many applications residing on a single Prometheus instance and need to easily identify which metrics belong to which application.

### DIFF
--- a/lib/metrics/eventLoopLag.js
+++ b/lib/metrics/eventLoopLag.js
@@ -12,9 +12,11 @@ function reportEventloopLag(start, gauge) {
 	gauge.set(seconds, Date.now());
 }
 
-module.exports = registry => {
+module.exports = (registry, config = {}) => {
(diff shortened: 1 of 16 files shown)


siimon:
Looks good to me! @SimenB do you have any opinons on this?

lib/metrics/eventLoopLag.js
```diff
@@ -12,9 +12,11 @@ function reportEventloopLag(start, gauge) {
 	gauge.set(seconds, Date.now());
 }
 
-module.exports = registry => {
+module.exports = (registry, config = {}) => {
```
SimenB: we support node 4, I don't think default params are supported there?

SimenB: Hah, I lie! We've dropped it. Good to go, then 🙂

PR #201, opened 2018-06-29
Fix #200

### LINKED ISSUE #200: TypeError: Cannot read property 'prefix' of undefined in openFileDescriptor
heey! 

optional prefix for metrics is failing because config is undefined.

thank you!

```
/opt/atlassian/pipelines/agent/build/node_modules/prom-client/lib/metrics/processOpenFileDescriptors.js:13
	const namePrefix = config.prefix ? config.prefix : '';
	                          ^
TypeError: Cannot read property 'prefix' of undefined
```

frankie567:
Fix the issue #200. The default value for `config` was missing. I added a unit test.

I took the liberty to add Sinon in devDependencies in order to fake the platform, so that we can run the unit test on any platform ; because this metric is otherwise only available on Linux.

Regards!

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
 		"lint-staged": "^7.0.0",
 		"lolex": "^2.1.3",
 		"prettier": "1.11.1",
+		"sinon": "^6.0.1",
 		"typescript": "^2.5.2"
 	},
 	"dependencies": {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+* Fixed `processOpenFileDescriptors` metric when no custom config was set
+
 ### Added
 
 ## [11.1.0] - 2018-06-29
--- a/lib/metrics/processOpenFileDescriptors.js
+++ b/lib/metrics/processOpenFileDescriptors.js
@@ -5,7 +5,7 @@ const fs = require('fs');
 
 const PROCESS_OPEN_FDS = 'process_open_fds';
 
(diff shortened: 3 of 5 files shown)


siimon:
Thanks!

Do you mind updating the changelog aswell?

frankie567:
Done! Hope it matches the way you do it usually 🙂

siimon:
Great! Thanks 👍

siimon:
Published in 11.1.1

package.json
```diff
@@ -31,6 +31,7 @@
 		"lint-staged": "^7.0.0",
 		"lolex": "^2.1.3",
 		"prettier": "1.11.1",
+		"sinon": "^6.0.1",
```
SimenB: We have Jest which can do this already

frankie567: Didn't know about that, sorry. I can make another PR to fix that if you wish.

SimenB: That'd be awesome!

frankie567: I tried a bit, but I have to admit I'm not very confident with Jest. You'll find below what I tried, but without luck. As you seem to be a big contributor to Jest, I hope you can point me in the right direction :)

**Attempt 1**

`package.json`
```json
"jest": {
  "testEnvironment": "node",
  "testEnvironmentOptions": {
    "process": {
      "platform": "linux" 
    }
  },
  "testRegex": ".*Test\\.js$"
},
```

**Attempt 2**

`processOpenFileDescriptorsTest.js`
```js
jest.mock('process', () => {
  return {
    platform: 'linux'
  };
});
```

SimenB: Since all test files run in isolation in Jest, you can just assign.

```js
process.platform = 'linux';
```

PR #208, opened 2018-08-01
Complementing prefix in index.d.ts

OscSer:
adding prefix to DefaultMetricsCollectorConfiguration, related to the release v11.1.0.


SimenB:
It would be awesome to actually test the definition file as it seems to always drift...

Anyways, thanks for the PR!

Issue #216, opened 2018-09-12
Histogram scrape performance with multiple labels/label values


KevinAMurray:
Hi,

We have a use case where we are using histograms together with between 2 and 6 labels (depending on the metric), and where those labels have between 3 and 40 values.  What we are seeing is that the time to perform the scrape increases dramatically when we have more labels and more label values, to the point where the scrape operation could take a couple of seconds (for a worst case situation, which I would expect to happen after the service has been running for some time).

Whilst this clearly isn't a problem for Prometheus, in our particular case the node application is  single threaded (e.g. we cannot use Cluster or similar).  This means that a prometheus scrape will block that particular instance for a second or two whilst the scrape is happening.

I've explored the prom-client code, and I have made some improvements by effectively pre-computing information when the histogram's time series is created, and by performing some calculations during observe().  I.e. I've slightly changed the design trade-off between optimal observe() performance to improve the scrape performance, with an increase in memory requirements (though I *think* those memory requirements are probably the same when the scrape is happening).

Have other people encountered this sort of performance issue before, and are there other solutions? (As mentioned, effectively we can't use something like Cluster.)

I'm happy to provide code snippets (or a fork) as the basis of improvements or for more discussions.

Thanks,

Kevin

SimenB:
If we have performance issues with serialising the metrics data, I think it makes sense to take a look at either our storage format or our algorithm 🙂 A test case would be awesome, so we have a baseline. Maybe look into adding benchmarks as well to the repo?

/cc @siimon @zbjornson

KevinAMurray:
Okay -- let me look at modifying our benchmark program for histograms to provide a baseline (and allow others to check I'm not missing something obvious!)

KevinAMurray:
Hmm. I think there is a bug in the multipliers above so the numbers will be wrong....

SimenB:
Please give 11.1.2 a try 🙂 We can reopen if it's still an issue

KevinAMurray:
Thanks. I'm seeing about a double in performance. Very Nice! thanks. The performance is much more linear, which is great.

The benchmark test program is sitting in [my fork in the benchmark directory](https://github.com/KevinAMurray/prom-client/tree/master/benchmarks). Shall I PR it into the main branch?

[...]

PR #219, opened 2018-09-19
Fix histogram scrape performance

nowells:
When profiling histogram scraping I discovered that the major bottleneck is creating a clone of the accumulator in reduce. We should not actually need to skip mutation.

You can see the benchmark of `concat` vs `push` here:  https://codeburst.io/jsnoob-push-vs-concat-basics-and-performance-comparison-7a4b55242fa9

This should address https://github.com/siimon/prom-client/issues/216

I used this test script to profile.

```js
const prom = require('prom-client');
const _ = require('lodash');

const histogram = new prom.Histogram({
    name: 'histogram',
    help: 'histogram',
    labelNames: ['a', 'b', 'c', 'd'],
});

_.times(2, a => {
    _.times(3, b => {
        _.times(10, c => {
            _.times(100, d => {
                const end = histogram.startTimer();

                end({a,b,c,d});
            });
        });
    });
});

console.log('Starting metrics');
const now = Date.now();
const metrics = prom.register.getMetricsAsJSON();

console.log(`${(Date.now() - now) / 1000}s`);
```

Which resulted in before this patch of `7.021s` and after of `0.501s`

### DIFF
--- a/lib/histogram.js
+++ b/lib/histogram.js
@@ -288,7 +288,7 @@ function extractBucketValuesForExport(histogram) {
 
 function addSumAndCountForExport(histogram) {
 	return (acc, d) => {
-		acc.push.apply(acc, d.buckets);
+		acc.push(...d.buckets);
 
 		const infLabel = Object.assign({ le: '+Inf' }, d.data.labels);
 		acc.push(
--- a/lib/histogram.js
+++ b/lib/histogram.js
@@ -288,7 +288,7 @@ function extractBucketValuesForExport(histogram) {
 
 function addSumAndCountForExport(histogram) {
 	return (acc, d) => {
-		acc = acc.concat(d.buckets);
+		acc.push.apply(acc, d.buckets);
 
 		const infLabel = Object.assign({ le: '+Inf' }, d.data.labels);
 		acc.push(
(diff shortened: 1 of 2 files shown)


lib/histogram.js
```diff
@@ -288,7 +288,7 @@ function extractBucketValuesForExport(histogram) {
 
 function addSumAndCountForExport(histogram) {
 	return (acc, d) => {
-		acc = acc.concat(d.buckets);
+		acc.push.apply(acc, d.buckets);
```
SimenB: why not just `push(d.buckets);`?

SimenB: oh, it's an array. makes sense :)

SimenB:
Thanks for the PR!

lib/histogram.js
SimenB: it can be `acc.push(...d.buckets);`, we have a minimum node version of 6

nowells:
> Thanks for the PR! Mind updating the changelog as well?

Done

nowells:
Thanks for the merge @SimenB! I did not package bump because I assume that you do that as part of your release process. Thanks for the great package, and for such a quick response!

SimenB:
It's released as 11.1.2 now 🙂

nowells:
> It's released as 11.1.2 now 🙂

❤️ thanks!!!

siimon:
So nice!! Thank you!  ❤️

PR #220, opened 2018-09-20
Mutate labels to avoid excessive object cloning.

nowells:
I discovered another performance optimization we can perform, which is to mutate the `labels` and avoid `Object.assign` where we can in hot paths.

```
❯ node benchmarks/index.js
Progress:
published#registry#getMetricsAsJSON x 1,819 ops/sec ±4.51% (80 runs sampled)
published#registry#metrics x 1,301 ops/sec ±5.13% (76 runs sampled)
local#registry#getMetricsAsJSON x 9,025 ops/sec ±4.71% (82 runs sampled)
local#registry#metrics x 2,968 ops/sec ±3.59% (85 runs sampled)

Results:
╔══════════════════╤════════════════════════════╤═══════════════════════════╗
║ registry         │ published                  │ local                     ║
╟──────────────────┼────────────────────────────┼───────────────────────────╢
║ getMetricsAsJSON │ 1819.383794048214 ops/sec  │ 9024.761535386682 ops/sec ║
╟──────────────────┼────────────────────────────┼───────────────────────────╢
║ metrics          │ 1300.6959959677145 ops/sec │ 2968.460595395247 ops/sec ║
╚══════════════════╧════════════════════════════╧═══════════════════════════╝

Fastest is local#registry#getMetricsAsJSON
Fastest is local#registry#metrics
```

### DIFF
--- a/lib/histogram.js
+++ b/lib/histogram.js
@@ -282,9 +282,17 @@ function convertLabelsAndValues(labels, value) {
 
 function extractBucketValuesForExport(histogram) {
 	return bucketData => {
-		const buckets = histogram.upperBounds.map(
-			createBucketValues(bucketData, histogram)
-		);
+		const buckets = [];
+		const bucketLabelNames = Object.keys(bucketData.labels);
+		let acc = 0;
+		for (const upperBound of histogram.upperBounds) {
+			acc += bucketData.bucketValues[upperBound];
+			const lbls = { le: upperBound };
+			for (const labelName of bucketLabelNames) {
+				lbls[labelName] = bucketData.labels[labelName];
+			}
+			buckets.push(setValuePair(lbls, acc, `${histogram.name}_bucket`));
+		}
 		return { buckets, data: bucketData };
 	};
 }
@@ -293,23 +301,16 @@ function addSumAndCountForExport(histogram) {
 	return (acc, d) => {
 		acc.push(...d.buckets);
 
-		const infLabel = Object.assign({ le: '+Inf' }, d.data.labels);
-		acc.push(setValuePair(infLabel, d.data.count, `${histogram.name}_bucket`));
-		acc.push(setValuePair(d.data.labels, d.data.sum, `${histogram.name}_sum`));
+		const infLabel = { le: '+Inf' };
+		for (const label of Object.keys(d.data.labels)) {
+			infLabel[label] = d.data.labels[label];
+		}
 		acc.push(
+			setValuePair(infLabel, d.data.count, `${histogram.name}_bucket`),
+			setValuePair(d.data.labels, d.data.sum, `${histogram.name}_sum`),
 			setValuePair(d.data.labels, d.data.count, `${histogram.name}_count`)
 		);
 		return acc;
-	};
-}
-
-function createBucketValues(bucket, histogram) {
-	let acc = 0;
-	return upperBound => {
-		acc += bucket.bucketValues[upperBound];
-		const lbls = Object.assign({ le: upperBound }, bucket.labels);
-		const valuePair = setValuePair(lbls, acc, `${histogram.name}_bucket`);
-		return valuePair;
 	};
 }
 
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -29,39 +29,46 @@ class Registry {
 		const opts = Object.assign({}, defaultMetricsOpts, conf);
 		const item = metric.get();
 		const name = escapeString(item.name);
-		let help = escapeString(item.help);
-		help = ['#', 'HELP', name, help].join(' ');
-		const type = ['#', 'TYPE', name, item.type].join(' ');
+		const help = `# HELP ${name} ${escapeString(item.help)}`;
+		const type = `# TYPE ${name} ${item.type}`;
+		const defaultLabelNames = Object.keys(this._defaultLabels);
 
-		const values = (item.values || []).reduce((valAcc, val) => {
-			const merged = Object.assign({}, this._defaultLabels, val.labels);
+		let values = '';
+		for (const val of item.values || []) {
+			val.labels = val.labels || {};
+			for (const labelName of defaultLabelNames) {
+				val.labels[labelName] =
+					val.labels[labelName] || this._defaultLabels[labelName];
+			}
 
-			const labels = Object.keys(merged).map(
-				key => `${key}="${escapeLabelValue(merged[key])}"`
-			);
+			let labels = '';
+			for (const key of Object.keys(val.labels)) {
+				labels += `${key}="${escapeLabelValue(val.labels[key])}",`;
+			}
 
 			let metricName = val.metricName || item.name;
-			if (labels.length) {
-				metricName += `{${labels.join(',')}}`;
+			if (labels) {
+				metricName += `{${labels.replace(/,$/, '')}}`;
 			}
 
-			const line = [metricName, getValueAsString(val.value)];
-			if (opts.timestamps) {
-				line.push(val.timestamp);
+			let line = `${metricName} ${getValueAsString(val.value)}`;
+			if (opts.timestamps && val.timestamp) {
+				line += ` ${val.timestamp}`;
 			}
-			valAcc += line.join(' ').trim();
-			valAcc += '\n';
-			return valAcc;
-		}, '');
+			values += `${line.trim()}\n`;
+		}
 
-		const acc = [help, type, values].join('\n');
-		return acc;
+		return `${help}\n${type}\n${values}`.trim();
 	}
 
 	metrics(opts) {
-		return this.getMetricsAsArray()
-			.map(metric => this.getMetricAsPrometheusString(metric, opts))
-			.join('\n');
+		let metrics = '';
+
+		for (const metric of this.getMetricsAsArray()) {
+			metrics += `${this.getMetricAsPrometheusString(metric, opts)}\n\n`;
+		}
+
+		return metrics.replace(/\n$/, '');
 	}
 
 	registerMetric(metricFn) {
@@ -83,21 +90,25 @@ class Registry {
 	}
 
 	getMetricsAsJSON() {
-		return this.getMetricsAsArray().map(metric => {
+		const metrics = [];
+		const defaultLabelNames = Object.keys(this._defaultLabels);
+
+		for (const metric of this.getMetricsAsArray()) {
 			const item = metric.get();
-			if (!item.values) {
-				return item;
+
+			if (item.values) {
+				for (const val of item.values) {
+					for (const labelName of defaultLabelNames) {
+						val.labels[labelName] =
+							val.labels[labelName] || this._defaultLabels[labelName];
+					}
+				}
 			}
 
-			item.values = item.values.map(val =>
-				// Avoid mutation and merge metric labels with registry default labels
-				Object.assign({}, val, {
-					labels: Object.assign({}, this._defaultLabels, val.labels)
-				})
-			);
+			metrics.push(item);
+		}
 
-			return item;
-		});
+		return metrics;
 	}
 
 	removeSingleMetric(name) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Fixed performance by avoiding `Object.assign` on hot paths, as well as
+  mutating objects when appropriate.
+
 ## [11.1.2] - 2018-09-19
 
 ### Changed


nowells:
We could further cut 1 second off of the benchmark above by moving the assignment of `defaultLabels` to metric creation time, instead of at serialization time. This sounds reasonable to me, but it is technically a change in behavior as you could theoretically change the set of defaultLabels dynamically on calls and have them reflect on each call to serialize. I do not know if this was an intended feature, or simply an API design result. If that sounds interesting, let me know and I can take on creating a PR to change the default label assignment behavior.

performance

SimenB:
Thanks!

>I also decided to include a basic benchmarking suite. If you want me to improve the suite, change it, or remove it entirely let me know, but seems like it might be useful to compare/see performance over time.

That's awesome, we've long wanted a benchmarking suite! But if we add it, I think it should be written in JS using e.g. https://github.com/bestiejs/benchmark.js/.

nowells:
@SimenB happy to do it in JS, I only used the one I added here because I already had it sitting around in another project, so just copy and pasted. I'll take on creating the benchmark suite using that library (or something similar). I will rip out the current suite and leave just the performance changes (you can feel free to test with the test script locally to confirm)

SimenB:
@KevinAMurray linked to a suite written in the framework I suggested in https://github.com/KevinAMurray/prom-client/tree/master/benchmarks. Might be a good starting point 🙂

lib/registry.js
```diff
@@ -89,12 +89,11 @@ class Registry {
...
 			}
 
-			item.values = item.values.map(val =>
-				// Avoid mutation and merge metric labels with registry default labels
-				Object.assign({}, val, {
-					labels: Object.assign({}, this._defaultLabels, val.labels)
-				})
-			);
+			for (const val of item.values) {
+				for (const label of Object.keys(this._defaultLabels)) {
```
SimenB: Since this is a perf PR, we can lift `Object.keys(this._defaultLabels)` put of the inner loop to avoid iterating through that object every time

nowells: Definitely

lib/histogram.js
```diff
@@ -279,9 +279,16 @@ function convertLabelsAndValues(labels, value) {
 
 function extractBucketValuesForExport(histogram) {
 	return bucketData => {
-		const buckets = histogram.upperBounds.map(
-			createBucketValues(bucketData, histogram)
-		);
+		const buckets = [];
+		let acc = 0;
```
nowells: I inlined this to avoid creating a function on a big loop, also made performance profiling cleaner. If you don't like it I can revert, but I did get a performance gain by inlining the method.

SimenB: I'm fine with this, especially if we now have benchmarks where it shows improvement

lib/histogram.js
```diff
@@ -280,12 +280,13 @@ function convertLabelsAndValues(labels, value) {
 function extractBucketValuesForExport(histogram) {
 	return bucketData => {
 		const buckets = [];
+		const bucketLabels = Object.entries(bucketData.labels);
```
SimenB: `Object.entries` is not available in Node 6 :(

nowells: Whoops! I'll switch back to just shifting Object.keys up. Sorry!

SimenB: no worries! I thought we had the linter set up to yell, apparently not...

nowells:
I switched to `benchtable` (which just extends `benchmark` internally) for better test setup as well as reporting. If you would rather it just be vanilla `benchmark` it should be easy to switch back.

```
❯ node benchmarks/index.js
getMetricsAsJSON for inputs published x 17.88 ops/sec ±7.60% (35 runs sampled)
getMetricsAsJSON for inputs local x 68.51 ops/sec ±8.18% (58 runs sampled)
Fastest is getMetricsAsJSON for inputs local
+-----------+------------------+
|           │ getMetricsAsJSON |
+-----------+------------------+
| published │ 17.88 ops/sec    |
+-----------+------------------+
| local     │ 68.51 ops/sec    |
+-----------+------------------+
```

lib/histogram.js
SimenB: #221 fwiw 🙂

SimenB:
`benchtable` looks nice!

nowells:
> benchtable looks nice!

It is not maintained (2016 was last commit) but it just extends `benchtable@^2` so it seemed reasonable. But I would understand if you would rather not pull in a seemingly unmaintained library, but I figure that it probably "just works" and it is just an extension of a library that is actively maintained.

SimenB:
Yeah, no worries. I'd rather think of it as "complete" and not "unmaintained" 😀

SimenB:
Mind updating the changelog?

nowells:
> Mind updating the changelog?

Sure thing!

nowells:
```
> node ./benchmarks/index.js

getMetricsAsJSON for inputs published x 1,627 ops/sec ±5.03% (78 runs sampled)
getMetricsAsJSON for inputs local x 8,694 ops/sec ±4.96% (82 runs sampled)
metrics for inputs published x 1,252 ops/sec ±6.36% (76 runs sampled)
metrics for inputs local x 2,250 ops/sec ±5.84% (81 runs sampled)
Fastest is getMetricsAsJSON for inputs local
+------------------+---------------+---------------+
|                  │ published     │ local         |
+------------------+---------------+---------------+
| getMetricsAsJSON │ 1,627 ops/sec │ 8,694 ops/sec |
+------------------+---------------+---------------+
| metrics          │ 1,252 ops/sec │ 2,250 ops/sec |
+------------------+---------------+---------------+
```

siimon:
Great work! I'll try and have a closer look tonight

lib/registry.js
```diff
@@ -49,19 +54,23 @@ class Registry {
 			if (opts.timestamps) {
 				line.push(val.timestamp);
 			}
-			valAcc += line.join(' ').trim();
-			valAcc += '\n';
-			return valAcc;
-		}, '');
+			values += line.join(' ').trim();
```
SimenB: `values += line.join(' ').trim() + '\n';` to save an assignment

benchmarks/index.js
```diff
@@ -0,0 +1,58 @@
...
+		registry.metrics();
+	})
+	.on('cycle', event => {
+		if (!event.target.error) {
+			// eslint-disable-next-line no-console
+			console.log(event.target.toString());
+		}
+	})
+	.on('complete', () => {
+		// eslint-disable-next-line no-console
```
SimenB: Instead of this, just add it as exception here: https://github.com/siimon/prom-client/blob/25255c39b5bb225e89e77e69d506061bbb7c4af0/.eslintrc#L64-L69

nowells:
```
❯ node benchmarks/index.js
getMetricsAsJSON for inputs published x 1,764 ops/sec ±4.77% (82 runs sampled)
getMetricsAsJSON for inputs local x 9,399 ops/sec ±3.89% (84 runs sampled)
metrics for inputs published x 1,386 ops/sec ±3.71% (83 runs sampled)
metrics for inputs local x 3,016 ops/sec ±4.65% (85 runs sampled)
Fastest is getMetricsAsJSON for inputs local
+------------------+---------------+---------------+
|                  │ published     │ local         |
+------------------+---------------+---------------+
| getMetricsAsJSON │ 1,764 ops/sec │ 9,399 ops/sec |
+------------------+---------------+---------------+
| metrics          │ 1,386 ops/sec │ 3,016 ops/sec |
+------------------+---------------+---------------+
```

KevinAMurray:
(not come across BenchTable -- very nice!)

Benchmark is a great addition. And loving the performance increases.

package.json
```diff
@@ -39,6 +41,7 @@
 		"lint-staged": "^7.0.0",
 		"lolex": "^2.1.3",
 		"prettier": "1.14.2",
+		"prom-client": "^11.1.2",
```
zbjornson: Love testing against a published version. 👍 

This might be a non-issue, but we'll have to update this (and the lock files) after every release (or at least when we care about benchmarks). I don't see a way to exclude a module from the lock files, which would give some options for improving that situation. Not listing it as a dependency at all could be another option, and then we could put the `require` for it in a try/catch?

SimenB: we could add a CI check or something? `npm show prom-client version` returns latest published, and if the local version does not match, throw?

nowells: Yeah, that is interesting. We could have a `postinstall` step that only runs locally that installs the latest version? But yeah, I hadn't thought of the lockfile issue. Let me know what you prefer.

SimenB:
Could you add some prose about the benchmarks to a `CONTRIBUTING.md` or something? Not much, just the fact that they exist and how to run them 🙂

lib/registry.js
```diff
@@ -26,42 +26,51 @@ class Registry {
...
 		const item = metric.get();
 		const name = escapeString(item.name);
 		let help = escapeString(item.help);
-		help = ['#', 'HELP', name, help].join(' ');
-		const type = ['#', 'TYPE', name, item.type].join(' ');
-
-		const values = (item.values || []).reduce((valAcc, val) => {
-			const merged = Object.assign({}, this._defaultLabels, val.labels);
+		help = `# HELP ${name} ${help}`;
+		const type = `# TYPE ${name} ${item.type}`;
```
siimon: 👏 that was long overdue!

lib/registry.js
```diff
@@ -26,42 +26,51 @@ class Registry {
 	}
 
 	getMetricAsPrometheusString(metric, conf) {
-		const opts = Object.assign({}, defaultMetricsOpts, conf);
```
siimon: The point with using Object.assign like this was both a good way to extend and overwrite the default settings but also a way to make sure a specific configuration value always was there. I know that you changed on how you call this method from the metrics function but this function is public available so it is a breaking change.

package.json
```diff
@@ -31,15 +32,18 @@
...
 		"eslint-plugin-node": "^7.0.1",
 		"eslint-plugin-prettier": "^2.1.2",
 		"express": "^4.13.3",
 		"husky": "^0.14.3",
 		"jest": "^23.6.0",
 		"lint-staged": "^7.0.0",
+		"lodash": "^4.17.11",
 		"lolex": "^2.1.3",
 		"prettier": "1.14.3",
+		"prom-client": "^11.1.2",
```
siimon: I guess you could wildcard it? Doesn't solve the lock problem though.

package.json
zbjornson: Oo, using a package.json script is another good idea. Wouldn't want it to run all the time though, so if it's in `postinstall`, would have to check somehow if it's being installed in a git checkout or as a dependency, I think?

(Might not be worth the trouble to make this "nice" -- could just manually bump it as necessary.)

lib/registry.js
nowells: I can revert this change, not the biggest perf win. Good call on the public api change potential.

nowells:
> Could you add some prose about the benchmarks to a CONTRIBUTING.md or something? Not much, just the fact that they exist and how to run them 🙂

I was planning on just making the benchmarking part of the `test` phase, and failing if the performance was not within a standard deviation. Would you rather have just the CONTRIBUTING, just the automatic failures, or both?

SimenB:
I think HW varies too much for it to make sense to run as part of the test build. But I think we can paste the results of whatever we end up merging into a doc as a baseline

nowells:
> I think HW varies too much for it to make sense to run as part of the test build.

The build will run not with the results of a previous run, but will run the two jobs (with published and local) so I don't think the HW matters, it will all be relative to the current HW compute power. If we were going based on previous results where it was on different hardware I would agree. Thoughts?

nowells:
1) I added the automatic check to fail if the installed version of `prom-client` is not equal to the latest on `npm`.
2) I setup the benchmark suite to fail the build if the performance of the local changes are worse than the published (we might want to tweak in the future to give it some wiggle room)
3) I may have gone a little overboard on fleshing out the benchmarks suite.

prom-client benchmark results

SimenB:
>I may have gone a little overboard on fleshing out the benchmarks suite.

Haha, this is awesome! 😀 You might want to consider creating a module we can install for the setup part, but this looks really good. Thank you so much for working on it!

SimenB:
Also, would you mind creating a separate PR for the benchmark (once fully iterated)? That will keep this PR more focused on perf improvements

bin/check-versions.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+INSTALLED_VERSION=`node -p -e "require('prom-client/package.json').version"`
```
SimenB: you don't need the `-e`, `-p` is enough 🙂

nowells:
I extracted the benchmark suite to https://github.com/siimon/prom-client/pull/222

nowells:
Thanks for the great review everyone. And thanks for the merge @siimon! Will you be releasing this as a new version, or waiting to batch it up with other changes? Thanks!!!

SimenB:
`prom-client 11.1.3 published 🎉`

PR #221, opened 2018-09-20
chore: add eslint-plugin-node

SimenB:
This will warn us against unsupported features without waiting for CI
to run

Also upgrade all dated dependencies

Motivation is if we use e.g. Object.entries, we normally have to wait for CI to fail. Now we'll get a linting error in our IDE or precommit:

```
  31:4  error  The 'Object.entries' is not supported until Node.js 7.0.0. The configured version range is '>=6.1'  node/no-unsupported-features/es-builtins
```

### DIFF
--- a/lib/summary.js
+++ b/lib/summary.js
@@ -4,18 +4,21 @@
 'use strict';
 
 const util = require('util');
-const globalRegistry = require('./registry').globalRegistry;
+const { globalRegistry } = require('./registry');
 const type = 'summary';
-const getProperties = require('./util').getPropertiesFromObj;
-const getLabels = require('./util').getLabels;
-const hashObject = require('./util').hashObject;
-const validateLabels = require('./validation').validateLabel;
-const validateMetricName = require('./validation').validateMetricName;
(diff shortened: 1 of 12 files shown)


lib/summary.js
```diff
@@ -13,7 +13,7 @@ const validateLabels = require('./validation').validateLabel;
 const validateMetricName = require('./validation').validateMetricName;
 const validateLabelNames = require('./validation').validateLabelName;
 const TDigest = require('tdigest').TDigest;
-const isObject = require('util').isObject;
+const isObject = require('./util').isObject;
```
SimenB: probably a typo, but `isObject` from core `util` module is deprecated

package.json
```diff
@@ -9,7 +9,7 @@
 		"index.d.ts"
 	],
 	"engines": {
-		"node": ">=6"
+		"node": ">=6.1"
```
SimenB: 6.1 for `cpuUsage`. Probably OK to sneak in here, it's years old by this point

PR #223, opened 2018-10-11
impl BigNumber with gauge

derori:


SimenB:
@derori sorry about the missing feedback!

I think the correct solution here is to use `BigInt` (https://tc39.github.io/proposal-bigint/) if your runtime supports it (available from node 10.8.0). I haven't tested, but it should work fine. If we do not support, PR very much welcome adding support for that :)

fanatid:
`BigInt` sounds awesome, but node LTS releases without `BigInt` will exists at least next 1 year...
I highly recommend to consider `bn.js` instead `bignumber.js`

SimenB:
The current node LTS has bigint. I'm not saying we should code it into _this_ library, but in your app, which should run on node 10, you can use `BigInt`. But I'm saying that if passing a `BigInt` does not work, we should update to make sure it does 🙂 A `BigInt` stringified looks just like a number so Prometheus should be happy with it

PR #228, opened 2018-10-22
startTimer with factor parameter

manorlh:
added factor option to startTimer, for open the possibility to use other units than second.


SimenB:
Is this supported by any other client library? Prometheus docs explicitly says to use base units: https://prometheus.io/docs/practices/naming/#base-units

siimon:
Will not be merged since we are following the Prometheus docs https://prometheus.io/docs/practices/naming/#base-units

Issue #205, opened 2018-07-20
Support sliding window for summary metrics


florianmutter:
Prom Client should support sliding windows for summaries as described [here](https://prometheus.io/docs/concepts/metric_types/#summary) and [here](https://prometheus.io/docs/practices/histograms/#quantiles).

This is needed for metrics like request duration where you don't want your request from a week ago influencing the quantiles that are calculated for now.

siimon:
We are using https://github.com/welch/tdigest for quantiles calculations, not sure if that library has support for setting max ages. If it does it should be trivial to implement, if not then it will be abit more work..

florianmutter:
As far as I can tell they don't have built in support. You would need to implement it in the prom client. Like the Java implementation you would need multiple t-digest buckets which have a given length and where the quantiles are calculated. Here is the official Java-Implementation with does that: https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Summary.java#L70-L77

PR #229, opened 2018-11-05
Sliding windows for summaries

carsotho:
Resolves https://github.com/siimon/prom-client/issues/205. Feedback is welcome!

### DIFF
--- a/lib/summary.js
+++ b/lib/summary.js
@@ -18,7 +18,8 @@ const {
 	validateMetricName,
 	validateLabelName
 } = require('./validation');
-const { TDigest } = require('tdigest');
+//const { TDigest } = require('tdigest');
+const timeWindowQuantiles = require('./timeWindowQuantiles');
 
 class Summary {
 	/**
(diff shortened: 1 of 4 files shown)


siimon:
Thanks for this! I will try and have a look at this today!

lib/summary.js
```diff
@@ -60,12 +60,17 @@ class Summary {
...
-				registers: [globalRegistry]
+				registers: [globalRegistry],
+				maxAgeSeconds: obj.maxAgeSeconds,
+				ageBuckets: obj.ageBuckets
 			};
 		}
 
 		validateInput(config.name, config.help, config.labelNames);
 
+		this.maxAgeSeconds = config.maxAgeSeconds || 600;
```
siimon: This is a breaking change, altering the current behaviour. Could the defaults be changed so it doesn't change the current behaviour?

siimon: That's great but it's still a breaking change and I want to avoid them as much as possible if it's doable (which I think it is in this case)

siimon: I get your point but I don't think it's worth a breaking change for this, however I think for the next major bump we probably should change the default to the sliding window instead.

carsotho: Alright I'll fix it :)

Issue #142, opened 2017-07-11
Memory leak


gwilakers:
Hi!

I currently have an implementation of prom client running that seems to continuously grow the node heap until an eventual crash/service restart when collecting API metrics. Currently, I am collecting the default metrics and a couple custom ones that just count websocket connections. These all work fine. The problems arise from the API metrics I try to collect using a middleware that sits on our Hapi routes. This middleware listens for 'onRequest' and then 'response' for collecting a response time in milliseconds, while adding labels for method, path, statusCode, and usernames.

Here is what the middleware looks like (I do not believe the problem is with this):

```
/**
* This is the middleware that looks at each request as it comes through the server
* @param {Object} server This is the instantiated Hapi server that we wish to listen on.
* @param {Object} options This is an object that can contain additional data.
* @param {Function} next This is the callback function that allows additional requests to continue.
* @return {Function} Callback function.
*/
const middlewareWrapper = (server, options, next) => {
  server.ext('onRequest', (request, reply) => {
    request.startTime = process.hrtime()
    return reply.continue()
  })
  server.on('response', (request) => {
    logic.apiPerformance(request.response.statusCode, request.startTime, sanitizeRequestData(request))
  })
  return next()
}
```

The metrics themselves are used inside of that apiPerformance function which looks like this:

```
  /**
  * This function takes various information about a request, performs some basic calculations, and then sends that off to loggly.
  * @param {Integer} statusCode This is the HTTP statusCode that will be returned to the user with the response.
  * @param {Array} startTime The difference in time from the request recieved to the response given. https://nodejs.org/api/process.html#process_process_hrtime_time
  * @param {Object} responseData An object with sanitized response data.
  */
  apiPerformance: (statusCode, startTime, responseData) => {
    const diff = process.hrtime(startTime)
    const responseTime = ((diff[0] * 1e3) + (diff[1] * 1e-6))
    const time = new Date()
    logger.transports.loggly.log('info', 'Route Log', {
      responseCode: statusCode,
      route: responseData.route,
      host: responseData.host,
      port: responseData.port,
      method: responseData.method,
      userData: responseData.userData,
      responseTime: `${responseTime}ms`,
      params: responseData.params,
      timestamp: time.toISOString(),
      tags: ['events', 'api', 'performance']
    }, (error, response) => {
      if (error) console.log(error)
    })
    if (config.metrics.prometheus.enabled) {
[...]

PR #234, opened 2018-12-28
Compress t-digest to prevent memory leakage

atd-schubert:
As already described in #142 the Summary is leaking memory. This fix will call the compress method of t-digist. To prevent compressing a too small set of measurements and increasing performance, it will call the compress method every 1000 measurements by default. You can also specify another count with option `compressCount` in the Summary options.

### DIFF
--- a/lib/summary.js
+++ b/lib/summary.js
@@ -20,6 +20,8 @@ const {
 } = require('./validation');
 const timeWindowQuantiles = require('./timeWindowQuantiles');
 
+const DEFAULT_COMPRESS_COUNT = 1000; // every 1000 measurements
+
 class Summary {
 	/**
 	 * Summary
@@ -78,6 +80,8 @@ class Summary {
 		this.percentiles = config.percentiles;
 		this.hashMap = {};
 		this.labelNames = config.labelNames || [];
+
+		this.compressCount = config.compressCount || DEFAULT_COMPRESS_COUNT;
 
 		if (this.labelNames.length === 0) {
 			this.hashMap = {
@@ -256,6 +260,9 @@ function observe(labels) {
 
 		summaryOfLabel.td.push(labelValuePair.value);
 		summaryOfLabel.count++;
+		if (summaryOfLabel.count % this.compressCount === 0) {
+			summaryOfLabel.td.compress();
+		}
 		summaryOfLabel.sum += labelValuePair.value;
 		this.hashMap[hash] = summaryOfLabel;
 	};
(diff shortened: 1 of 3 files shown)


lib/summary.js
```diff
@@ -256,6 +260,9 @@ function observe(labels) {
 
 		summaryOfLabel.td.push(labelValuePair.value);
 		summaryOfLabel.count++;
+		if (summaryOfLabel.count % (this.compressCount || DEFAULT_COMPRESS_COUNT)) {
```
SimenB:
```suggestion
 		if (summaryOfLabel.count % this.compressCount === 0) {
```

atd-schubert:
@SimenB of course, here is the updated version. Thanks for your work and have a happy new year 🎆🎉🎈!

siimon:
Thanks for your investigation and fixes! The PR looks good to me.. Just out of curiosity, why does this fix the leak?

siimon:
Still missing the update in the changelog btw!

kristok:
@siimon could you merge if all is good?

siimon:
Sorry for the slow updates, but I still want to know what implications this has and why it solves the leak.. Is it a known problem in the tdigest library or are we using it in a wrong way?

firasdib:
@atd-schubert Do you care to inform us :-)?

atd-schubert:
@siimon @firasdib Sorry I was away the last two weeks...

During my research I found out that Summaries are leaking memory while Histograms don't. I wrote a leakage test for it (see: https://github.com/siimon/prom-client/issues/142#issuecomment-425107388).

While analyzing the heap dumps I recognized that t-digest add each value to an array. Of course this will end-up in a leaking behavior. In the description of the t-digest module is mentioned that you should compress the data especially after adding a big dataset. After compressing it will still have an accurate result, but don't have to keep all entries in the array.

In prom-client data is added continuously. Compressing on every entry will end up in a lot of processing and make the module quite slow. In addition, compressing of a small dataset will not have any effect. This is why I decide to compress every 1000 measurements as default.

After making these changes the above mentioned leakage test will not detect a leaking behavior anymore. As long as you add enough measurements, of course.

So in the end - and according to your question - it is a known issue of the t-digest library, but it works as designed and with the compression you are able to eliminate or at least reduce that issue. So maybe it is also some kind of using it the wrong way...

nelsongomes:
@siimon can you make a deploy with this please?

nelsongomes:
@siimon, @SimenB can you please merge this PR?

siimon:
Published in 11.5.3

Issue #187, opened 2018-04-11
remove() method missing on Metrics


thefloweringash:
I'm trying to write an exporter that handles a dynamic set of gauges. For example purposes, an exporter that reports the free disk space on mounted filesystems like `node_filesystem_avail` from the node exporter is approximately equivalent. When a filesystem is mounted, I can call `labels('mountpoint', '/')` and write to the returned object. When a filesystem is unmounted, I can't see a way to remove only this labeling. There's only `clear()`, which requires recreating all the other labels for remaining mounted filesystems.

The [Prometheus "Writing Client Libraries" section on labels][labels] recommends:

> Metrics with labels SHOULD support a `remove()` method with the same signature as `labels()` that will remove a Child from the metric no longer exporting it, [...]

[labels]: https://prometheus.io/docs/instrumenting/writing_clientlibs/#labels

PR #242, opened 2019-01-18
Add remove() method for handling dynamic labels

lizardruss:
This addresses #187

This implements a `remove()` method on each metric type, based on [Prometheus "Writing Client Libraries" section on labels](https://prometheus.io/docs/instrumenting/writing_clientlibs/#labels)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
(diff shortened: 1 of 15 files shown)


siimon:
Thanks! The implementation looks good to me except that the typedefinition file needs to be updates too.

Oh and update the changelog too!

lizardruss:
I've updated the typings and the changelog 👍

PR #244, opened 2019-01-31
Fix rare ERR_IPC_CHANNEL_CLOSED bug in cluster mode

zbjornson:
A cluster worker can still be in the list of workers even if its IPC channel has closed (e.g. when the worker exits abruptly). Fix by checking that its IPC channel is open before sending a message.

This happens extremely infrequently (less than once a month for us, running several hundred VMs scraping every 10 seconds), but I've tested in an artificial situation that this makes communication safe:

```js
const cluster = require("cluster");

if (cluster.isMaster) {
	for (let i = 0; i < 4; i++) {
		const worker = cluster.fork();
		// THIS SHOULDN'T LOG WITH THE FIX
		worker.on("error", err => console.log("worker error", err));
	}

	cluster.on("exit", (worker, code, signal) => {
		setTimeout(() => cluster.fork(), 500);
	});

	setInterval(() => {
		for (const id in cluster.workers) {
			if (cluster.workers[id].isConnected()) // TOGGLE THIS LINE
				cluster.workers[id].send({type: "hi!"});
		}
	}, 5);
} else {
	setTimeout(() => process.exit(1), 500);
}
```

Also removes some code that was for compat with Node.js pre-v6, since v11 of this lib dropped support for pre-v6.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- Check that cluster worker is still connected before attempting to query it for
+  metrics. (#244)
+
 ### Added
 
 ## [11.2.1]
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -38,8 +38,6 @@ class AggregatorRegistry extends Registry {
 		const requestId = requestCtr++;
 
 		return new Promise((resolve, reject) => {
-			const nWorkers = Object.keys(cluster.workers).length;
-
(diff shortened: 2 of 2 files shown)


kevinolson:
@zbjornson any chance we could get this merged and a new tag pushed out as I'm also running into this issue. Thanks!

zbjornson:
@simen okay with you?

zbjornson:
@SimenB okay with you?

siimon:
Published as 11.3.0

Issue #239, opened 2019-01-04
Prometheus with PM2


KayHS:
Hi

Does anyone have a solution to how i can use prometheus with PM2?

thank you

zbjornson:
See #165 - I don't think it's possible, unfortunately.

KayHS:
Ye ive seen else where people aggregating the values, but i'm not sure how it works.

zbjornson:
I'm not sure exactly what you're asking. As far as I know, PM2 doesn't let you run your own code in the master process, which is required with the way `prom-client` aggregates metrics. There's info on how that works here: https://github.com/siimon/prom-client#usage-with-nodejss-cluster-module. Does that answer your question?

KayHS:
I came across the following package https://github.com/redar9/pm2-cluster-prometheus pm2 module. Which is attempting to do aggregation, however you are saying its not possible.

zbjornson:
Interesting, they reimplemented a lot of the cluster logic from this module using pm2's plugin interface. I see you opened an issue there asking how to use it -- hopefully they respond. It looks like it depends on consul and keymetrics, so it might not be a general-purpose package.

KayHS:
@zbjornson ye hopefully they respond. 

I cannot get metrics from pm2 without paying for their service.

SimenB:
Might be worth a note in the readme, but I agree there's no code change to be made

ChrisLahaye:
I created a library for this purpose and provide an example for usage with prom-client.

See https://github.com/ChrisLahaye/pm2-messages

Issue #165, opened 2017-11-21
How to run with Pm2 module?


fmr683:
If I"m running with pm2 module where should I put this code https://github.com/siimon/prom-client/blob/master/example/cluster.js ? or is it necessary?

zbjornson:
It looks like pm2 doesn't let you run your own code in the master process. Without the ability to do that, you'll be getting metrics from a single worker, which will be incorrect (partial) data if you're using multiple workers.

siimon:
Seems to be answered so I'm closing this one

ChrisLahaye:
I created a library for this purpose and provide an example for usage with prom-client.

See https://github.com/ChrisLahaye/pm2-messages

Issue #140, opened 2017-07-07
High CPU usage since 9 version


SkeLLLa:
Hi. I've noticed high CPU usage by prom-client since version 9 (10 also affected). Version 8 works fine.

Code is pretty simple:
```js
const client = require('prom-client');
client.collectDefaultMetrics({ timeout: 1000 });
```

With default metrics enabled app eats ~60% of CPU. Without - ~0.1%. Node version 8.1.X.
If I have some time to look, I'll provide some additional info.

UPD:
It seems that problem is somewhere in general code, not specific metric, because even "version" metric eats 6-8% CPU. If I enable more metrics, it eats CPU proportionally.

SimenB:
Oh, that's interesting. The only thing that really changed is timestamp support (#69 and #103). Other than that, just how metrics are constructed and ability to merge them, neither of which should impact runtime performance.

Do you see the same thing with metrics without timestamps? Or if you remove timestamp support entirely?

SimenB:
Diff: https://github.com/siimon/prom-client/compare/v8.1.1...v9.0.0

SimenB:
Just removing timestamp from the default metrics made no difference

SkeLLLa:
@SimenB I've run with CPU profile and it shown that most of time it spends in "idle" state (aka `(program)`), so nothing unusual. It shouldn't eat so much CPU.

SimenB:
hmm, I just see the issue on v10, not v9.


sup.js:

```js
require('.').collectDefaultMetrics()

// Keep process alive
setInterval(() => {}, 1000)
```

Then `git checkout v9.1.1 && node sup.js` in one terminal, and `git checkout v10.0.0 && node sup.js` in the other.

![image](https://user-images.githubusercontent.com/1404810/27958503-aec6253c-6323-11e7-9ba5-aa1763a51849.png)

Top one is v10

SkeLLLa:
@SimenB for me it reproduces on both: ^9.0.0 and ^10.0.0

SimenB:
Running on node@8 puts v10 at 12% instead of 16%, but v9.1.1 stays at the same (0.1%).

> @SimenB for me it reproduces on both: ^9.0.0 and ^10.0.0

Odd!

SkeLLLa:
Here's CPU chart. 9 is on the left, 8 on the right:
http://i.imgur.com/0ypaho2.png

SimenB:
v10 `--prof` output

```
[...]

PR #248, opened 2019-03-09
Example of using prom-client with PM2 cluster

yekver:
I have added an example of using prom-client with PM2 cluster.

This is directly related to #80, #165 and #239

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 coverage/
+/example/pm2-cluster.js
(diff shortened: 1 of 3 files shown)


zbjornson:
Thanks for this! Definitely useful info to provide. Slight preference from me to put this in a wiki page instead of in the readme+examples. Having it in the readme+examples I think implies it's something we provide support for (whereas at least I myself have no experience with pm2 and wouldn't be able to maintain the example as pm2's API is updated).

PradeepJaiswar:
@yekver the code works but the aggregation even on a simple service with a pm2 cluster eating heap and CPU. like anything. 

is this related https://github.com/siimon/prom-client/issues/140

Stuck here as many of my services runs using the pm2 cluster because of this I can't put this there.

PR #255, opened 2019-04-17
feat: add option to disable timestamps on defaultMetrics

dolchi21:
Some timestamps can now be turned off by the user.

We have a server that does not have the correct time set.
Prometheus ignores these metrics because timestamps are too far in the future.
We currently can not change the server time so we are not collecting these critical metrics.


dolchi21:
> Works for me! Could you add a test, update the docs, TS typing and changelog?

Is it OK now?

SimenB:

Issue #262, opened 2019-05-20
Summary#observe does not throw when adding additional labels to labelset


samt:
## Reproduction:

```js
const { Summary } = require('prom-client');

const summary = new Summary({
    name: 'foobar',
    help: 'Foo Bar',
    labelNames: ['foo']
});

summary.observe({ foo: 'bar', baz: 'qaz' }, 10);
console.log('did NOT throw');
```

### Expected

```
Error: Added label "baz" is not included in initial labelset: [ 'foo' ]
    at Object.keys.forEach.label (/Users/samt/Repos/prom-client/lib/validation.js:26:10)
    at Array.forEach (native)
    at validateLabel (/Users/samt/Repos/prom-client/lib/validation.js:24:22)
    at /Users/samt/Repos/prom-client/lib/gauge.js:238:3
    at Gauge.set (/Users/samt/Repos/prom-client/lib/gauge.js:91:32)
    at Object.<anonymous> (/Users/samt/Repos/reprod-prom/index.js:19:7)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
```

### Actual

```
did NOT throw
```

---

Issue is caused by [`lib/summary.js#L245`](https://github.com/siimon/prom-client/blob/master/lib/summary.js#L245)

```js
validateLabel(this.labelNames, this.labels);
```

**Should be**:

```js
validateLabel(this.labelNames, labels);
```

PR #263, opened 2019-05-20
Fix labelset validation in Summary

samt:
See #262


SimenB:
Thanks! Can you update the changelog and add a test? I see an old snapshot changed, so I guess we need a test for that case (wrong type) in addition to the missing label name?

PR #264, opened 2019-05-21
fix(types): incorrect return type of collectDefaultMetrics(), should be Timeout in Node

nazar-pc:


SimenB:
Thanks! Mind adding a `/// <reference types="node" />` at the top of the file?

nazar-pc:
Looks like `@types/node` dependency is needed. Ideally in `dependencies`.

SimenB:
Let's do dev dependencies so we don't pollute consumers' code with it. Ref https://github.com/facebook/jest/issues/8092

siimon:
Can you please look into the broken ci?

nazar-pc:
Updated dependencies, should be good now

index.d.ts
```diff
@@ -655,7 +655,7 @@ export interface DefaultMetricsCollectorConfiguration {
  */
 export function collectDefaultMetrics(
 	config?: DefaultMetricsCollectorConfiguration
-): number;
+): NodeJS.Timeout;
```
SimenB: Do we not the the references comment at the top?

nazar-pc: Looks like it is not needed?

nazar-pc: Looks like it is not needed? Actually, it is possible to use something like `ReturnType` to not even specify it explicitly if you want that.

SimenB: `ReturnType<typeof global.setTimeout`? Might work, although since this package only works in Node it should be fine

SimenB: If the references part is not needed, great!

nazar-pc: Refactored to that

index.d.ts
```diff
@@ -655,7 +655,7 @@ export interface DefaultMetricsCollectorConfiguration {
  */
 export function collectDefaultMetrics(
 	config?: DefaultMetricsCollectorConfiguration
-): number;
+): ReturnType<typeof setTimeout>;
```
SimenB:
```suggestion
): ReturnType<typeof setInterval>;
```

SimenB:
If you remove the unneeded package(-lock).json changes (now that we use the return type), we can merge this 🙂

nazar-pc:
I've changed as suggested, tests are good. Globals are strange, but I've tested in my project and it seems to work as it should.

SimenB:

PR #269, opened 2019-06-04
chore: bump dependencies

SimenB:
Been a long time. We still have some `audit` warnings due to the dependency on `npm` via `@clevernature/benchmark-regression`.

/cc @nowells (I don't think you can do anything about it)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,112 @@
 				"@babel/highlight": "^7.0.0"
 			}
 		},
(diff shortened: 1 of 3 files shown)


nowells:
Oh, sorry about that! I'll try to go fix those up ASAP!

SimenB:
I don't think you can do anything since they are bundled with npm. It was more of a heads up than anything 🙂

PR #271, opened 2019-06-07
fix: guard against missing constructor

### LINKED ISSUE #270: TypeError: Cannot read property 'constructor' of null
**Version: prom-client 11.5.0**

**TypeError: Cannot read property 'constructor' of null**
  at aggregateByObjectName (/node_modules/prom-client/lib/metrics/helpers/processMetricsHelpers.js:7:37)
  at config.timestamps (/node_modules/prom-client/lib/metrics/processHandles.js:34:26)
  at initialisedMetrics.forEach.metric (/node_modules/prom-client/lib/defaultMetrics.js:70:47)
  at Array.forEach (<anonymous>:null:null)
  at Timeout.updateAllMetrics [as _onTimeout] (/node_modules/prom-client/lib/defaultMetrics.js:70:22)
  at ontimeout (timers.js:498:11)
  at Timer.unrefdHandle (timers.js:611:5)

yvasiyarov:
Fix for #270 
I can not reproduce this  bug in my environment, but I suppose that it can happen due to the null values returned by process._getActiveHandles().

### DIFF
--- a/lib/metrics/helpers/processMetricsHelpers.js
+++ b/lib/metrics/helpers/processMetricsHelpers.js
@@ -6,7 +6,7 @@ function aggregateByObjectName(list) {
 	for (let i = 0; i < list.length; i++) {
 		const listElement = list[i];
 
-		if (!listElement || typeof listElement.constructor !== 'undefined') {
+		if (!listElement || typeof listElement.constructor === 'undefined') {
 			continue;
 		}
 
--- a/lib/metrics/helpers/processMetricsHelpers.js
+++ b/lib/metrics/helpers/processMetricsHelpers.js
@@ -3,11 +3,17 @@
 function aggregateByObjectName(list) {
 	const data = {};
 
-	for (const key in list) {
-		if (data.hasOwnProperty(list[key].constructor.name)) {
-			data[list[key].constructor.name] += 1;
+	for (let i = 0; i < list.length; i++) {
+		const listElement = list[i];
+
+		if (!listElement || typeof listElement.constructor !== 'undefined') {
+			continue;
+		}
+
+		if (data.hasOwnProperty(listElement.constructor.name)) {
+			data[listElement.constructor.name] += 1;
 		} else {
-			data[list[key].constructor.name] = 1;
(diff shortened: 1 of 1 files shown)


lib/metrics/helpers/processMetricsHelpers.js
```diff
@@ -3,11 +3,13 @@
 function aggregateByObjectName(list) {
 	const data = {};
 
-	for (const key in list) {
-		if (data.hasOwnProperty(list[key].constructor.name)) {
-			data[list[key].constructor.name] += 1;
-		} else {
-			data[list[key].constructor.name] = 1;
+	for (let i = 0; i < list.length; i++) {
+		if (list[i] !== null && typeof list[i]) {
```
SimenB: what does the `typeof` do?

SimenB: What about

```js
for (const key in list) {
	const listElement = list[key];

	if (!listElement) {
		continue;
	}

	if (data.hasOwnProperty(listElement.constructor.name)) {
		data[listElement.constructor.name] += 1;
	} else {
		data[listElement.constructor.name] = 1;
	}
}
```

?

yvasiyarov: @SimenB you are right - this check for "undefined" will never works.
I've rewrote this code as you proposed. New commit squashed with original one

lib/metrics/helpers/processMetricsHelpers.js
```diff
@@ -3,11 +3,17 @@
 function aggregateByObjectName(list) {
 	const data = {};
 
-	for (const key in list) {
-		if (data.hasOwnProperty(list[key].constructor.name)) {
-			data[list[key].constructor.name] += 1;
+	for (let i = 0; i < list.length; i++) {
+		const listElement = list[i];
+
+		if (!listElement || typeof listElement.constructor !== 'undefined') {
```
SimenB:
```suggestion
		if (!listElement || typeof listElement.constructor === 'undefined') {
```

PR #273, opened 2019-06-15
fix: avoid mutation bug in registry

Raynos:
The defaultLabels feature is buggy when used in combination
with the `.labels()` sub counter feature.

We mutate the object in the serialization implementation which
means the next time you observe the value the labels object has
been incorrectly mutated.

The functionality of `.labels()` is supposed to be the caching
of the labels object itself instead of constantly re-using it.

I may actually be the only one using the caching feature of the
`.labels()` method which is why this bug was not caught earlier.

The `.labels()` method is not broken if you use it as

```
histogram.labels(...).observe(int)
```

But is broken if you do

```
class Foo {
    constructor() {
        this.myHistogram = histogram.labels(...)
    }

    method() {
        this.myHistogram.observe(int)
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- fix: avoid mutation bug in registry
+
 ### Added
 
 ## [11.5.0] - 2019-06-04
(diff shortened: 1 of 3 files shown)


Raynos:
I've updated changelog and added a test that fails without my patch.

SimenB:
https://github.com/siimon/prom-client/releases/tag/v11.5.2

Raynos:
Awesome.

PR #274, opened 2019-06-17
feat: implement GC metrics collection without native(C++) modules

yvasiyarov:
I've implemented GC monitoring using perf_hooks builtin module (no dependencies on third party native modules).
Unfortunately perf_hooks module does not provide ability to measure heap size before GC, so I do not report amount of garbage collected( like https://github.com/SimenB/node-prometheus-gc-stats does)

But this implementation reports GC duration percentiles(with split by GC type) which can be quite useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - chore: add test for `process_start_time_seconds`
 
 ### Added
+- `nodejs_gc_runs` metric to the `collectDefaultMetrics()`. It counts number of GC runs with split by GC type.
+- `nodejs_gc_duration_summary` metric to the `collectDefaultMetrics()`. It counts 0.5, 0.75, 0.9, 0.99 percentiles of GC duration (in seconds).
 
 ## [11.5.3] - 2019-06-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- `nodejs_gc_duration_seconds` histogram metric to the `collectDefaultMetrics()`. Default buckets are `[0.001, 0.01, 0.1, 1, 2, 5]`. You can pass custom buckets inside configuration object: `collectDefaultMetrics({gcDurationBuckets:[0.1, 0.2, 0.3]})`
+- feat: implement GC metrics collection without native(C++) modules.
 
 ## [11.5.3] - 2019-06-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- `nodejs_gc_duration_seconds` histogram metric to the `collectDefaultMetrics()`. Default buckets are [0.001, 0.01, 0.1, 1, 2, 5]. You can pass custom buckets inside configuration object: collectDefaultMetrics({timeout: 10000, gcDurationBuckets:[0.001, 0.005, 0.01,.....]})
+- `nodejs_gc_duration_seconds` histogram metric to the `collectDefaultMetrics()`. Default buckets are `[0.001, 0.01, 0.1, 1, 2, 5]`. You can pass custom buckets inside configuration object: `collectDefaultMetrics({gcDurationBuckets:[0.1, 0.2, 0.3]})`
 
 ## [11.5.3] - 2019-06-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - chore: add test for `process_start_time_seconds`
 
 ### Added
-- `nodejs_gc_runs` metric to the `collectDefaultMetrics()`. It counts number of GC runs with split by GC type.
-- `nodejs_gc_duration_summary` metric to the `collectDefaultMetrics()`. It counts 0.5, 0.75, 0.9, 0.99 percentiles of GC duration (in seconds).
+
+- `nodejs_gc_duration_seconds` histogram metric to the `collectDefaultMetrics()`. Default buckets are [0.001, 0.01, 0.1, 1, 2, 5]. You can pass custom buckets inside configuration object: collectDefaultMetrics({timeout: 10000, gcDurationBuckets:[0.001, 0.005, 0.01,.....]})
 
 ## [11.5.3] - 2019-06-27
 
--- a/example/server.js
+++ b/example/server.js
@@ -80,7 +80,10 @@ server.get('/metrics/counter', (req, res) => {
 });
 
 //Enable collection of default metrics
-require('../').collectDefaultMetrics();
+require('../').collectDefaultMetrics({
+	timeout: 10000,
+	gcDurationBuckets: [0.001, 0.005, 0.01]
+});
 
 console.log('Server listening to 3000, metrics exposed on /metrics endpoint');
 server.listen(3000);
--- a/example/server.js
+++ b/example/server.js
@@ -82,7 +82,7 @@ server.get('/metrics/counter', (req, res) => {
 //Enable collection of default metrics
 require('../').collectDefaultMetrics({
 	timeout: 10000,
-	gcDurationBuckets: [0.001, 0.005, 0.01]
+	gcDurationBuckets: [0.001, 0.01, 0.1, 1, 2, 5] // These are the default buckets.
 });
 
 console.log('Server listening to 3000, metrics exposed on /metrics endpoint');
--- a/example/server.js
+++ b/example/server.js
@@ -56,6 +56,19 @@ if (cluster.isWorker) {
 	}, 2000);
 }
 
+// Generate some garbage
+const t = [];
+setInterval(() => {
+	for (let i = 0; i < 100; i++) {
+		t.push(new Date());
+	}
+}, 10);
+setInterval(() => {
+	while (t.length > 0) {
+		t.pop();
+	}
+});
+
 server.get('/metrics', (req, res) => {
 	res.set('Content-Type', register.contentType);
 	res.end(register.metrics());
--- a/lib/defaultMetrics.js
+++ b/lib/defaultMetrics.js
@@ -11,6 +11,7 @@ const processRequests = require('./metrics/processRequests');
 const heapSizeAndUsed = require('./metrics/heapSizeAndUsed');
 const heapSpacesSizeAndUsed = require('./metrics/heapSpacesSizeAndUsed');
 const version = require('./metrics/version');
+const gc = require('./metrics/gc');
 const { globalRegistry } = require('./registry');
 const { printDeprecationCollectDefaultMetricsNumber } = require('./util');
 
@@ -25,7 +26,8 @@ const metrics = {
 	processRequests,
 	heapSizeAndUsed,
 	heapSpacesSizeAndUsed,
-	version
+	version,
+	gc
 };
 const metricsList = Object.keys(metrics);
 
--- a/lib/metrics/gc.js
+++ b/lib/metrics/gc.js
@@ -1,5 +1,4 @@
 'use strict';
-const Counter = require('../counter');
 const Histogram = require('../histogram');
 
 let perf_hooks;
@@ -11,21 +10,8 @@ try {
 	// node version is too old
 }
 
-const NODEJS_GC_RUNS_TOTAL = 'nodejs_gc_runs_total';
-const NODEJS_GC_DURATION = 'nodejs_gc_duration';
-const DEFAULT_GC_DURATION_BUCKETS = [
-	0.001,
-	0.005,
-	0.01,
-	0.025,
-	0.05,
-	0.1,
-	0.25,
-	0.5,
-	1,
-	2.5,
-	5
-];
+const NODEJS_GC_DURATION_SECONDS = 'nodejs_gc_duration_seconds';
+const DEFAULT_GC_DURATION_BUCKETS = [0.001, 0.01, 0.1, 1, 2, 5];
 
 const kinds = [];
 kinds[perf_hooks.constants.NODE_PERFORMANCE_GC_MAJOR] = 'major';
@@ -42,17 +28,10 @@ module.exports = (registry, config = {}) => {
 	const buckets = config.gcDurationBuckets
 		? config.gcDurationBuckets
 		: DEFAULT_GC_DURATION_BUCKETS;
-	const gcCount = new Counter({
-		name: namePrefix + NODEJS_GC_RUNS_TOTAL,
-		help:
-			'Count of garbage collections. kind label is one of major, minor, incremental or weakcb.',
-		labelNames: ['kind'],
-		registers: registry ? [registry] : undefined
-	});
 	const gcHistogram = new Histogram({
-		name: namePrefix + NODEJS_GC_DURATION,
+		name: namePrefix + NODEJS_GC_DURATION_SECONDS,
 		help:
-			'Histogram of garbage collections. kind label is one of major, minor, incremental or weakcb.',
+			'Garbage collection duration by kind, one of major, minor, incremental or weakcb.',
 		labelNames: ['kind'],
 		buckets,
 		registers: registry ? [registry] : undefined
@@ -62,7 +41,6 @@ module.exports = (registry, config = {}) => {
 		const entry = list.getEntries()[0];
 		const labels = { kind: kinds[entry.kind] };
 
-		gcCount.inc(labels, 1);
 		// Convert duration from milliseconds to seconds
 		gcHistogram.observe(labels, entry.duration / 1000);
 	});
@@ -73,4 +51,4 @@ module.exports = (registry, config = {}) => {
 	return () => {};
 };
 
-module.exports.metricNames = [NODEJS_GC_RUNS_TOTAL, NODEJS_GC_DURATION];
+module.exports.metricNames = [NODEJS_GC_DURATION_SECONDS];
--- a/lib/metrics/gc.js
+++ b/lib/metrics/gc.js
@@ -1,0 +1,78 @@
+'use strict';
+const Counter = require('../counter');
+const Summary = require('../summary');
+
+let perf_hooks;
+
+try {
+	// eslint-disable-next-line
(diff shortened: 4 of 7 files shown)


lib/metrics/gc.js
```diff
@@ -0,0 +1,78 @@
...
+			gcKindName = 'unknown';
+			break;
+	}
+	return gcKindName;
+}
+
+module.exports = (registry, config = {}) => {
+	if (!semver.gte(process.version, '8.5.0')) {
+		return () => {};
+	}
```
zbjornson: Instead of checking `process.version`, is it sufficient to test `if (!perf_hooks)`? That would get rid of the new dependency too.

lib/metrics/gc.js
```diff
@@ -0,0 +1,78 @@
...
+		name: namePrefix + NODEJS_GC_RUNS,
+		help:
+			'Count of garbage collections. gc_type label is one of major, minor, incremental or weakcb.',
+		labelNames: ['gc_type'],
+		registers: registry ? [registry] : undefined
+	});
+	const gcSummary = new Summary({
+		name: namePrefix + NODEJS_GC_DURATION_SUMMARY_MS,
+		help:
+			'Summary of garbage collections(in ms). gc_type label is one of major, minor, incremental or weakcb.',
```
zbjornson: Prometheus encourages using base units (here: seconds, see https://prometheus.io/docs/practices/naming/#base-units).

lib/metrics/gc.js
yvasiyarov: fixed

lib/metrics/gc.js
yvasiyarov: fixed

yvasiyarov:
@zbjornson I've fixed unit tests. Regarding naming of metrics I do not have strong opinion.
From one side added metrics serve same function as https://github.com/SimenB/node-prometheus-gc-stats module, so it might be a good idea to use same names for metrics.

On the other side this metrics is not drop in replacement for node-prometheus-gc-stats . As I said some metrics is available only at node-prometheus-gc-stats (and this information is not available through perf_hooks interface). So there might be a reason to have both sets of metrics available at the same time. 

If we will use same names for metrics - than prom-client will be conflicting with  node-prometheus-gc-stats module.

yvasiyarov:
@SimenB what can I do to get this PR merged ? 
if you want me to fix something, please, let me know.

yvasiyarov:
@zbjornson sorry for disturbing you, but how can I have this PR merged ?

lib/metrics/gc.js
siimon: The counter name should have the total suffix as @zbjornson mentioned

siimon:
I’m not a huge fan of adding checks in prom-client for detecting other libraries, in this case gc-stats. If we could go around it with different naming that would be great.

I’ve not read up on the native gc stats api, but what is preferred between this and the “old” solution? When should you use this and not the other? Should there be some guidance in the docs on this?

lib/metrics/gc.js
zbjornson: I didn't follow too closely, but in #284 was the point that `_total` is an anti-pattern in Prometheus? I only mentioned it here because node-prometheus-gc-stats uses the `_total` suffix.

zbjornson:
Re-reviewing... sorry for letting this sit.

The `gc-stats` module (used by node-prometheus-gc-stats) measures "bytes reclaimed" just by calling v8's `GetHeapStatistics` in the GC prologue and epilogue; it's not something v8 provides in the GC epilogue. I tested if `process.memoryUsage().heapUsed` could be used to approximate that, but it can't (there's no userland GC prologue so you're measuring newly allocated, live objects too). bytes_reclaimed is a useful metric; I think it would be worth opening an issue in nodejs/node suggesting it be added to `perf_hooks`.

(FWIW, the GC duration reported by perf_hooks and gc-stats is very similar.)

That means the current benefit of this PR is no native add-on for users to deal with and duration reported as a Summary (see also [issue in node-prometheus-gc-stats to convert to a histogram](https://github.com/SimenB/node-prometheus-gc-stats/issues/3)), but the downside is no bytes_reclaimed metric.

One possible route forward is to essentially move node-prometheus-gc-stats into this module: add a note in the readme like "GC duration is available as a default metric. If you also desire the number of bytes reclaimed by each collection, you can install `gc-stats` and this module will use it automatically." Then this module would have `try { require("gc-stats") } catch {}` -- i.e. silently fall back to just using perf-hooks. Then we wouldn't have to worry about metric name conflicts between this and node-prometheus-gc-stats. I'm happy to help with this change if you'd like.

TBD is the `_total` suffix. I guess leave it off since that's best practice, although inconsistent with the rest of this module at the moment. Alternatively, add it for now and correct it in #284...

zbjornson:
> The gc-stats module (used by node-prometheus-gc-stats) measures "bytes reclaimed" just by calling v8's GetHeapStatistics in the GC prologue and epilogue

Now that I think about it, I think this method of determining bytes reclaimed is not accurate because [v8's GC is concurrent](https://v8.dev/blog/trash-talk). I've asked [hooraybuffer](https://twitter.com/hooraybuffer?lang=en) on Twitter to confirm. Given that, I think this perf_hooks impl is the best route forward.

yvasiyarov:
@zbjornson  thank you for spending your time on this PR and research you have made regarding "bytes reclaimed"  metric. 
Now, what should I do to make this PR merged ?
Do I need to add "_total" suffix or not ?

zbjornson:
@siimon @SimenB how would you feel about deprecating node-prometheus-gc-stats due to the (likely) accuracy issue with "bytes reclaimed" mentioned above, and instead just monitoring GC duration in this module? (hooryabuffer didn't reply unfortunately.)

lib/metrics/gc.js
```diff
@@ -0,0 +1,78 @@
...
+
+	const namePrefix = config.prefix ? config.prefix : '';
+	const gcCount = new Counter({
+		name: namePrefix + NODEJS_GC_RUNS,
+		help:
+			'Count of garbage collections. gc_type label is one of major, minor, incremental or weakcb.',
+		labelNames: ['gc_type'],
+		registers: registry ? [registry] : undefined
+	});
+	const gcSummary = new Summary({
```
sam-github: why a summary and not a histogram? I thought histograms are generally preferred because they can be aggregated across instances.

lib/metrics/gc.js
```diff
@@ -0,0 +1,78 @@
...
+		name: namePrefix + NODEJS_GC_DURATION_SUMMARY,
+		help:
+			'Summary of garbage collections. gc_type label is one of major, minor, incremental or weakcb.',
+		labelNames: ['gc_type'],
```
sam-github: `type`? `gc_` seems redundant for a `nodejs_gc_duration` metric. Or simply `kind`? I like to reuse names, so that I don't have to remember that what nodejs docs call "kind" prom-client calls "type".

lib/metrics/gc.js
```diff
@@ -0,0 +1,78 @@
...
+		maxAgeSeconds: 600,
```
sam-github: Is a max age necessary? Should the user be allowed to configure one if they want? Histograms don't need this config, though (IIRC). https://nodejs.org/api/perf_hooks.html#perf_hooks_performanceentry_kind

lib/metrics/gc.js
```diff
@@ -0,0 +1,78 @@
...
+		ageBuckets: 5,
+		percentiles: [0.5, 0.75, 0.9, 0.99],
+		registers: registry ? [registry] : undefined
+	});
+
+	const obs = new perf_hooks.PerformanceObserver(list => {
+		const entry = list.getEntries()[0];
+		const labels = { gc_type: gcKindToString(entry.kind) };
```
sam-github: A numeric lookup into an array might be faster and more readable than a switch statement (offered without bencharking evidence, though :-)/.
```
const kinds = []
kinds[perf_hooks.constants.NODE_PERFORMANCE_GC_MAJOR] = 'major';
// ...etc
        const kind = kinds[entry.kind] || 'unknown';
        const labels = { kind }
```

lib/metrics/gc.js
```diff
@@ -0,0 +1,78 @@
...
+		registers: registry ? [registry] : undefined
+	});
+	const gcSummary = new Summary({
+		name: namePrefix + NODEJS_GC_DURATION_SUMMARY,
+		help:
+			'Summary of garbage collections. gc_type label is one of major, minor, incremental or weakcb.',
+		labelNames: ['gc_type'],
+		maxAgeSeconds: 600,
+		ageBuckets: 5,
+		percentiles: [0.5, 0.75, 0.9, 0.99],
```
sam-github: histograms do require buckets, though, and perhaps users should be able to override the defaults?

lib/defaultMetrics.js
```diff
@@ -25,7 +26,8 @@ const metrics = {
 	processRequests,
 	heapSizeAndUsed,
 	heapSpacesSizeAndUsed,
-	version
+	version,
```
sam-github: using trailling comma style avoids this kind of diff churn, but that's a matter of project style, not for this PR

lib/metrics/gc.js
siimon: Yeah, probably not needed since histograms is preferred and should be used

siimon:
I see it as a great fit to deprecate the gc-stats module if it could be added into prom-client with no third party dependencies!

Sorry for taking so long on this, will try and keep an eye on this! @yvasiyarov if you could make the requested changes @sam-github pointed out that would be great and after that I think we can merge this one.

@sam-github thanks for taking your time on reviewing this!

lib/defaultMetrics.js
SimenB: (I'll put up a PR for it)

yvasiyarov:
@siimon @sam-github thank you for the feedback. I will make requested changes in a few days

lib/metrics/gc.js
yvasiyarov: fixed

lib/metrics/gc.js
yvasiyarov: fixed

lib/metrics/gc.js
yvasiyarov: fixed

lib/metrics/gc.js
yvasiyarov: accordingly to promtool check, "_total" suffix is mandatory for counters. So I've added "total" suffix

lib/metrics/gc.js
yvasiyarov: fixed

lib/metrics/gc.js
yvasiyarov: fixed

yvasiyarov:
@siimon @SimenB @zbjornson @sam-github guys, please, have a look at updated PR

lib/defaultMetrics.js
sam-github: Pls mark resolved.

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - chore: add test for `process_start_time_seconds`
 
 ### Added
+- `nodejs_gc_runs` metric to the `collectDefaultMetrics()`. It counts number of GC runs with split by GC type.
```
sam-github: Unnecessary duplicate metric? The histogram contains the counts, so a seperate metrics is a waste of storage. gc counts (and rates) can be graphed from the histogram.

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
...
+- `nodejs_gc_duration_summary` metric to the `collectDefaultMetrics()`. It counts 0.5, 0.75, 0.9, 0.99 percentiles of GC duration (in seconds).
```
sam-github: Docs are out of date. They should describe the gcDurationBuckets default and how to change.

lib/metrics/gc.js
```diff
@@ -0,0 +1,76 @@
...
+
+try {
+	// eslint-disable-next-line
+	perf_hooks = require('perf_hooks');
+} catch (e) {
+	// node version is too old
+}
+
+const NODEJS_GC_RUNS_TOTAL = 'nodejs_gc_runs_total';
+const NODEJS_GC_DURATION = 'nodejs_gc_duration';
```
sam-github: I think the naming convention here would be `nodejs_gc_duration_seconds`,  like `http_request_duration_seconds` from https://prometheus.io/docs/practices/naming/#metric-names

lib/metrics/gc.js
```diff
@@ -0,0 +1,76 @@
...
+const DEFAULT_GC_DURATION_BUCKETS = [
```
sam-github: I'm having troulbe finding it , but I think recommended label cardinality is `<= 10`. GC durations should be low, and anything approaching 1 second is likely a severe problem. It'd be nice to have a reasonable range, if you have enough production experience to know what that is. I suggest that the cardinality be brought down to 7, these aren't supposed to be very fine-grained, maybe?
```
[0.001, 0.01, 0.1, 1, 2, 5]
```
Or pehaps no 5, but one more bucket at the low end?

lib/metrics/gc.js
```diff
@@ -0,0 +1,76 @@
...
+let perf_hooks;
+
+try {
+	// eslint-disable-next-line
+	perf_hooks = require('perf_hooks');
+} catch (e) {
+	// node version is too old
+}
+
+const NODEJS_GC_RUNS_TOTAL = 'nodejs_gc_runs_total';
```
sam-github: Naming looks good. I've the impression that prometheus discourages duplicate collection of metrics, and the totals are already available in the duration counts below, but I'm not expert enough to know for sure.

lib/metrics/gc.js
```diff
@@ -0,0 +1,76 @@
...
+		name: namePrefix + NODEJS_GC_RUNS_TOTAL,
+		help:
+			'Count of garbage collections. kind label is one of major, minor, incremental or weakcb.',
+		labelNames: ['kind'],
+		registers: registry ? [registry] : undefined
+	});
+	const gcHistogram = new Histogram({
+		name: namePrefix + NODEJS_GC_DURATION,
+		help:
+			'Histogram of garbage collections. kind label is one of major, minor, incremental or weakcb.',
```
sam-github:
```suggestion
			'Garbage collection duration by kind, one of major, minor, incremental or weakcb.',
```
"Histogram" is already documented in the /metrics output type.

lib/metrics/gc.js
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	const namePrefix = config.prefix ? config.prefix : '';
+	const buckets = config.gcDurationBuckets
+		? config.gcDurationBuckets
+		: DEFAULT_GC_DURATION_BUCKETS;
+	const gcCount = new Counter({
+		name: namePrefix + NODEJS_GC_RUNS_TOTAL,
+		help:
+			'Count of garbage collections. kind label is one of major, minor, incremental or weakcb.',
```
sam-github:
```suggestion
			'Garbage collection runs by kind, one of major, minor, incremental or weakcb.',
```

CHANGELOG.md
yvasiyarov: yes, you are right.  Histogram already has nodejs_gc_duration_seconds_count metric with the same information

CHANGELOG.md
yvasiyarov: updated documentation. I also used custom GC duration buckets in the example/server.js

lib/metrics/gc.js
yvasiyarov: fixed

lib/metrics/gc.js
yvasiyarov: I think what is "reasonable range" really depends on what your software is doing.  I think [0.001, 0.01, 0.1, 1, 2, 5] buckets can be a good point to start with. Updated defaults

lib/metrics/gc.js
yvasiyarov: metric removed

lib/metrics/gc.js
yvasiyarov: fixed

lib/metrics/gc.js
yvasiyarov: this metric is removed

yvasiyarov:
@sam-github thank you for your feedback, I really appreciate that.  Please have another look at PR

CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- `nodejs_gc_duration_seconds` histogram metric to the `collectDefaultMetrics()`. Default buckets are [0.001, 0.01, 0.1, 1, 2, 5]. You can pass custom buckets inside configuration object: collectDefaultMetrics({timeout: 10000, gcDurationBuckets:[0.001, 0.005, 0.01,.....]})
```
sam-github:
```suggestion
- `nodejs_gc_duration_seconds` histogram metric to the `collectDefaultMetrics()`. Default buckets are `[0.001, 0.01, 0.1, 1, 2, 5]`. You can pass custom buckets inside configuration object: `collectDefaultMetrics({gcDurationBuckets:[0.1, 0.2, 0.3]})`
```

example/server.js
```diff
@@ -67,7 +80,10 @@ server.get('/metrics/counter', (req, res) => {
 });
 
 //Enable collection of default metrics
-require('../').collectDefaultMetrics();
+require('../').collectDefaultMetrics({
+	timeout: 10000,
+	gcDurationBuckets: [0.001, 0.005, 0.01]
```
sam-github:
```suggestion
	gcDurationBuckets: [0.001, 0.01, 0.1, 1, 2, 5] // These are the default buckets.
```

sam-github:
Sorry, one other suggestion. The docs that got added to the CHANGELOG.md, I think they should go in the README.md, and the CHANGELOG.md should just have a oneliner: Add: `feat: implement GC metrics collection without native(C++) modules`.

yvasiyarov:
@sam-github I've updated both CHANGELOG.md and README.md

siimon:
🎉👍

siimon:
Will try and merge some more PRs and make a release when I’m back on a regular computer on Monday 👍

TranBaVinhSon:
When will this pull request be released ? I need gc metrics in my project. Many thanks !

SimenB:
After #333 lands

PR #275, opened 2019-06-18
feat: collect bootstrap timing metrics

yvasiyarov:
Collect bootstrap timing metrics

### DIFF
--- a/lib/metrics/bootstrapTime.js
+++ b/lib/metrics/bootstrapTime.js
@@ -4,7 +4,7 @@ const Gauge = require('../gauge');
 let perf_hooks;
 
 try {
-	// eslint-disable-next-line
+	// eslint-disable-next-line node/no-unsupported-features/node-builtins
 	perf_hooks = require('perf_hooks');
 } catch (e) {
 	// node version is too old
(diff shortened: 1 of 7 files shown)


lib/metrics/bootstrapTime.js
```diff
@@ -0,0 +1,84 @@
+'use strict';
+const Gauge = require('../gauge');
+
+let perf_hooks;
+
+try {
+	// eslint-disable-next-line
```
SimenB: can you disable the specific rule?

lib/metrics/bootstrapTime.js
```diff
@@ -0,0 +1,84 @@
...
+		help: 'Node.js event loop start time (in seconds).',
+		registers: registry ? [registry] : undefined
+	});
+
+	return () => {
+		const entry = perf_hooks.performance.nodeTiming;
+		const now = Date.now();
+
+		if (entry.nodeStart !== -1) {
+			nodeStart.set({}, entry.nodeStart, now);
```
SimenB: we have an option for disabling timestamps - mind adding code handling it?

SimenB (changes requested):
Thanks! I think this looks pretty good, just missing code for dealing with disabled timestamps

lib/metrics/bootstrapTime.js
yvasiyarov: fixed

lib/metrics/bootstrapTime.js
yvasiyarov: fixed

lib/metrics/bootstrapTime.js
yvasiyarov: I'm sorry, but its looks like I can not disable only specific rule.
For node 6.x I should use 
`//eslint-disable-next-line node/no-missing-require`
But for newer versions of node I should use  
`//eslint-disable-next-line node/no-unsupported-features/node-builtins`
But I can not specify both rules. So I rolled back this to just 
`//eslint-disable-next-line `

PR #277, opened 2019-06-19
feat: implement monitoring of callbacks executed by process.nextTick()

yvasiyarov:
Nearly every talk about Node.js performance begins with event loop monitoring.
And one of most often recommendation is - do not run cpu intensive operations in nextTick().
So I thought it will be useful to measure how long does it takes to execute every single callback scheduled by nextTick() and how many ticks are being executed by nextTick().

To make this information actionable I also wanna implement some way to export top longest callbacks executed by nextTick() or somehow export callbacks which executes longer than some threshold. But it's not implemented yet. Any suggestions how this kind of information can be exported to prometheus are welcome :-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 ### Added
+- `nodejs_gc_runs` metric to the `collectDefaultMetrics()`. It counts number of GC runs with split by GC type.
+- `nodejs_gc_duration_summary` metric to the `collectDefaultMetrics()`. It counts 0.5, 0.75, 0.9, 0.99 percentiles of GC duration (in seconds).
 
 ## [11.5.3] - 2019-06-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,16 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 ### Added
+
 - `nodejs_gc_runs` metric to the `collectDefaultMetrics()`. It counts number of GC runs with split by GC type.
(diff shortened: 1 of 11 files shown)


zbjornson:
This is an interesting metric! Do you have any estimate of how much overhead it adds to the running process (performance impact)? Wondering if it should be opt-in/a separate custom metric.

zbjornson:
> I also wanna implement some way to export top longest callbacks executed by nextTick() or somehow export callbacks which executes longer than some threshold

I don't think this is possible when there are multiple callbacks running (which is basically always for a real production server). Callback A will delay callback B and all subsequent callbacks that were scheduled in the current event loop iteration, so it wouldn't be an accurate measurement, I think. You'd probably have to track execution time of the callback handlers more likely, which would be hard.

yvasiyarov:
> This is an interesting metric! Do you have any estimate of how much overhead it adds to the running process (performance impact)? Wondering if it should be opt-in/a separate custom metric.

I don't have numbers in my hands, but anyway I think its a good idea to make it opt-in. Will implement  it today

yvasiyarov:
> > I also wanna implement some way to export top longest callbacks executed by nextTick() or somehow export callbacks which executes longer than some threshold
> 
> I don't think this is possible when there are multiple callbacks running (which is basically always for a real production server). Callback A will delay callback B and all subsequent callbacks that were scheduled in the current event loop iteration, so it wouldn't be an accurate measurement, I think. You'd probably have to track execution time of the callback handlers more likely, which would be hard.

I separately measure execution of every callback. So execution time of callback A measured separately from callback B.

yvasiyarov:
@zbjornson I've made monitoring of process.nextTick optional, disabled by default.

jdmarshall:
#518 is a better option here, now.

PR #280, opened 2019-07-25
Added opportunity add any registry to merge with clusters

### LINKED ISSUE #183: Allow master to add this own metrics
Hi ! 
I work on clustered Node app.
I have some issues when trying to have worker(s) metrics with cluster ones.

Is master process alllowed to expose his own metrics (registered on default Registry) ?

https://github.com/siimon/prom-client/blob/master/lib/cluster.js#L169

rimdus:
Hello. We used cluster in project. Master process make logics to exclusively write to db extra data and used metrics too. This commit allow to add master repository to merge with clusters optionals.

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -17,6 +17,7 @@ const GET_METRICS_REQ = 'prom-client:getMetricsReq';
 const GET_METRICS_RES = 'prom-client:getMetricsRes';
 
 let registries = [Registry.globalRegistry];
+let additionRegistry;
 let requestCtr = 0; // Concurrency control
 let listenersAdded = false;
 const requests = new Map(); // Pending requests for workers' local metrics.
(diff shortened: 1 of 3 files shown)


siimon:
@zbjornson do you have any opinions?

zbjornson:
This will fix #183 👍 and I think is the correct approach. Could you add a test case for it and update the changelog please?

siimon:
Still waiting for a test before merging this one

zbjornson:
Poke @rimdus -- could you add a test please? This is a more versatile solution than #324 so would like to get this in.

lib/cluster.js
```diff
@@ -31,12 +32,16 @@ class AggregatorRegistry extends Registry {
...
+	 * @param {Registry} opt.registry additional registry to merge with
 	 * @return {Promise<string>} Promise that resolves with the aggregated
 	 *   metrics.
 	 */
-	clusterMetrics(callback) {
+	clusterMetrics(callback, opt) {
 		const requestId = requestCtr++;
-
+		if (opt && opt.registry) {
+			additionRegistry = opt.registry;
```
KeithWoods, line 43: I might be missing something here as I'm new to this codebase, but it seems odd to call an instance method (`anAggregatorRegistryInstance.clusterMetrics()`) which has the side effect of affecting a static method (`AggregatorRegistry.aggregate()`). There is an implicit dependency here in that the execution path will ultimately call `AggregatorRegistry.aggregate()` with the hope the right  `additionRegistry` is still set. 

Perhaps this `additionalRegistry` should be stored with the inflight request (perhaps by modifying the shape of data stored in `requests`) state so it's metrics can later be retrieved before the call to `AggregatorRegistry.aggregate()`.

Again I'm pretty new to looking at this, so forgive me if this is off base.

zbjornson: (16 months later)

Fair point, that would be a good improvement.

We already have module-level state with `AggregatorRegistry.setRegistries()`. That should maybe be changed also (in a separate PR).

PR #282, opened 2019-08-16
feat: Return the final duration value from startTimer callback

mxcoder:
I have a use case in which this value its really useful to me for logging purposes, if we return this value in the helper, I could save a few lines of code


SimenB:
This needs a test, and be mentioned in the docs. Should we also return the value from other types of metrics?

Not sure why CI failed...

/cc @zbjornson @siimon

zbjornson:
Seems like a good feature to me. Agree with @SimenB 's comments.

> Should we also return the value from other types of metrics?

Not opposed to it, but not sure the use cases exist either. This PR returns the timer duration (not the whole histogram), so to parallel that, a gauge might return the value by which it was incremented/decremented, which seems less useful.

mxcoder:
Hi, is there something else I can do to push this PR forward? Should I add tests and docs?

SimenB:
>Should I add tests and docs?

yes 🙂

zbjornson:
Ping @mxcoder 🙂 can you add tests and update the docs and changelog please?

mxcoder:
Hey, apologies for the long delay.

I committed the small required adjustments.

Thanks

PR #285, opened 2019-08-28
feat: add normalizeLabelName and normalizeMetricName

edvardchen:


SimenB:
What are these normalizations used for?

Also, please update the docs, [type defs](https://github.com/siimon/prom-client/blob/master/index.d.ts) and changelog

zbjornson:
Ping @edvardchen. If you'd still like this merged, can you let us know the answer to the above please?

edvardchen:
There is no need right now. Closed.

Issue #293, opened 2019-10-09
Lazy-load Node.js Cluster module


laurentgoudet:
[Passenger](https://github.com/phusion/passenger) provides its own clustering, and hence is not compatible with Node.js `cluster` module. In order to improve the developer experience, it provides a non-functional shim while warning developers about the underlying issue:

```
Trace: You required the Node Cluster module, which is incompatible with Passenger, a non-functional shim was returned and your app may still work. However, please remove the related code as soon as possible.
    at Module.require (/usr/share/passenger/helper-scripts/node-loader.js:63:12)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/home/app/webapp/node_modules/prom-client/lib/cluster.js:11:17)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at Module.require (/usr/share/passenger/helper-scripts/node-loader.js:80:25)
```

It would be nice if `prom-client` would lazy-load the Node.js `cluster` module until its aggregation features are used, i.e. until an `AggregatorRegistry` instance is created, in order to provide a smooth experience when integrating with [Passenger](https://github.com/phusion/passenger).

For reference a similar issue was reported & fixed on [log4js-node](https://github.com/log4js-node/log4js-node) 2 years ago, although I don't think a `disableClustering: true` flag is needed here as one can simply not use the `AggregatorRegistry` in the first place when not using Node.js clustering.

PR #294, opened 2019-10-09
Lazy-load cluster module to fix Passenger support

laurentgoudet:
Fixes https://github.com/siimon/prom-client/issues/293.

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -180,18 +180,18 @@ function addListeners() {
 				}
 			}
 		});
-	} else {
-		// Respond to master's requests for worker's local metrics.
-		process.on('message', message => {
-			if (message.type === GET_METRICS_REQ) {
-				process.send({
-					type: GET_METRICS_RES,
-					requestId: message.requestId,
-					metrics: registries.map(r => r.getMetricsAsJSON())
-				});
-			}
-		});
 	}
 }
+
+// Respond to master's requests for worker's local metrics.
+process.on('message', message => {
+	if (cluster().isWorker && message.type === GET_METRICS_REQ) {
+		process.send({
+			type: GET_METRICS_RES,
+			requestId: message.requestId,
+			metrics: registries.map(r => r.getMetricsAsJSON())
+		});
+	}
+});
 
 module.exports = AggregatorRegistry;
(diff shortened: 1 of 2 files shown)


lib/cluster.js
```diff
@@ -8,10 +8,12 @@
  * cluster master.
  */
 
-const cluster = require('cluster');
 const Registry = require('./registry');
 const { Grouper } = require('./util');
 const { aggregators } = require('./metricAggregators');
+// We need to lazy-load the 'cluster' module as some application servers -
+// namely Passenger - crash when it is imported.
+const cluster = () => require('cluster');
```
SimenB: could we do closer to how babel lazy loading works?

e.g.

```js
function cluster() {
  const data = require('cluster');

  cluster = function() {
    return data;
  };

  return data;
}
```

while `require` is cached, avoiding the require-call at all seems prudent

lib/cluster.js
```diff
@@ -174,20 +176,18 @@ function addListeners() {
 				}
 			}
 		});
+	} else {
+		// Respond to master's requests for worker's local metrics.
+		process.on('message', message => {
```
SimenB: does it still work when moved into this `addListeners` function?

lib/cluster.js
laurentgoudet: Yep I did it that way because `require` is cached, but sure I'll change it

lib/cluster.js
laurentgoudet: Hm yeah you're right. Since that code listens for master's requests of type `GET_METRICS_REQ`, which are only sent when  `clusterMetrics()` is called, I assumed it would be fine, e.g. https://github.com/siimon/prom-client/blob/master/example/cluster.js would work.

However if one is instantiating the `AggregatorRegistry` only for the `master` process, the workers won't have the listener setup. I can fix it by moving the `isWorker` check inside the `process.on` callback though, so that the `cluster` module is not instantiated until an inter-process message is received.

laurentgoudet:
Heya, any chance to get this merged :)? I'd love to drop my fork ^^

zbjornson:
Sorry for the delay. Can you update the changelog please? Then this is good to land.

laurentgoudet:
done :)

Issue #298, opened 2019-10-23
Relax necessity of declaring labelset in advance


dt-rush:
There is no inherent restriction in prometheus for this. Labels can appear or disappear from the timeseries a node exports. This restriction in this JS client appears to be a consequence of choices made in the design of the internal `hashmap` property used in the 4 metrics classes, the `utils.setValue()` function, and possibly the `utils.getLabels()` function. This issue is a placeholder for further investigation.

zbjornson:
Not opposed to relaxing the requirement, but also note that https://github.com/siimon/prom-client/pull/299#issuecomment-550011185 made it so Typescript/ts-lint users will get a warning if they try to use a label that wasn't pre-declared.

dt-rush:
Thanks for pointing that out @zbjornson.

I'm working on a PR and tests for this atm.

For further justification for this change, I cite the robustperception.io blog post on predeclaring metrics:

> The first reason relates to what was previously covered in Existential issues with metrics, metrics that only some times exist are difficult to deal with in PromQL. By declaring the metric before your application uses it, the client library can initialise it to 0. This applies even if the Counter is never subsequently incremented. However this doesn't work for metrics with labels: **the client library doesn't know what labels might come up**, so you have to do it yourself.

(emphasis added)

dt-rush:
@zbjornson as mentioned in the discussion in #299, that'll be reverted as it adds to the unnecessary rigidity of using this client.

dt-rush:
That is all sorted nicely, however, there is one more issue to bring up: 

In the [Writing client libaries # Labels](https://prometheus.io/docs/instrumenting/writing_clientlibs/#labels) docs, we read:

> Client libraries MUST NOT under any circumstances allow users to have different label names for the same metric for Gauge/Counter/Summary/Histogram or any other Collector offered by the library.

The wording of this is not exactly easy to interpret, since its only possible meaning seems a contradiction with what is said elsewhere, that the appearance of a new labelset will create a new timeseries (that is, there's no problem having different labelNames for the same metric). There is no problem even in the current system of pre-declaring a labelset, `{a, b, c, d}` and then at various times exporting metrics labelled with subsets: `{a, b}`, `{c, d}`,`{c}`, `{b, d}`, etc... that is, "different label names for the same metric".

This is brought up before [in this SO question](https://stackoverflow.com/questions/57261955/different-set-of-labels-for-the-same-counter-name) where the user similarly brings up this contradiction.

There is really no difference, if exporting

[...]

PR #299, opened 2019-11-05
fix: TypeScript declaration - strict label values

kobiburnley:


SimenB:
Hey! Could you explain what this changes in usage?

kobiburnley:
@SimenB hello!

This gives TypeScript users stricter type checking on labels (verifies label keys against the labels array provided in the metric constructor)

```TypeScript
type LabelValues<T extends string> = Partial<Record<T, string | number>>;

class Counter<T extends string> {
    constructor(public labelNames: T[]) {

    }

    inc(LabelValues: LabelValues<T>) {

    }
}

const counter = new Counter(["method", "statusCode"])

counter.inc({
    method: "",
    statusCode: "sdgsgd"
})

counter.inc({
    method: "",
    statuesCode: "sdgsgd" // won't compile :)
})
```

[demo](http://www.typescriptlang.org/play/?ssl=23&ssc=3&pln=1&pc=1#code/C4TwDgpgBAMghgIwgGwGp2QVwgZwDwAqUEAHsBAHYAmOUOwATgJYUDmAfFALxQAKcDYEwx4AShADGAewZVCAGjqMWrKAB8oFTAFskDduwDcAKGMTkcHLQDCUzBXINCxMpRpLmbTgG9jUf1DSFPQMmBLAMgAUYJgIyEwSUBZIyABycNq4AFxQBADaALoAlFC+fgEAvqYBUCwSkfAp6FjZsIgozdj4BOwlZTVVVWZSwcCBdg4QDNyaEADuULb2jpF5AESZwAAWUlRrimv0cMCYOLZUEGvFptLLUwB0dZG+NZs7VDlr++X+RydnuwgnxwVFYOFYe2MFSKNwmjkeFHqLwCb12n2+NT+XXOQKgh1B4L2UAA9MSoHMRgByMbSbRgJjIaBZGHQoA)

SimenB:
Very nice, thank you! Could you update the changelog as well?

kobiburnley:
@SimenB done.

ankon:
This change now requires quite some changes in user code, from
```typescript
interface Metrics {
	LAST_PING_RECEIVED_ID: prometheus.Gauge;
}
```
to
```typescript
interface Metrics {
	LAST_PING_RECEIVED_ID: prometheus.Gauge<string>;
}
```

I'm not sure why the type `T` is required there, would someone be using something else than a `string` here? Could it be defaulted to `T extends string = string`?

sam-github:
This makes me wonder if some tests actually written in typescript against the ts defs would be useful, to confirm that the ts defs are actually valid and useful (speaking as a non-ts developer who has had to update the defs lately, and did so with trepidation).

siimon:
Yeah you are probably right.. I think the only test right now just sees if the typefile compiles, not that it really works as expected

ankon:
> This makes me wonder if some tests actually written in typescript against the ts defs would be useful, to confirm that the ts defs are actually valid and useful (speaking as a non-ts developer who has had to update the defs lately, and did so with trepidation).

As a first step: Instead of writing "tests" you could write an example, and check that the compiler is happy about it, i.e. can compile the example code against the type descriptions. When converting to typescript we even started one step earlier and simply let tsc compile the definitions as a check, to kill out the odd syntax issues and make developers more comfortable with the .d.ts files.

But, that's probably a different issue -- for the API change related to the now-required type parameters, do you think I should whip together a PR and try to "undo" this?

SimenB:
https://www.npmjs.com/package/tsd is pretty nice for writing type tests.

But yeah, defaulting to `string` should work fine, I think?

ankon:
I think so too, yes.

zbjornson:
I'd go for Microsoft's dtslint for testing and linting declarations, https://github.com/microsoft/dtslint. Easy to use, handles different versions of tsc well, and I think it's that all of DefinitelyTyped uses.

dt-rush:
This might be a nice feature to be able to enable *should the developer choose to enable it*. As discussed in #298 there's actually no reason to require pre-declaration of labelsets and the more that we double-down on this necessity the less flexible the client library is for a whole variety of use cases where labelsets are not known in advance.

Unfortunately, it's added in such a way that it's not optional, this type information is a permanent part of what the typescript compiler desires. Therefore in line with a PR to solve #298 I'll be including a reversion of this (a reversion was also desired for an additional valid reason by @ankon above).

Issue #311, opened 2020-01-03
process_max_fds is reporting system limit, not the process limit


sam-github:
I think  this is a bug.

`process_max_fds` is using [`/proc/sys/fs/file-max`](https://github.com/siimon/prom-client/blob/cc9ce780ce59207c48fec79c539de947801348ee/lib/metrics/processMaxFileDescriptors.js#L28) which is

> a system-wide limit on the number of open files for all processes

This is the _OS_ limit, not the _process_ limit.

The process limit can be read from `/proc/self/limits`.

See http://man7.org/linux/man-pages/man5/proc.5.html for the /proc docs, which can't be deep linked into, sorry.

I might be able to find time to fix this if its agreed its wrong, but not today.

SimenB:
I think I wrote that one a couple of years ago, and I didn't purposefully take the OS limit. PR very much welcome 🙂

I _think_ I based it on https://github.com/prometheus/client_java/blob/master/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/StandardExports.java#L82 which comes from `osBean`, which I probably just googled and went with first I found 😅

sam-github:
I check, go's promclient uses `/proc/[pid]/limits`: 
https://github.com/prometheus/procfs/blob/496ec92a4c3ca16a9902b10e44fc80628ed72745/proc_limits.go#L121

I'm working on this.

PR #314, opened 2020-01-23
fix: process_max_fds is process limit, not OS

sam-github:
Fix https://github.com/siimon/prom-client/issues/311

In case anyone is wondering why I don't check for a exact match against the ulimit output, instead of just for "reasonableness", its because there isn't really any way to check the ulimit for the current process other than reading `/proc/self/limits`. Shelling out to ulimit would gives the correct value, but for the child shell... but that ulimit will be the default for /bin/sh, which might not be the ulimit for the current jest process (and is not on my machine, my login /bin/zsh has a different ulimit than /bin/sh).

### DIFF
--- a/lib/metrics/processMaxFileDescriptors.js
+++ b/lib/metrics/processMaxFileDescriptors.js
@@ -33,6 +33,12 @@ module.exports = (registry, config = {}) => {
 			isSet = true;
 
 			fileDescriptorsGauge.set(Number(maxFds));
+
+			// Only for interal use by tests, so they know when the
+			// value has been read.
+			if (config.ready) {
+				config.ready();
+			}
 		});
 	};
 };
--- a/lib/metrics/processMaxFileDescriptors.js
+++ b/lib/metrics/processMaxFileDescriptors.js
@@ -34,8 +34,8 @@ module.exports = (registry, config = {}) => {
 
 			let maxFds;
 			lines.find(line => {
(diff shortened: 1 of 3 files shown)


sam-github:
Not sure what your CI coverage is, but I manually tested on Linux and Mac (which has no /proc).

siimon:
Thanks, looks good to me! Do you mind update the changelog too?

Thanks!

sam-github:
updated

lib/metrics/processMaxFileDescriptors.js
```diff
@@ -25,14 +25,33 @@ module.exports = (registry, config = {}) => {
...
 			if (err) {
 				return;
 			}
 
+			const lines = limits.split('\n');
+
+			let maxFds;
+			lines.find(line => {
+				const parts = line.split(/  +/);
+				if (parts[0] === 'Max open files') {
```
zbjornson: You could save most of the `split` operations if you instead do something like `if (line.startsWith('Max open files')) { line.split(/\s+/)... }`

sam-github: I pushed your suggestion. PTAL. Note that this is called only once during process lifetime.

zbjornson: Heh, good point. Thanks for making the change anyway :)

sam-github:
Is there anything left for me to do?

PR #315, opened 2020-01-24
chore: drop support for Node.js 8.x

sam-github:
Node.js 8.x is EOL. Since dropping support for a Node.js release line is
semver-major, might as well drop 8.x at the same time as 6.x.

cf. 32920d0e1d946197b

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "index.d.ts"
   ],
   "engines": {
-    "node": ">=8"
+    "node": ">=10"
   },
   "scripts": {
     "benchmarks": "node ./benchmarks/index.js",
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - 8
   - 10
   - 12
   - node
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,92 +1,92 @@
 {
-  "name": "prom-client",
```
zbjornson: Can you undo the indentation changes in this file please?

sam-github: not easily, its enforced by the prom-client tooling  :-).

Its changed automatically by the eslint config hooks, which husky force-injected. It took me a bit to figure out how to work around this, but I managed to do it by deleting the `.git/hooks/pre-commit` that was put there by husky during `npm install`.

zbjornson: Oh man, I'm sorry, I didn't realize that was the cause. You could apply the indentation change in another commit if you want to, otherwise that'll keep happening.

sam-github: no problem, I was puzzled, too! See #318

zbjornson:
Can you rebase this please?

sam-github:
Sorry, just saw the rebase request now. Thanks for landing.

PR #316, opened 2020-01-24
chore: add windows to CI platforms

sam-github:
This increases CI coverage to all the "big three" (Windows, OSX, Linux).
prom-client support for Windows and OS X is incomplete, but it shouldn't
break for users who have it installed on their apps in development on
those platforms.

-----

I don't have Windows, but lets see if CI passes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,5 @@ node_js:
 os:
   - linux
   - osx
+  - windows
 osx_image: xcode7.3


sam-github:
Hm, windows wasn't happy. I'll wait for 8.x support to be removed then take another look at this.

zbjornson:
We have AppVeyor for Windows currently, so I'm not sure if we need to use Travis for it. (Travis tends to be slower than AppVeyor also.) Or does this change CI in another way?

sam-github:
Didn't realize appveyor was Windows.

SimenB:
fwiw, I'm in favour of moving to gh actions for ci, both since it'll allow us to have a single ci instead of multiple, and it's _way_ faster

PR #318, opened 2020-01-24
chore: reindent package.json

sam-github:
The husky pre-commit script enforces use of TAB in .json files, and
package.json is using SPC SPC. This isn't noticable until someone
changes one line in the package.json, at which point the entire file
gets reindented.

-----

I did this by adding a single TAB to the start of one line in package.json, then git commiting it, this triggered husky's git-commit hook to reformat the file, et voila, this is the result.


sam-github:
This is trivial, can it land?

zbjornson:
Oops, yes, thanks!

PR #320, opened 2020-01-28
chore: remove ignored package-lock.json

sam-github:
Its unused by consumers of prom-client, and not helpful to maintainers.

> package-lock.json ... cannot be published, and it will be ignored if
> found in any place other than the toplevel package.
- https://docs.npmjs.com/files/package-lock.json

See:
- https://github.com/siimon/prom-client/pull/306#issuecomment-579435110


zbjornson:
Hrm,

> npm ci
> npm ERR! cipm can only install packages with an existing package-lock.json or npm-shrinkwrap.json with lockfileVersion >= 1. Run an install with npm@5 or later to generate it, then try again.

IMO it's fine to change appveyor.yml to use `npm install` instead of `npm ci` for libraries. I'd rather our CI catch regressions from the latest version of dependencies that are allowed by our package.json.

PR #323, opened 2020-01-29
chore: add GH Actions CI

SimenB:
Sets up GH Actions. Can replace both travis and appveyor.

It's (way) faster and has better integration with GH itself, which might matter if we ever wanna do releases from CI or some such. It also automatically picks up typescript and eslint errors and adds inline annotations, which is nice 🙂

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   build:
-    name: Test on Node.js v${{ matrix.node-version }}
+    name: Test on Node.js v${{ matrix.node-version }} and OS ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,0 +1,23 @@
+name: Node.js CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: Test on Node.js v${{ matrix.node-version }}
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [10.x, 12.x, 13.x]
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -67,6 +67,7 @@
 		]
 	},
 	"prettier": {
+		"endOfLine": "auto",
```
SimenB: so it passes on windows. By default prettier reads from `.editorconfig`. We could add a `.gitattributes` file as well

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,23 @@
+name: Node.js CI
+
+on: [push, pull_request]
```
SimenB: could do on push to master, and all PRs, so we don't get what happened here - both for PR and for push

package.json
SimenB: alternatively, we could have separate jobs for tests (all OSes and node versions) and lint/typescript (LTS on ubuntu)

zbjornson: I'm a bit confused. Is prettier using CRLF on Windows, despite `.editorconfig` currently containing `end_of_line = lf`? Or where are the CRLFs coming from? (My pref is to use LF on all platforms.)

SimenB: GitHub by default checks out with crlf. We can add a gitattributes file as well

SimenB: See https://github.community/t5/GitHub-Actions/git-config-core-autocrlf-should-default-to-false/td-p/30445

zbjornson: I'm +1 on using `.gitattributes`. autocrlf is evil :)

sam-github:
Haven't played with gitactions enough to comment on the config, but moving towards one CI system rather than 2 sounds like a win.

Maybe you want to add the build status badge, too?

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository

SimenB:
Speed wise: 
image

siimon:
Great work and a good suggestion 👍

PR #322, opened 2020-01-29
chore: update outdated dependencies

SimenB:
- updates all dependencies to `latest`
  - most are just dropping old nodes
  - `lint-staged` warns on `git add` since it does it automatically, so I just removed that part
  - `eslint` has new recommended rules, fixed the `no-prototyupe-builtins` error
  - prettier wanted some new spacing
- added trailing commas all over
- updated ecmascript version in eslint in order to support ^

All are in individual commits to ease review.

PR #328, opened 2020-01-31
chore: use trailing commas

SimenB:
Extracted from #322.

This reduces diffs and makes them easier to read.

As an aside, prettier will default to `es5` setting for this in the upcoming v2


siimon:
Conflicts 💩

SimenB:
Yeah, I'll rebase this after all the PRs that are currently being worked on lands. It's all autofix, so it's a trivial patch to reapply 👍

PR #329, opened 2020-02-01
fix: collect metrics on scrape, not timeout

sam-github:
Only the event loop "lag" is still async, see the in-src notes.
    
Fixes: https://github.com/siimon/prom-client/issues/180

---

This is a WIP, it lacks tests (though it passes the existing ones).

Open questions:
- should I rearrange the commits? There are some unrelated ones in the PR, I could pull them into a seperate PR set if wanted. I figured out a way to do this that doesn't change the lib/metrics interface for the defaults, for example, so the metrics changes are mostly cleanup and could be done with or without the removal of the timeout.
- are the lib/metrics/ used directly by anyone? I'd like to remove the do-nothing closures for many of them that are called on every collect but are a no-op.
- do we need to support reconfiguration of the default metrics on the ***same*** registry? I can do it, its just a bit more state that needs to be removed now, and some notes in source suggested it wasn't wanted. Then again, not doing it might break some users. Opinions?

Also, my typescript is close to non-existent, pay close attention to my attempt to update the typescript definitions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Dropped support for end-of-life Node.js versions 6.x and 8.x
 - Dropped the previously deprecated support for positional parameters in
   constructors, only the config object forms remain.
+- Default metrics are collected on scrape of metrics endpoint, not on an
+  interval. The `timeout` option to `collectDefaultMetrics(conf)` is no longer
+  supported or needed, and the function no longer returns a `Timeout` object.
 
 ### Changed
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -590,7 +590,6 @@ export function exponentialBuckets(
 ): number[];
 
 export interface DefaultMetricsCollectorConfiguration {
-	timeout?: number;
 	timestamps?: boolean;
 	register?: Registry;
 	prefix?: string;
@@ -601,11 +600,10 @@ export interface DefaultMetricsCollectorConfiguration {
 /**
  * Configure default metrics
  * @param config Configuration object for default metrics collector
- * @return The setInterval number
  */
 export function collectDefaultMetrics(
 	config?: DefaultMetricsCollectorConfiguration
-): ReturnType<typeof setInterval>;
+): void;
 
 export interface defaultMetrics {
 	/**
--- a/lib/defaultMetrics.js
+++ b/lib/defaultMetrics.js
@@ -33,12 +33,7 @@ const metrics = {
 };
 const metricsList = Object.keys(metrics);
 
-let existingInterval = null;
-// This is used to ensure the program throws on duplicate metrics during first run
-// We might want to consider not supporting running the default metrics function more than once
-let init = true;
-
-module.exports = function startDefaultMetrics(config) {
+module.exports = function collectDefaultMetrics(config) {
 	if (config !== null && config !== undefined && !isObject(config)) {
 		throw new Error('config must be null, undefined, or an object');
 	}
@@ -52,33 +47,39 @@ module.exports = function startDefaultMetrics(config) {
 		config
 	);
 
-	if (existingInterval !== null) {
-		clearInterval(existingInterval);
+	const registry = config.register || globalRegistry;
+	const last = registry
+		.collectors()
+		.find(collector => collector._source === metrics);
+
+	if (last) {
+		throw new Error(
+			'Cannot add the default metrics twice to the same registry'
+		);
 	}
 
-	const initialisedMetrics = metricsList.map(metric => {
-		const defaultMetric = metrics[metric];
-		if (!init) {
-			defaultMetric.metricNames.map(
-				globalRegistry.removeSingleMetric,
-				globalRegistry
-			);
-		}
-
-		return defaultMetric(config.register, config);
+	const scrapers = metricsList.map(key => {
+		const metric = metrics[key];
+		return metric(config.register, config);
 	});
 
-	function updateAllMetrics() {
-		initialisedMetrics.forEach(metric => metric.call());
+	// Ideally the library would be based around a concept of collectors and
+	// async callbacks, but in the short-term, trigger scraping of the
+	// current metric value synchronously.
+	// - // https://prometheus.io/docs/instrumenting/writing_clientlibs/#overall-structure
+	function defaultMetricCollector() {
+		scrapers.forEach(scraper => scraper());
 	}
 
-	updateAllMetrics();
+	// defaultMetricCollector has to be dynamic, because the scrapers are in
+	// its closure, but we still want to identify a default collector, so
+	// tag it with a value known only to this module (the const metric array
+	// value) so we can find it later.
+	defaultMetricCollector._source = metrics;
+	registry.registerCollector(defaultMetricCollector);
 
-	existingInterval = setInterval(updateAllMetrics, config.timeout).unref();
-
-	init = false;
-
-	return existingInterval;
+	// Because the tests expect an immediate collection.
+	defaultMetricCollector();
 };
 
 module.exports.metricsList = metricsList;
--- a/lib/metrics/eventLoopLag.js
+++ b/lib/metrics/eventLoopLag.js
@@ -1,22 +1,5 @@
 'use strict';
 
-// Notes:
-//
-// Because legacy lag_seconds is collected async, the value will always lag by
-// one scrape interval.
-//
-// Reconcile with:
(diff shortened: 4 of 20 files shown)


index.d.ts
```diff
@@ -663,8 +663,8 @@ export function collectDefaultMetrics(
 
 /**
  * Configure default metrics
- * @param timeout The interval how often the default metrics should be probed
- * @deprecated A number to defaultMetrics is deprecated, please use \`collectDefaultMetrics({ timeout: ${timeout} })\`.
+ * @param timeout Deprecated, ignored if present.
+ * @deprecated Deprecated, ignored if present.
```
SimenB: can we just remove the entire argument? next release is a major anyways

SimenB: we can do so (plus other deprecated stuff) separately, perhaps?

sam-github:
No hurry, enjoy your weekend. I'm willing to remove code for sure, I'll get rid of the timeout relics.

SimenB:
>are the lib/metrics/ used directly by anyone?

it's not supported, so I'm not afraid of breaking people's usage if it. And since next release is a major, I think it's fine.

>do we need to support reconfiguration of the default metrics on the same registry?

I don't think so - what's the use case for it?

lib/metrics/processHandles.js
```diff
@@ -28,19 +28,11 @@ module.exports = (registry, config = {}) => {
...
-		: () => {
-				const handles = process._getActiveHandles();
-				updateMetrics(gauge, aggregateByObjectName(handles), false);
-				totalGauge.set(handles.length);
-		  };
-
-	return updater;
+	return () => {
+		const handles = process._getActiveHandles();
+		updateMetrics(gauge, aggregateByObjectName(handles), false);
```
SimenB: do we need this `updateMetrics` helper now?

index.d.ts
sam-github: how about do it now, it makes any subsequent changes easier if the deprecated code isn't there: https://github.com/siimon/prom-client/pull/330

sam-github:
>> do we need to support reconfiguration of the default metrics on the same registry?
> I don't think so - what's the use case for it?

No idea, but existing code supports it:
- https://github.com/siimon/prom-client/blob/9b4ef105cbc914e0ca50d9fbca9113deeb3db663/lib/defaultMetrics.js#L36
- https://github.com/siimon/prom-client/blob/9b4ef105cbc914e0ca50d9fbca9113deeb3db663/lib/defaultMetrics.js#L61-L66

I can reproduce the master behaviour in this PR, but if the comment is correct, then its OK to just delete the support. Thought I'd ask, though, rather than try to sneak it into a larger change. Full disclosure and all. :-)

index.d.ts
SimenB: Sure, whatever's easiest!

SimenB:
It seems I added that comment 3 years ago - I have no recollection of it 😅 

But no, I think it makes sense to throw if called again. IIRC it should do so if we just remove the guard, as that's the default registry behavior

index.d.ts
siimon: Hell yeah! I'm all for removing all deprecated stuffs.. been there for so long so it should be ok to remove those things now

siimon:
> > are the lib/metrics/ used directly by anyone?
> 
> it's not supported, so I'm not afraid of breaking people's usage if it. And since next release is a major, I think it's fine.
> 
> > do we need to support reconfiguration of the default metrics on the same registry?
> 
> I don't think so - what's the use case for it?

I agree, in my world the public interface to this library is what is in /index.js

siimon:
Regarding the reconfiguration I don't either see the use case for it. I think it could be removed 

> It seems I added that comment 3 years ago - I have no recollection of it 😅
> 
> But no, I think it makes sense to throw if called again. IIRC it should do so if we just remove the guard, as that's the default registry behavior

I agree here aswell. The only potential problem I could see is in tests where you run it multiple times. But that should probably be handled when you write the tests with a proper cleanup.

siimon:
I think it looks great, really appreciate your time working on this!

index.d.ts
sam-github: removed in #330 and I rebased this WIP on top of #330

lib/metrics/processHandles.js
sam-github: Yes, its shared code, used by lib/metrics/processRequests.js and lib/metrics/processHandles.js.

I think I can remove the second argument to aggregateByObjectName() though, its always passed `false`.

sam-github: done

sam-github:
> The only potential problem I could see is in tests where you run it multiple times. But that should probably be handled when you write the tests with a proper cleanup.

Or write the test to use a non-default registry, the throw that is in there will trigger if the default metrics are added multiple times _to the same registry_. They can be added to a number of different registries.

sam-github:
Added a test for https://github.com/siimon/prom-client/pull/329#issuecomment-583175871

sam-github:
I addressed the comments, I guess its ready for review again.

It does contain #330, maybe that should land first, then I'll rebase this. Or this could land and close #330. Whatever you all prefer.

siimon:
@sam-github the changes LGTM! Just got a conflict after the latest merge, so if you could fix that and update the PR and I'll merge right away!

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Breaking
 
 - Dropped support for end-of-life Node.js versions 6.x and 8.x
+- Dropped the previously deprecated support for positional parameters in
+  constructors, only the config object forms remain.
 
 ### Changed
```
SimenB: should mention the change in behaviour here

index.d.ts
```diff
@@ -662,14 +606,6 @@ export function collectDefaultMetrics(
 	config?: DefaultMetricsCollectorConfiguration
 ): ReturnType<typeof setInterval>;
```
SimenB: this return type is wrong now, right? should be `: void`

lib/defaultMetrics.js
```diff
@@ -31,57 +33,54 @@ const metrics = {
...
+	// value) so we can find it later.
+	defaultMetricCollector._source = metrics;
+	registry.registerCollector(defaultMetricCollector);
 
-	init = false;
+	// Because the tests expect an immediate collection.
+	defaultMetricCollector();
 
-	return existingInterval;
+	return 0; // Backwards compat, used to be a timeout in milliseconds.
```
SimenB: no need for this I think, it's landing in a major anyways. Just returning nothing seems fine

lib/defaultMetrics.js
```diff
@@ -31,57 +33,54 @@ const metrics = {
 };
 const metricsList = Object.keys(metrics);
 
-let existingInterval = null;
-// This is used to ensure the program throws on duplicate metrics during first run
-// We might want to consider not supporting running the default metrics function more than once
-let init = true;
-
 module.exports = function startDefaultMetrics(config) {
```
SimenB:
```suggestion
module.exports = function collectDefaultMetrics(config) {
```

maybe? The name makes no real difference

lib/metrics/eventLoopLag.js
```diff
@@ -1,5 +1,22 @@
...
+// https://nodejs.org/api/perf_hooks.html#perf_hooks_perf_hooks_monitoreventloopdelay_options
+//
+// (perf_hooks exists only `>= v11.10.0`)
+//
+// - Delay would be a prom histogram, but perf_hooks doesn't support that yet.
+// - Current possibility is to report h.max as the gauge value, which is
+// somewhat compatible with the current code.
+// - Might be possible to report perf_hook's histogram as a summary, using its
+// percentile() function. Summaries are less than ideal with prometheus but at
+// least the data is available in 12.x+. See #309
```
SimenB:
```suggestion
// least the data is available in 12.x+. See https://github.com/siimon/prom-client/issues/309
```

Just easier for people if they come over this in `node_modules` and want to take a look.

👏 for detailed comment, btw!

lib/metrics/processOpenFileDescriptors.js
```diff
@@ -20,14 +20,12 @@ module.exports = (registry, config = {}) => {
 	});
 
 	return () => {
-		fs.readdir('/proc/self/fd', (err, list) => {
-			if (err) {
-				return;
-			}
-
-			// Minus 1, as this invocation created one
```
SimenB: can we keep this comment?

sam-github:
@SimenB Thanks for the careful review, I made all the suggested changes, PTAL.

PR #330, opened 2020-02-02
fix: remove deprecated non-config object support

sam-github:
Metric constructors and startDefaultMetrics() all had deprecated support
for positional arguments. That support is removed, only config objects
are now supported. Since there are no deprecations, the deprecation code
is removed as well.


test/gaugeTest.js
```diff
@@ -13,7 +13,7 @@ describe('gauge', () => {
 		});
 		describe('with a parameter for each variable', () => {
```
siimon: I think this whole `describe` can be removed, the tests are duplicated but with different arguments. See the `describe('with parameters as object', ...` (https://github.com/siimon/prom-client/blob/e9bad9cb57f8dcf396a95d4f1e7442f4f3d09f97/test/gaugeTest.js#L216)

It is the same for all the tests for the different metric types.

lib/counter.js
```diff
@@ -25,34 +24,24 @@ const {
...
-
-			config = {
-				name,
-				help,
-				labelNames: labels,
-				registers: [globalRegistry]
-			};
+	constructor(config) {
+		if (!isObject(config)) {
+			throw new Error('constructor expected a config object');
```
SimenB: `TypeError`?

lib/counter.js
```diff
@@ -25,34 +24,24 @@ const {
...
+		}
+		config = Object.assign(
```
SimenB: We can use object spread

sam-github: Object.assign is used all over the place ATM, could someone else do js dialect updates later?

SimenB: of course 👍

sam-github:
I deleted the duplicate tests. Then the snapshots started to fail, but I think its because there are now snapshots for tests that don't exist.

PTAL

sam-github:
> Then the snapshots started to fail,

Perhaps I should have said, so I followed jest's directions, and ran `npm run test-unit -- -`, so now the tests pass. See above :-)/.

PR #335, opened 2020-02-21
feat: improve Registry type by adding the collector functions

poolik:
`registerCollector()` and `collectors()` were added to the Registry, expose them in TS type.


siimon:
Update the changelog please, then it's ready for merge! :)

poolik:
@siimon thanks for the quick feedback, updated now.

PR #337, opened 2020-02-25
Update README to reflect more v12 changes

zbjornson:
Most significantly, adds a section about `registerCollector`.

Also some minor cleanup and condensing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- Updated readme to reflect more v12 changes.
+
 ### Added
 
 - feat: exposed `registry.registerCollector()` and `registry.collectors()` methods in TypeScript declaration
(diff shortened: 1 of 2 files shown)


sam-github:
I wasn't thinking that registerCollector() would be used outside of the defaults, its a somewhat unfortunate interface (a version that supports callbacks would be better, but since .metrics() doesn't take a callback, that's not an option ATM), but it is what it is, and if prom-client wants to expose it I guess its at least simple and should be easy to keep stable.

zbjornson:
> since .metrics() doesn't take a callback, that's not an option ATM

The functions passed in to `registerCollector` have to be synchronous right now though, right? So we could easily do something like this:

```js
const gauge = new client.Gauge({
  name: 'metric_name',
  help: 'metric_help',
  collect() {
    this.set(/* value or labels+value */);
  }
});
```

We could also support `collect` returning a Promise, in which case `metrics` would also have to return a Promise. `clusterMetrics()` is already async, so on the plus side, those interfaces would converge.

siimon:
Thanks for doing this! Looks great and nice to get some input for someone that speaks english natively! 

I don't see any reason for not exposing the `registerCollector`, guess it could be useful even with the sync limitation?

sam-github:
Yes, I guess maybe its useful. I'm not sure if people are adding it to the .ts and the README because they actually are using it (or want to), or just because its there.

> The functions passed in to registerCollector have to be synchronous right now though, right?

Yes, but that's not because I thought it was easier, or didn't have the time to make them async, or anything like that.

They are sync because they are called by `.metrics()`, and ***it is sync*** (so everything it calls has to be sync, too).

It looked to me that making .metrics async was a pretty major change that would be guaranteed to break every single user of prom-client (at least, I can't think of a reason to use prom-client and not call `.metrics()`, maybe I'm missing something), unlike the other changes, which hit marginal (I hope) features, or features that had been deprecated for a fair while.

Shudrum:
Hey cool this update whould have saved me some time to understand things ^^

BTW, the examples should be updated too. Do you mind to update it also, or I open a PR also?

Thank you!

zbjornson:
Hey @Shudrum , thanks for pointing those out. I'm going to open a PR today with the async changes I'm proposing above, so before changing the examples, let's see if that change moves forward. :)

Shudrum:
Perfect! Thank you @zbjornson!

mattolson:
I especially like that the inaccuracy around GC stats is updated. Let's get this merged!

zbjornson:
Moot with #361

PR #235, opened 2018-12-28
Performance improvement

SerayaEryn:
This improves the performance of `hashObject()` and `getMetricAsPrometheusString()`.

I also added a benchmark for the metric type Summary.

Here is the benchmark result:

```bash
### Summary:

⚠ registry ➭ getMetricsAsJSON#1 with 64 is 0.5266% acceptably slower.
⚠ registry ➭ getMetricsAsJSON#2 with 8 is 1.965% acceptably slower.
⚠ registry ➭ getMetricsAsJSON#2 with 4 and 2 with 2 is 0.4088% acceptably slower.
⚠ registry ➭ getMetricsAsJSON#2 with 2 and 2 with 4 is 1.382% acceptably slower.
✓ registry ➭ getMetricsAsJSON#6 with 2 is 2.998% faster.
✓ registry ➭ metrics#1 with 64 is 10.60% faster.
✓ registry ➭ metrics#2 with 8 is 13.43% faster.
✓ registry ➭ metrics#2 with 4 and 2 with 2 is 8.662% faster.
✓ registry ➭ metrics#2 with 2 and 2 with 4 is 8.565% faster.
✓ registry ➭ metrics#6 with 2 is 4.454% faster.
✓ histogram ➭ observe#1 with 64 is 10.07% faster.
✓ histogram ➭ observe#2 with 8 is 27.06% faster.
✓ histogram ➭ observe#2 with 4 and 2 with 2 is 24.47% faster.
✓ histogram ➭ observe#2 with 2 and 2 with 4 is 22.92% faster.
✓ histogram ➭ observe#6 with 2 is 18.33% faster.
✓ summary ➭ observe#1 with 64 is 2.374% faster.
✓ summary ➭ observe#2 with 8 is 22.51% faster.
✓ summary ➭ observe#2 with 4 and 2 with 2 is 27.86% faster.
✓ summary ➭ observe#2 with 2 and 2 with 4 is 25.46% faster.
✓ summary ➭ observe#6 with 2 is 17.03% faster.
```

PR #340, opened 2020-02-27
Minor refactor

zekth:
This refactor a little sketchy pattern.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- chore: improve performance of `registry.getMetricAsPrometheusString`
+
 ### Added
 
 - feat: exposed `registry.registerCollector()` and `registry.collectors()` methods in TypeScript declaration
@@ -43,7 +45,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - feat: implement GC metrics collection without native(C++) modules.
-- faet: implement advanced event loop monitoring
+- feat: implement advanced event loop monitoring
 
 ## [11.5.3] - 2019-06-27
 
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -50,12 +50,10 @@ class Registry {
 			if (size > 0) {
 				let labels = '';
 				let i = 0;
-				for (; i < size; i++) {
-					labels += `${keys[i]}="${escapeLabelValue(val.labels[keys[i]])}"`;
-					if (i !== size - 1) {
-						labels += ',';
-					}
+				for (; i < size - 1; i++) {
+					labels += `${keys[i]}="${escapeLabelValue(val.labels[keys[i]])}",`;
 				}
+				labels += `${keys[i]}="${escapeLabelValue(val.labels[keys[i]])}"`;
 				metricName += `{${labels}}`;
 			}
 
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -43,14 +43,20 @@ class Registry {
 				}
 			}
 
-			let labels = '';
-			for (const key of Object.keys(val.labels)) {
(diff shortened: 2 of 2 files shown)


lib/registry.js
```diff
@@ -43,14 +43,14 @@ class Registry {
...
+			const labels = [];
 			for (const key of Object.keys(val.labels)) {
-				labels += `${key}="${escapeLabelValue(val.labels[key])}",`;
+				labels.push(`${key}="${escapeLabelValue(val.labels[key])}"`);
 			}
 
 			let metricName = val.metricName || item.name;
 			if (labels) {
-				metricName += `{${labels.substring(0, labels.length - 1)}}`;
+				metricName += `{${labels.join(',')}}`;
```
zbjornson: Thanks for the PRs!

Unfortunately this approach is slower:

```
✗ registry ➭ metrics#1 with 64 is 11.13% slower.
⚠ registry ➭ metrics#2 with 8 is 7.170% acceptably slower.
⚠ registry ➭ metrics#2 with 4 and 2 with 2 is 0.4725% acceptably slower.
⚠ registry ➭ metrics#2 with 2 and 2 with 4 is 8.324% acceptably slower.
⚠ registry ➭ metrics#6 with 2 is 6.719% acceptably slower.
```

If you want to change it, could you try out the approach outlined in https://github.com/siimon/prom-client/pull/235#discussion_r244376041? You can run `yarn/npm benchmarks` while on your branch to compare against the current release.

zekth:
Update:
```
✓ registry ➭ getMetricsAsJSON#1 with 64 is 1.652% faster.
✓ registry ➭ getMetricsAsJSON#2 with 8 is 0.8094% faster.
✓ registry ➭ getMetricsAsJSON#2 with 4 and 2 with 2 is 4.417% faster.
⚠ registry ➭ getMetricsAsJSON#2 with 2 and 2 with 4 is 2.478% acceptably slower.
⚠ registry ➭ getMetricsAsJSON#6 with 2 is 1.328% acceptably slower.
✓ registry ➭ metrics#1 with 64 is 13.89% faster.
✓ registry ➭ metrics#2 with 8 is 8.079% faster.
✓ registry ➭ metrics#2 with 4 and 2 with 2 is 7.516% faster.
✓ registry ➭ metrics#2 with 2 and 2 with 4 is 9.523% faster.
⚠ registry ➭ metrics#6 with 2 is 0.7138% acceptably slower.
```

siimon:
I don't really see the point, sorry.. Could you give me some insight on why you think this refactor is needed?

zekth:
> I don't really see the point, sorry.. Could you give me some insight on why you think this refactor is needed?

At first it was only to tackle the `substring` pattern, but as it was slower just changed the implementation like suggested and improved the perf.

lib/registry.js
```diff
@@ -43,14 +43,20 @@ class Registry {
...
+			const keys = Object.keys(val.labels);
+			const size = keys.length;
+			if (size > 0) {
+				let labels = '';
+				let i = 0;
+				for (; i < size; i++) {
+					labels += `${keys[i]}="${escapeLabelValue(val.labels[keys[i]])}"`;
+					if (i !== size - 1) {
+						labels += ',';
+					}
```
zbjornson: Almost. The exact way I like to do this is to loop up to `size - 1` and not have an `if` in the loop body (see comment I linked to previously). I think that will save a line also.

siimon:
Alright, could you update the changelog please?

CHANGELOG.md
```diff
@@ -45,7 +46,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - feat: implement GC metrics collection without native(C++) modules.
-- faet: implement advanced event loop monitoring
+- feat: implement advanced event loop monitoring
```
zekth: Fixed this typo also

PR #341, opened 2020-02-27
[CI] Add cache handling

zekth:
Add cache handling to the CI and also add package-lock.

Will improve speed by not polling from `npm` at each run

With cache : 1m15s
Without cache : 2m9s


zbjornson:
Can caching be enabled without the lock file? We actually don't want that checked in to git.

zekth:
> Can caching be enabled without the lock file? We actually don't want that checked in to git.

It can. Will update so.

siimon:
Awesome! Thanks for the PR

PR #343, opened 2020-03-01
fix: remove useless using timestamp param after #333

amel-true:
I found some useless (after #333) using of third parameter `timestamp` in `Gauge.prototype.set` and `Counter.prototype.inc` functions and removed it.


siimon:
👍

PR #354, opened 2020-03-24
docs: examples with recorded output

guyellis:
Not sure if this is of value? I find examples like this useful to run from the command line to get an idea of how something functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,22 @@ const collectDefaultMetrics = client.collectDefaultMetrics;
 collectDefaultMetrics();
 ```
 
+### Custom Metrics
+
+All metric types have two mandatory parameters: `name` and `help`. Refer to
+<https://prometheus.io/docs/practices/naming/> for guidance on naming metrics.
+
+For metrics based on point-in-time observations (e.g. current memory usage, as
+opposed to HTTP request durations observed continuously in a histogram), you
+should provide a `collect()` function, which will be invoked when Prometheus
+scrapes your metrics endpoint. `collect()` can either be synchronous or return a
+promise. See **Gauge** below for an example. (Note that you should not update
+metric values in a `setInterval` callback; do so in this `collect` function
+instead.)
+
+See [**Labels**](#labels) for information on how to configure labels for all
+metric types.
+
 #### Counter
 
 Counters go up, and reset when the process restarts.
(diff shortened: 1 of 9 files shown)


example/counter.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
+
+// Counter
+// Single Label
+// Multiple Values
+
+const register = require('..').register;
+
+const Counter = require('..').Counter;
```
SimenB, line 9:
```suggestion
const { Counter, register } = require('..');
```

example/default-metrics.js
```diff
@@ -0,0 +1,149 @@
+'use strict';
+
+const register = require('..').register;
+
+require('..').collectDefaultMetrics({
```
SimenB, line 5:
```suggestion
const { collectDefaultMetrics, register } = require('..');

collectDefaultMetrics({
```

example/gauge.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
+
+// Gauge
+// Single Label
+// Multiple Values
+
+const register = require('..').register;
+
+const Gauge = require('../').Gauge;
```
SimenB, line 9:
```suggestion
const { Gauge, register } = require('..');
```

SimenB: etc 😅

SimenB (approved):
I like it. @siimon @zbjornson any reason not to include these?

guyellis:
@SimenB @siimon @zbjornson - I've rebased this PR and I think that I've fixed all those requires as per @SimenB comments.

README.md
```diff
@@ -104,22 +104,6 @@ const collectDefaultMetrics = client.collectDefaultMetrics;
...
-opposed to HTTP request durations observed continuously in a histogram), you
-should provide a `collect()` function, which will be invoked when Prometheus
-scrapes your metrics endpoint. `collect()` can either be synchronous or return a
-promise. See **Gauge** below for an example. (Note that you should not update
-metric values in a `setInterval` callback; do so in this `collect` function
-instead.)
-
-See [**Labels**](#labels) for information on how to configure labels for all
-metric types.
-
```
zbjornson, line 122: Oops, this should stay here.

zbjornson (changes requested):
> any reason not to include these?

Only reason might be that these could become stale when new metrics are added or changed, but otherwise good with me!

README.md
guyellis: No idea why I did this - reverted! Good catch.

zbjornson (approved):
🍻 great, thank you!

Issue #238, opened 2019-01-04
nodejs_version_info empty after registry reset


ssube:
I've been using this library in a small bot with great results, and recently implemented a feature where SIGINT and SIGHUP will reload the config file and reset any stateful counters (for development). I'm using `collectDefaultMetrics({ registry })` to collect general Node metrics and call `registry.resetMetrics` upon receiving the signals.

After calling `resetMetrics` and waiting for the next interval tick, most metrics re-appear, but not all of them:

```
# HELP nodejs_heap_space_size_available_bytes Process heap space size available from node.js in bytes.
# TYPE nodejs_heap_space_size_available_bytes gauge
nodejs_heap_space_size_available_bytes{space="new"} 658168 1546562858531
nodejs_heap_space_size_available_bytes{space="old"} 1583448 1546562858531
nodejs_heap_space_size_available_bytes{space="code"} 456544 1546562858531
nodejs_heap_space_size_available_bytes{space="map"} 514160 1546562858531
nodejs_heap_space_size_available_bytes{space="large_object"} 1397517824 1546562858531

# HELP nodejs_version_info Node.js version info.
# TYPE nodejs_version_info gauge
```

It would make sense that `nodejs_version_info` is only collected once, as it is very unlikely to change, but it seems to get lost after the registry is reset.

What's the best way to make sure that metric is always populated? It would be pretty easy for me to clear the interval and call `collectDefaultMetrics` again, if that makes sense.

**Edit:** reading further, it looks like this may be a non-issue if #180 is implemented as discussed, since my code would call that each time rather than set and reset an interval. Would that be the most correct solution?

siimon:
Yeah I would say that fixing it as suggested in #180 would be the way to fix this problem as well!

sam-github:
@ssube was this fixed by #180?

ssube:
Not quite - after resetting the registry, the `nodejs_version_info` gauge remains empty forever.

I had to make a few other changes for v12.0 (https://github.com/ssube/isolex/pull/837), but have gotten it working with a hard-coded (for the moment) timeout of 10 (https://github.com/ssube/isolex/pull/837/files#diff-5a28ddf4a6b3741f1800b6fcc801c9b1R42):

```
    this.collector({
      register: this.metrics,
      timeout: 10,
    });
```

Metrics:

```
# HELP nodejs_heap_space_size_available_bytes Process heap space size available from Node.js in bytes.
[...]

PR #361, opened 2020-03-29
change: add "collect" fn, make scrape fns async, ...

zbjornson:
Finally, this is the semver-major change to the interface I'm proposing per discussion in #337. What do you guys think? @sam-github I can't seem to request a review from you, but would love your thoughts.

(This is also the first non-"mechanical" entry in the changelog, FWIW.)

---

* Adds a `collect` fn, which may be async, to each metric type. This provides a cleaner interface than registering collectors, and allows the collector to do asynchronous work. (Previously it would be one scrape behind if it had to do async work.)

* In turn, makes `registry.metrics()` and `registry.get...` async, and removes the `collectors`/`registerCollector` functions.

* Fixes `process_max_fds` and `process_start_time_seconds` so that they have values after the registry is reset (same vein as #238)

* Cleans up the readme a bit (removes some out-of-date info)

---

I don't suspect this will be a problem, but this does increase the number of Promises created on each scrape. I could decrease that by one per metric fairly easily.

I also forgot to update the benchmark code. Will do if y'all are +1s on this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,13 @@
 # Prometheus client for node.js [![Build Status](https://travis-ci.org/siimon/prom-client.svg?branch=master)](https://travis-ci.org/siimon/prom-client) [![Build status](https://ci.appveyor.com/api/projects/status/k2e0gwonkcee3lp9/branch/master?svg=true)](https://ci.appveyor.com/project/siimon/prom-client/branch/master) [![Actions Status](https://github.com/siimon/prom-client/workflows/Node.js%20CI/badge.svg?branch=master)](https://github.com/siimon/prom-client/actions)
 
-A prometheus client for node.js that supports histogram, summaries, gauges and
+A prometheus client for Node.js that supports histogram, summaries, gauges and
 counters.
 
 ## Usage
 
 See example folder for a sample usage. The library does not bundle any web
-framework, to expose the metrics just return the `metrics()` function in the
-registry.
+framework. To expose the metrics, respond to Prometheus's scrape requests with
+the result of `await registry.metrics()`.
 
 ### Usage with Node.js's `cluster` module
 
@@ -34,22 +34,17 @@ worker processes.
 
 ## API
 
-### Configuration
-
-All metric types has 2 mandatory parameters, name and help.
-
 ### Default metrics
 
 There are some default metrics recommended by Prometheus
 [itself](https://prometheus.io/docs/instrumenting/writing_clientlibs/#standard-and-runtime-collectors).
-To collect these, call `collectDefaultMetrics`
+To collect these, call `collectDefaultMetrics`. In addition, some
+Node.js-specific metrics are included, such as event loop lag, active handles,
+GC and Node.js version. See [lib/metrics](lib/metrics) for a list of all
+metrics.
 
 NOTE: Some of the metrics, concerning File Descriptors and Memory, are only
 available on Linux.
-
-In addition, some Node-specific metrics are included, such as event loop lag,
-active handles, GC and Node.js version. See what metrics there are in
-[lib/metrics](lib/metrics).
 
 `collectDefaultMetrics` optionally accepts a config object with following entries:
 
@@ -62,7 +57,6 @@ To register metrics to another registry, pass it in as `register`:
 
 ```js
 const client = require('prom-client');
-
 const collectDefaultMetrics = client.collectDefaultMetrics;
 const Registry = client.Registry;
 const register = new Registry();
@@ -73,9 +67,7 @@ To use custom buckets for GC duration histogram, pass it in as `gcDurationBucket
 
 ```js
 const client = require('prom-client');
-
 const collectDefaultMetrics = client.collectDefaultMetrics;
-
 collectDefaultMetrics({ gcDurationBuckets: [0.1, 0.2, 0.3] });
 ```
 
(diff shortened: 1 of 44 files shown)


README.md
```diff
@@ -102,7 +94,22 @@ const collectDefaultMetrics = client.collectDefaultMetrics;
...
-### Counter
+### Custom Metrics
+
+All metric types have two mandatory parameters: `name` and `help`. Refer to
+<https://prometheus.io/docs/practices/naming/> for guidance on naming metrics.
+
+For metrics based on point-in-time observations (e.g. current memory usage, as
+opposed to HTTP request durations observed continuously in a histogram), you
+should provide a `collect` function, which will be invoked when Prometheus
+scrapes your registry. `fn` can either be synchronous or return a promise. See
```
sam-github:
```suggestion
scrapes your registry. `collect()` can either be synchronous or return a promise. See
```

?

README.md
```diff
@@ -102,7 +94,22 @@ const collectDefaultMetrics = client.collectDefaultMetrics;
...
 
-### Counter
+### Custom Metrics
+
+All metric types have two mandatory parameters: `name` and `help`. Refer to
+<https://prometheus.io/docs/practices/naming/> for guidance on naming metrics.
+
+For metrics based on point-in-time observations (e.g. current memory usage, as
+opposed to HTTP request durations observed continuously in a histogram), you
+should provide a `collect` function, which will be invoked when Prometheus
```
sam-github:
```suggestion
should provide a `collect()` function, which will be invoked when Prometheus
```

README.md
sam-github:
```suggestion
scrapes your metrics endpoint. `fn` can either be synchronous or return a promise. See
```

README.md
```diff
@@ -112,66 +119,91 @@ const counter = new client.Counter({
...
 const client = require('prom-client');
-new client.Histogram({
+new client.Gauge({
   name: 'metric_name',
   help: 'metric_help',
-  buckets: [0.1, 5, 15, 50, 100, 500],
+  async collect() {
+    // Invoked when the registry collects its metrics' values.
+    await somethingAsync();
+    this.set(/* the current value */);
```
sam-github:
```suggestion
    const currentValue = await somethingAsync();
    this.set(currentValue);
```

README.md
```diff
@@ -112,66 +119,91 @@ const counter = new client.Counter({
...
+
+##### Utility Functions
+
+```js
+// Set value to current time:
+gauge.setToCurrentTime();
+
+// Record durations:
+const end = gauge.startTimer();
+xhrRequest(function (err, res) {
```
sam-github: Is xhrRequest a thing on the server side?

README.md
```diff
@@ -346,19 +380,19 @@ AggregatorRegistry.setRegistries([registry1, registry2]);
 
 ### Register
 
-You can get all metrics by running `register.metrics()`, which will output a
-string for prometheus to consume.
+You can get all metrics by running `await register.metrics()`, which will return
+a string for Prometheus to consume.
```
sam-github:
```suggestion
a string in the Prometheus exposition format.
```

README.md
```diff
@@ -442,7 +473,9 @@ new client.Histogram({
 The content-type prometheus expects is also exported as a constant, both on the
 `register` and from the main file of this project, called `contentType`.
 
-## Garbage Collection
+### Garbage Collection Metrics
 
-To avoid dependencies in this module, GC stats are kept outside of it. If you
-want GC stats, you can use https://github.com/SimenB/node-prometheus-gc-stats
+To avoid native dependencies in this module, GC statistics for bytes reclaimed
```
sam-github: Isn't the data provided by that module now provided using the builtin APIs?

zbjornson: Everything except "bytes reclaimed" is now via builtin APIs. You might actually know the answer to this, or know who to ask... `gc-stats` just calls `GetHeapStatistics` in the GC prologue and epilogue and diffs them to calculate "bytes reclaimed". But since v8's GC is concurrent, I don't think that's an accurate method. Do you know if that's true? If so, we should remove this section of the readme.

---

Thanks for the review, agree with all the other changes.

sam-github: It might be accurate, but I'm not sure its useful, I suspect it might have been emitted just because it seemed possible. If many bytes are reclaimed, is that good? If only a few are reclaimed, is that good? The total usage and how it varies with time strike me as the useful metrics, and also allow diff scrape-to-scrape.

muety:
Great changes! Hope they're gonna be merged soon :)

zbjornson:
That was a much longer hiatus that intended, sorry.

Made the suggested changes and updated the benchmarks. @siimon and @SimenB ?

benchmarks/registry.js
```diff
@@ -16,15 +16,19 @@ function setupRegistrySuite(suite) {
...
 		);
 	});
 
 	labelSetups.forEach(({ name, counts }) => {
-		suite.add(`metrics#${name}`, (client, registry) => registry.metrics(), {
-			setup: setup(counts),
-		});
+		suite.add(
+			`metrics#${name}`,
+			async (client, registry) => await registry.metrics(),
```
SimenB, line 27: this doesn't need to change does it? Just returning the promise seems like it'd work?

example/cluster.js
```diff
@@ -11,12 +11,10 @@ if (cluster.isMaster) {
...
 	}
 
-	metricsServer.get('/cluster_metrics', (req, res) => {
-		aggregatorRegistry.clusterMetrics((err, metrics) => {
-			if (err) console.log(err);
-			res.set('Content-Type', aggregatorRegistry.contentType);
-			res.send(metrics);
-		});
+	metricsServer.get('/cluster_metrics', async (req, res) => {
+		const metrics = await aggregatorRegistry.clusterMetrics();
```
SimenB, line 15: should we try-catch? Express doesn't handle promises (async function) too well

zbjornson:
Changes made and rebased on master.

zbjornson:
@siimon unless I hear otherwise from you, I'll merge this in a few days.

lib/gauge.js
```diff
@@ -81,7 +81,11 @@ class Gauge extends Metric {
 		return startTimer.call(this, labels)();
 	}
 
-	get() {
+	async get() {
+		if (this.collect) {
+			const v = this.collect();
+			if (v instanceof Promise) await v;
```
ankon, line 87: I think it would be cleaner to simply ensure that you can `await` things:
```js
await Promise.resolve(v)
```

Then again, NodeJS does that internally anyways, see:
```js
function foo() {
	return 2;
}

async function bar() {
	const x = foo();
	const y = await x;
	console.log(x, y); // Logs '2 2'
}

bar()
```

So the cleanest here (and below) would be to just do `await v`, no?

zbjornson: The `if` avoids unnecessary ticks of the event loop by only awaiting the `collect` function if it's asynchronous. (Quite a few ticks saved even with the default metrics.)

PR #365, opened 2020-04-10
Expose global registry

jorge07:
## Problem

Global registry not accessible and having `already registered` errors on test suites.

## Why 

Helps when running test suites. It's a good way to clean the registry on each iteration.

## Current workaround

```ts
((prom.Registry as any).globalRegistry as prom.Registry).clear();
```

Thanks

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -81,7 +81,13 @@ export class Registry {
 	 * @param registers The registers you want to merge together
 	 */
 	static merge(registers: Registry[]): Registry;
+	
+	/**
+	 * Global registry
+	 */
+	globalRegistry: Registry;
 }
+
 export type Collector = () => void;
 
 /**
--- a/index.d.ts
+++ b/index.d.ts
@@ -85,7 +85,7 @@ export class Registry {
 	/**
 	 * Global registry
 	 */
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -81,7 +81,13 @@ export class Registry {
 	 * @param registers The registers you want to merge together
 	 */
 	static merge(registers: Registry[]): Registry;
+	
+	/**
+	 * Global registry
+	 */
+	globalRegistry: Registry;
```
zbjornson, line 88: The global registry is exported (from the JS and .d.ts files) at the top level as `register`. I'm fine adding this, but i don't know if the original intent was for `Register.globalRegistry` to be private.

If this is added, it should be `static globalRegistry: Registry` though.

jorge07: Applied.

It is a better way to do this https://github.com/jorge07/billing-api/blob/master/tests/ui/http/transaction/get.test.ts#L18?
I basically want to clean the registry before each test so wherever is needed I can probe metrics measure that I need without having a conflict registering a metric that already exists

jorge07: `prom.register.clear();` seems the way to go. Closing as no issue here. Thanks!

PR #372, opened 2020-05-11
chore: upgrade to Jest 26

SimenB:
Also add Node 14 to CI.

~Lolex has been renamed to `@sinonjs/fake-timers`. Jest bundles it now, but this keeps the diff down~ meh, second commit uses it. Note that I fake system time to be `0` (epoch). This is the default when using `lolex`/`@sinonjs/fake-timers`, while Jest default to the current system real time. _Not_ setting this gives a bunch of failing tests (e.g. we expect seconds to match ms from `Date.now()` etc). That's a bug in our tests, but I don't think it should be fixed in this PR

### DIFF
--- a/test/gaugeTest.js
+++ b/test/gaugeTest.js
@@ -4,7 +4,6 @@ describe('gauge', () => {
 	const Gauge = require('../index').Gauge;
 	const Registry = require('../index').Registry;
 	const globalRegistry = require('../index').register;
-	const lolex = require('@sinonjs/fake-timers');
 	let instance;
 
 	describe('global registry', () => {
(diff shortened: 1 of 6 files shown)


test/gaugeTest.js
```diff
@@ -43,19 +42,21 @@ describe('gauge', () => {
 			});
 
 			it('should start a timer and set a gauge to elapsed in seconds', () => {
-				const clock = lolex.install();
+				jest.useFakeTimers('modern');
```
SimenB, line 45: passing `modern` uses `@sinonjs/fake-timers` under the hood. It'll be the default from Jest 27

test/gaugeTest.js
```diff
@@ -43,19 +42,21 @@ describe('gauge', () => {
...
-				clock.uninstall();
+				jest.useRealTimers();
 			});
 
 			it('should set to current time', () => {
-				const clock = lolex.install();
+				jest.useFakeTimers('modern');
+				jest.setSystemTime(0);
 				instance.setToCurrentTime();
 				expectValue(Date.now());
```
SimenB, line 58: example of broken test. `Date.now() === 0` so value is also `0`. Setting it to an actual time gives test errors like this:

image

PR #276, opened 2019-06-18
Improve performance of registry.metrics()

SerayaEryn:
This improves the performance of `registry.metrics()` by approximately 40%.
This is achieved by creating a specific formatter for each metric.

Here are the benchmark results:

```
### Summary:

⚠ registry ➭ getMetricsAsJSON#1 with 64 is 2.343% acceptably slower.
✓ registry ➭ getMetricsAsJSON#2 with 8 is 0.8804% faster.
⚠ registry ➭ getMetricsAsJSON#2 with 4 and 2 with 2 is 1.183% acceptably slower.
✓ registry ➭ getMetricsAsJSON#2 with 2 and 2 with 4 is 0.2041% faster.
✓ registry ➭ getMetricsAsJSON#6 with 2 is 0.7355% faster.
✓ registry ➭ metrics#1 with 64 is 40.57% faster.
✓ registry ➭ metrics#2 with 8 is 38.32% faster.
✓ registry ➭ metrics#2 with 4 and 2 with 2 is 41.24% faster.
✓ registry ➭ metrics#2 with 2 and 2 with 4 is 43.03% faster.
✓ registry ➭ metrics#6 with 2 is 41.95% faster.
✓ histogram ➭ observe#1 with 64 is 4.848% faster.
✓ histogram ➭ observe#2 with 8 is 3.919% faster.
⚠ histogram ➭ observe#2 with 4 and 2 with 2 is 0.5036% acceptably slower.
✓ histogram ➭ observe#2 with 2 and 2 with 4 is 1.179% faster.
✓ histogram ➭ observe#6 with 2 is 11.81% faster.
⚠ summary ➭ observe#1 with 64 is 2.889% acceptably slower.
⚠ summary ➭ observe#2 with 8 is 2.747% acceptably slower.
✓ summary ➭ observe#2 with 4 and 2 with 2 is 2.210% faster.
✓ summary ➭ observe#2 with 2 and 2 with 4 is 0.1158% faster.
⚠ summary ➭ observe#6 with 2 is 3.125% acceptably slower.
```

PR #368, opened 2020-04-19
remove unnecessary necessity of predeclared, known-at-declaration-time labelset for metrics

dt-rush:
Please see discussion in #298 for context

+ improved validation testing

Change-type: minor
Connects-to: #298
Signed-off-by: dt-rush <nickp@balena.io>

PR #373, opened 2020-05-14
improve performance of registry.metrics()

doochik:
I've rebased original PR #276 and achieve the same results

```
### Summary:

✓ registry ➭ getMetricsAsJSON#1 with 64 is 1.900% faster.
✓ registry ➭ getMetricsAsJSON#2 with 8 is 2.054% faster.
⚠ registry ➭ getMetricsAsJSON#2 with 4 and 2 with 2 is 7.658% acceptably slower.
✓ registry ➭ getMetricsAsJSON#2 with 2 and 2 with 4 is 8.235% faster.
✓ registry ➭ getMetricsAsJSON#6 with 2 is 4.261% faster.
✓ registry ➭ metrics#1 with 64 is 6.524% faster.
✓ registry ➭ metrics#2 with 8 is 35.91% faster.
✓ registry ➭ metrics#2 with 4 and 2 with 2 is 26.44% faster.
✓ registry ➭ metrics#2 with 2 and 2 with 4 is 31.02% faster.
✓ registry ➭ metrics#6 with 2 is 34.78% faster.
✓ histogram ➭ observe#1 with 64 is 10.78% faster.
✓ histogram ➭ observe#2 with 8 is 9.954% faster.
✓ histogram ➭ observe#2 with 4 and 2 with 2 is 18.51% faster.
✓ histogram ➭ observe#2 with 2 and 2 with 4 is 5.529% faster.
✓ histogram ➭ observe#6 with 2 is 4.469% faster.
⚠ summary ➭ observe#1 with 64 is 0.7088% acceptably slower.
✓ summary ➭ observe#2 with 8 is 0.8064% faster.
⚠ summary ➭ observe#2 with 4 and 2 with 2 is 0.8744% acceptably slower.
⚠ summary ➭ observe#2 with 2 and 2 with 4 is 3.878% acceptably slower.
✓ summary ➭ observe#6 with 2 is 2.453% faster.
```

### DIFF
--- a/lib/counter.js
+++ b/lib/counter.js
@@ -9,11 +9,6 @@ const { validateLabel } = require('./validation');
 const { Metric } = require('./metric');
 
 class Counter extends Metric {
-	constructor(...args) {
-		super(...args);
-		this.type = 'counter';
-	}
-
 	/**
 	 * Increment counter
 	 * @param {object} labels - What label you want to be incremented
@@ -61,6 +56,8 @@ class Counter extends Metric {
 		return removeLabels.call(this, this.hashMap, labels);
 	}
 }
+
+Counter.prototype.type = 'counter';
 
 const reset = function () {
 	this.hashMap = {};
--- a/lib/counter.js
+++ b/lib/counter.js
@@ -4,12 +4,16 @@
 'use strict';
 
 const util = require('util');
-const type = 'counter';
 const { hashObject, isObject, getLabels, removeLabels } = require('./util');
 const { validateLabel } = require('./validation');
 const { Metric } = require('./metric');
 
 class Counter extends Metric {
+	constructor(...args) {
+		super(...args);
+		this.type = 'counter';
+	}
+
 	/**
 	 * Increment counter
 	 * @param {object} labels - What label you want to be incremented
@@ -37,7 +41,7 @@ class Counter extends Metric {
 		return {
 			help: this.help,
 			name: this.name,
-			type,
+			type: this.type,
 			values: Object.values(this.hashMap),
 			aggregator: this.aggregator,
 		};
--- a/lib/formatter.js
+++ b/lib/formatter.js
@@ -1,0 +1,72 @@
+'use strict';
+
+const generateFunction = require('generate-function');
+
+function getFormatter(metric, defaultLabels) {
+	const defaultLabelNames = Object.keys(defaultLabels);
+	const name = metric.name;
+	const help = escapeString(metric.help);
+	const type = metric.type;
+	const labelsCode = getLabelsCode(metric, defaultLabelNames, defaultLabels);
+
+	const gen = generateFunction();
+	gen(`
+  function format(item, escapeLabelValue, getValueAsString) {
+    const info = '# HELP ${name} ${help}\\n# TYPE ${name} ${type}\\n';
+    let values = '';
+    for (const val of item.values || []) {
+      val.labels = val.labels || {};
+      let metricName = val.metricName || '${name}';
+      const labels = \`${labelsCode}\`;
+      const hasLabels = labels.length > 0;
+      metricName += \`\${hasLabels ? '{' : ''}\${labels}\${hasLabels ? '}' : ''}\`;
+      let line = \`\${metricName} \${getValueAsString(val.value)}\`;
+      values += \`\${line}\\n\`;
+    }
+    return info + values;
+  }`);
+	return gen.toFunction();
+}
+
+function getLabelsCode(metric, defaultLabelNames, defaultLabels) {
+	let labelString = '';
+	const labelNames = getLabelNames(metric, defaultLabelNames);
+	for (let index = 0; index < labelNames.length; index++) {
+		const labelName = labelNames[index];
+		if (labelName === 'quantile') {
+			labelString += `\${val.labels.quantile != null ? \`quantile="\${escapeLabelValue(val.labels.quantile)}"\` : ''}`;
+		} else if (labelName === 'le') {
+			labelString += `\${val.labels.le != null ? \`le="\${escapeLabelValue(val.labels.le)}"\` : ''}`;
+		} else {
+			labelString += `${labelName}="\${val.labels['${labelName}'] ? escapeLabelValue(val.labels['${labelName}']) : escapeLabelValue('${defaultLabels[labelName]}')}"`;
+		}
+		if (index !== labelNames.length - 1 && labelString.length > 0) {
+			labelString += ',';
+		}
+	}
+	return labelString;
+}
+
+function getLabelNames(metric, defaultLabelNames) {
+	const labelNames = [...metric.labelNames];
+	for (const labelName of defaultLabelNames) {
+		if (!labelNames.includes(labelName)) {
+			labelNames.push(labelName);
+		}
+	}
+	if (metric.type === 'summary') {
+		labelNames.push('quantile');
+	}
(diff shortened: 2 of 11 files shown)


lib/formatter.js
```diff
@@ -0,0 +1,72 @@
...
+      let metricName = val.metricName || '${name}';
+      const labels = \`${labelsCode}\`;
+      const hasLabels = labels.length > 0;
+      metricName += \`\${hasLabels ? '{' : ''}\${labels}\${hasLabels ? '}' : ''}\`;
+      let line = \`\${metricName} \${getValueAsString(val.value)}\`;
+      values += \`\${line}\\n\`;
+    }
+    return info + values;
+  }`);
+	return gen.toFunction();
```
zbjornson, line 28: You could save a dependency and just use the Function constructor directly, since you're not making use of any of `generateFunctions`'s other abilities.

lib/formatter.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+
+const generateFunction = require('generate-function');
+
+function getFormatter(metric, defaultLabels) {
+	const defaultLabelNames = Object.keys(defaultLabels);
+	const name = metric.name;
+	const help = escapeString(metric.help);
+	const type = metric.type;
+	const labelsCode = getLabelsCode(metric, defaultLabelNames, defaultLabels);
```
zbjornson, line 10: ⭐ This is going to make it impossible to implement #298/landing #368. As you can tell from the comments in those, I'm not certain what the fate of that issue is though.

lib/formatter.js
```diff
@@ -0,0 +1,72 @@
...
+	let labelString = '';
+	const labelNames = getLabelNames(metric, defaultLabelNames);
+	for (let index = 0; index < labelNames.length; index++) {
+		const labelName = labelNames[index];
+		if (labelName === 'quantile') {
+			labelString += `\${val.labels.quantile != null ? \`quantile="\${escapeLabelValue(val.labels.quantile)}"\` : ''}`;
+		} else if (labelName === 'le') {
+			labelString += `\${val.labels.le != null ? \`le="\${escapeLabelValue(val.labels.le)}"\` : ''}`;
+		} else {
+			labelString += `${labelName}="\${val.labels['${labelName}'] ? escapeLabelValue(val.labels['${labelName}']) : escapeLabelValue('${defaultLabels[labelName]}')}"`;
```
zbjornson, line 41: Think this is equivalent and a little shorter:

```suggestion
			labelString += `${labelName}="\${escapeLabelValue(val.labels['${labelName}']  || defaultLabels[labelName])}"`;
```

lib/formatter.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+    return info + values;
+  }`);
+	return gen.toFunction();
+}
+
+function getLabelsCode(metric, defaultLabelNames, defaultLabels) {
+	let labelString = '';
+	const labelNames = getLabelNames(metric, defaultLabelNames);
+	for (let index = 0; index < labelNames.length; index++) {
```
zbjornson, line 34: ⭐ It's valid to only set values for a subset of labels when a metric is updated. With this test:

```js
it('should allow subsetse of labels', () => {
	instance = new Gauge({
		name: 'name_2',
		help: 'help',
		labelNames: ['code', 'success'],
	});
	instance.inc({ code: '200' }, 10);
	const str = globalRegistry.getSingleMetricAsString('name_2');
	console.log(str);
});
```

in this PR, you get 
```
# HELP name_2 help
# TYPE name_2 gauge
name_2{code="200",success="undefined"} 10
```

whereas in master, you get
```
# HELP name_2 help
# TYPE name_2 gauge
name_2{code="200"} 10
```

test/registerTest.js
```diff
@@ -181,49 +136,35 @@ describe('register', () => {
...
-						type: 'counter',
-					};
-				},
-			});
-			escapedResult = register.metrics();
-		});
-		it('backslash to \\\\', () => {
-			expect(escapedResult).toMatch(/\\\\/);
-		});
-		it('newline to \\\\n', () => {
```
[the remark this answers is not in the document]
doochik: These tests are useless because you can't create metric with name `test_"_\\_\n_metric`. It's not a valid name.

lib/counter.js
```diff
@@ -4,12 +4,16 @@
...
 const util = require('util');
-const type = 'counter';
 const { hashObject, isObject, getLabels, removeLabels } = require('./util');
 const { validateLabel } = require('./validation');
 const { Metric } = require('./metric');
 
 class Counter extends Metric {
+	constructor(...args) {
+		super(...args);
+		this.type = 'counter';
```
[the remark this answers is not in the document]
doochik: I can't define property in native ES classes :(

Do you mean `Counter.prototype.type = 'counter'`? It looks bad for me :(

In other case I can create getter for `type`:
```
class Counter extends Metric {
    get type() {
		return 'counter';
	}
}
```

lib/formatter.js
doochik: Can you clarify what do you mean? Should I rewrite it or implement inside generated function?

lib/counter.js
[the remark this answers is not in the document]
zbjornson: `Counter.prototype.type = 'counter'` is semantically correct even if it's a bit ugly. Getters have a small runtime overhead, so I would use the prototype.

lib/formatter.js
SimenB: should add this as a test, no?

doochik: Yes, of course. I’ll fix it

lib/formatter.js
doochik: @zbjornson can you take a look on this?

zbjornson: Thanks for the nudge, will review this PR this weekend.

On this particular comment: I'd (strongly) lean toward making the formatter work with dynamically declared labels. #298 has quite a bit of support so I suspect #368 will land. The main (probably minor) thing holding it up is https://github.com/siimon/prom-client/issues/298#issuecomment-620477855.

lib/formatter.js
zbjornson: The above test case passes, but now if the first label is undefined, there's a stray leading comma:

```
it.only('should allow combinations of labels', () => {
	instance = new Gauge({
		name: 'name_2',
		help: 'help',
		labelNames: ['code', 'success', 'ok'],
	});
	instance.inc({ code: '200' }, 10);
	instance.inc({ code: '200', ok: 'yes' }, 10);
	instance.inc({ ok: 'yes', success: 'false' }, 10);
	instance.inc({ success: 'false', ok: 'no' }, 10);
	const str = globalRegistry.getSingleMetricAsString('name_2');
	console.log(str);
});
```

prints

```
      # HELP name_2 help
      # TYPE name_2 gauge
      name_2{code="200"} 10
      name_2{code="200",ok="yes"} 10
      name_2{,success="false",ok="yes"} 10
      name_2{,success="false",ok="no"} 10
```

(notice last two lines)

Something like these should both be committed as test cases. Do you feel like adding them please?

---

Just checked, benchmarks still look significantly faster despite the new branch added in the last commits. 👍

lib/formatter.js
doochik: This eliminate the main idea of `formatter` - build labels string once.
I need to think how to rewrite this.

SimenB:
Any news here? 🙂

SimenB:
Does this still provide improvements after #594?

Issue #224, opened 2018-10-13
Request: an option for aggregating cluster metrics that preserves worker id


vincentwoo:
For background, I run a clustered Node setup where I'm interested in things like open file descriptors *per* worker (so we can display them as a stacked graph). There's not an easy way to do this for the metrics that are collected by default. I think it'd be useful to have a general aggregation pattern where a worker just sets a particular metric, and it's pulled into the master aggregation as-is with just a worker ID label applied.

Right now we're getting around this for some metrics by having child workers use IPC to communicate back to the master, setting worker id, which is clumsy.

zbjornson:
I agree this would be a reasonable option to offer (probably something like `collectDefaultMetrics({includeWorkerId: true})`). Happy to review a PR for it, assuming the other maintainers are good with it.

PR #374, opened 2020-06-11
feat: added support for adding labels to default metrics

rkoval:
I believe this resolves https://github.com/siimon/prom-client/issues/224. I was also needing a mechanism like this for applying labels to worker processes when running in a clustered environment.

I did it a bit more generically than just the suggested `collectDefaultMetrics({includeWorkerId: true})` because our project setup prevents us from using lib/cluster.js. We're using a pm2 wrapper around our nodejs parent process which has some nuances when using the nodejs `cluster` module directly from lib/cluster.js. As such, this should appease a wider audience of users who just want to generically apply labels to default metrics (including myself)

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,16 @@ const prefix = 'my_application_';
 collectDefaultMetrics({ prefix });
 ```
 
+To apply generic labels to all default metrics, pass an object to the `labels` property (useful if you're working in a clustered environment):
+
+```js
+const client = require('prom-client');
+const collectDefaultMetrics = client.collectDefaultMetrics;
+collectDefaultMetrics({
+  labels: { NODE_APP_INSTANCE: process.env.NODE_APP_INSTANCE },
+});
+```
+
 You can get the full list of metrics by inspecting
 `client.collectDefaultMetrics.metricsList`.
 
--- a/lib/metrics/version.js
+++ b/lib/metrics/version.js
@@ -8,11 +8,13 @@ const NODE_VERSION_INFO = 'nodejs_version_info';
 
 module.exports = (registry, config = {}) => {
 	const namePrefix = config.prefix ? config.prefix : '';
+	const labels = config.labels ? config.labels : {};
+	const labelNames = Object.keys(labels);
 
 	new Gauge({
 		name: namePrefix + NODE_VERSION_INFO,
 		help: 'Node.js version info.',
-		labelNames: ['version', 'major', 'minor', 'patch'],
+		labelNames: ['version', 'major', 'minor', 'patch', ...labelNames],
 		registers: registry ? [registry] : undefined,
 		aggregator: 'first',
 		collect() {
@@ -22,6 +24,7 @@ module.exports = (registry, config = {}) => {
 				versionSegments[0],
 				versionSegments[1],
 				versionSegments[2],
+				...Object.values(labels),
 			).set(1);
 		},
 	});
--- a/test/defaultMetricsTest.js
+++ b/test/defaultMetricsTest.js
@@ -69,15 +69,16 @@ describe('collectDefaultMetrics', () => {
 		}
 	});
 
-	it('should apply labels to metrics when configured', () => {
+	it('should apply labels to metrics when configured', async () => {
 		const labels = {
 			NODE_APP_INSTANCE: 0,
 		};
-		interval = collectDefaultMetrics({
+		collectDefaultMetrics({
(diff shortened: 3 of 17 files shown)


test/defaultMetricsTest.js
```diff
@@ -71,6 +71,21 @@ describe('collectDefaultMetrics', () => {
 		});
 	});
 
+	it('should apply labels to metrics when configured', () => {
+		const labels = {
```
SimenB, line 75: can you add a `expect.assertions(5)` (or whatever the number is)?

lib/metrics/version.js
```diff
@@ -8,24 +8,25 @@ const NODE_VERSION_INFO = 'nodejs_version_info';
...
 module.exports = (registry, config = {}) => {
 	const namePrefix = config.prefix ? config.prefix : '';
+	const labels = config.labels ? config.labels : {};
+	const labelNames = Object.keys(labels);
 
 	const nodeVersionGauge = new Gauge({
 		name: namePrefix + NODE_VERSION_INFO,
 		help: 'Node.js version info.',
-		labelNames: ['version', 'major', 'minor', 'patch'],
+		labelNames: ['version', 'major', 'minor', 'patch'].concat(labelNames),
```
SimenB, line 17:
```suggestion
		labelNames: ['version', 'major', 'minor', 'patch', ...labelNames],
```

I believe our node version support covers this syntax? Comment applies to the other `concat` calls as well

rkoval: https://github.com/rkoval/prom-client/commit/c34d28d30997c44b8263f84b92edea9fb1c6e97d

test/defaultMetricsTest.js
rkoval: https://github.com/rkoval/prom-client/commit/82a7106ec03f8b97c4ffd861d163d8b18e56ba6d

zbjornson: https://jestjs.io/docs/en/expect#expectassertionsnumber - it ensures that that many assertions are made, which makes sure tests don't falsely pass if some bug causes the async callback not to run.

test/defaultMetricsTest.js
```diff
@@ -84,6 +84,7 @@ describe('collectDefaultMetrics', () => {
 				expect(metricValue.labels).toMatchObject(labels);
 			});
 		});
+		expect.assertions(45);
```
zbjornson: (I'm not sure if it matters, but usually this is the first line of the `it` function.)

zbjornson:
I like this approach!

Sorry for the hassle but can you rebase on master please?

test/defaultMetricsTest.js
rkoval: do you know why this function produces inconsistent results / doesn't always have the same number of assertions? i can actually reproduce the inconsistencies locally: 
image
the first run has 47 assertions, but the second only has 45 (i didn't change any code between each run)

rkoval:
@zbjornson this has been rebased on the current master

test/defaultMetricsTest.js
SimenB: probably something async we're not properly waiting for

rkoval: ok, what would you like to do here? again, this type of assertion isn't used anywhere else in the codebase, so i might just have to close this PR and just use our own fork if it'll be blocked without it. i unfortunately don't have time to debug the issue

README.md
```diff
@@ -80,6 +80,16 @@ const prefix = 'my_application_';
 collectDefaultMetrics({ prefix });
 ```
 
+To apply generic labels to all default metrics, pass an object to the `labels` property (useful if you're working in a clustered environment):
+
+```js
+const client = require('prom-client');
+const collectDefaultMetrics = client.collectDefaultMetrics;
+collectDefaultMetrics({
+  labels: { NODE_APP_INSTANCE: process.env.NODE_APP_INSTANCE },
```
SimenB, line 89: need to add this to the TS types: https://github.com/siimon/prom-client/blob/48ed6c8b8112c5edf2b58985491a79632abe3112/index.d.ts#L567-L572

test/defaultMetricsTest.js
SimenB: went with a slightly different approach

SimenB:
@zbjornson @siimon PTAL

siimon:
👍

PR #375, opened 2020-06-25
Fix startTimer return type

Flaque:
# Overview

This PR fixes the TypeScript typings for `startTimer`. 

```ts
// Old 
startTimer(labels?: LabelValues<T>): (labels?: LabelValues<T>) => void;

// New 
startTimer(labels?: LabelValues<T>): (labels?: LabelValues<T>) => number;
```

This corresponds to the actual return value of the `startTimer` function [here.](https://github.com/siimon/prom-client/blob/master/lib/histogram.js#L110)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -361,7 +361,7 @@ export class Histogram<T extends string> {
 	 * @param labels Object with label keys and values
 	 * @return Function to invoke when timer should be stopped
 	 */
-	startTimer(labels?: LabelValues<T>): (labels?: LabelValues<T>) => void;
+	startTimer(labels?: LabelValues<T>): (labels?: LabelValues<T>) => number;
 
 	/**
 	 * Reset histogram values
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -276,7 +276,7 @@ export class Gauge<T extends string> {
 	 * @param labels Object with label keys and values
 	 * @return Function to invoke when timer should be stopped
 	 */
-	startTimer(labels?: LabelValues<T>): (labels?: LabelValues<T>) => void;
+	startTimer(labels?: LabelValues<T>): (labels?: LabelValues<T>) => number;
```
zbjornson, line 279: I'm not sure that this is correct. From testing, reading the code and comparing to the other definitions in index.d.ts, it looks like this is correct:

```ts
startTimer(labels?: LabelValues<T>): (labels?: LabelValues<T>) => void;
```

zbjornson: Ah I see part of the confusion: The code you linked to in your top post is for Histogram -- this is the Gauge class.

Flaque: Ah whoops. 🤦‍♂️

PR #377, opened 2020-07-08
a complete working example of a pushgateway push

mad01:
While trying to get up to speed and understand how to use the pushgateway api i got stuck not understanding
how to use the api. I have created a working example for how to send a counter metric to pushgateway. Was Thinking it might be useful to someone else to have a complete example of how to use it.


mad01:
were do i change that ?

boris-chernysh:
@mad01  CHANGELOG.md at the root of the project
an example can be seen in #388

PR #384, opened 2020-08-01
bug: fix regression in 4e6aacd

### LINKED ISSUE #382: Node app crashes for cluster_metrics API
Hi,
I am trying to implement cluster metrics using AggregatorRegistry. Earlier I tried it in my node application but after accessing the cluster_metric endpoint I used to get a blank response. So to understand how to implement it, I cloned the prom-client repository and tried running the cluster.js example which is included in the repo. But after accessing the cluster_metric on port number 3001, the app crashed with  **TypeError: metrics.forEach is not a function**. 

Screenshot 2020-07-31 at 11 29 30 AM

If anyone could please help me to find out what's going wrong.
Thanks.

zbjornson:
Fixes #382

This one's hard to test without forking processes...

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -43,7 +43,10 @@ class AggregatorRegistry extends Registry {
 		const requestId = requestCtr++;
 
 		return new Promise((resolve, reject) => {
+			let settled = false;
 			function done(err, result) {
+				if (settled) return;
+				settled = true;
 				if (err) reject(err);
 				else resolve(result);
 			}
@@ -53,11 +56,9 @@ class AggregatorRegistry extends Registry {
 				pending: 0,
 				done,
 				errorTimeout: setTimeout(() => {
-					request.failed = true;
 					const err = new Error('Operation timed out.');
 					request.done(err);
 				}, 5000),
-				failed: false,
 			};
 			requests.set(requestId, request);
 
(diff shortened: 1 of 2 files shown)


lib/cluster.js
```diff
@@ -178,12 +178,12 @@ function addListeners() {
 }
 
 // Respond to master's requests for worker's local metrics.
-process.on('message', message => {
+process.on('message', async message => {
```
SimenB, line 181: Is it safe to make this async?

zbjornson: I think so, but I wasn't handling the error. Just revised the commit and switched to a Promise.

lib/cluster.js
```diff
@@ -180,11 +180,18 @@ function addListeners() {
...
+		Promise.all(registries.map(r => r.getMetricsAsJSON()))
+			.then(metrics => {
+				process.send({
+					type: GET_METRICS_RES,
+					requestId: message.requestId,
+					metrics,
+				});
+			})
+			.catch(err => {
+				console.error('Failed to get metrics', err);
```
SimenB, line 192: should we use `process.send` to communicate _something_?

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - fix: provide nodejs_version_info metric value after calling `registry.resetMetrics()` (#238)
 - fix: provide process_max_fds metric value after calling `registry.resetMetrics()`
 - fix: provide process_start_time_seconds metric value after calling `registry.resetMetrics()`
+- fix: fix regression in 4e6aacd (unreleased) (#382)
```
SimenB, line 39: do we need to add this? since it's not released I think it might just cause confusion for people using the changelog to migrate?

zbjornson: I agree. CI will fail without it but looks like we can merge despite that failure.

SimenB (approved):
I like it 👍

siimon:
Awesome work! 👍

siimon (approved):

PR #386, opened 2020-08-05
Add function for setting a Histogram to zero for given labels

lukasstockner:
Since PR #139, histograms are initialized to zero if they do not use any labels - however, there currently is no way to do the same for histograms with labels (without directly accessing the internals of the histogram from the outside, which is of course not a great solution).

Therefore, this PR adds a function that lets the user explicitly zero out the metrics for given labels. Since prom-client has no way of knowing what the values of the labels can be, it's up to the user to call this with all expected values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- feat: added `zero()` to `Histogram` for setting the metrics for a given label combination to zero
+
 ## [13.1.0] - 2021-01-24
 
 ### Changed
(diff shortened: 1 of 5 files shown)


zbjornson (changes requested):
Excellent, thank you, this will fix #409. A few quick requests - can you please:

- [ ] Add the new function to index.d.ts
- [ ] Add a short mention in the readme
- [ ] Rebase on master

If you don't have time or I don't hear back in a few days, I can make these changes.

zbjornson (approved):
Awesome, thanks!

mkreidenweis-schulmngr:
I'd love to use this. Unfortunately it's not on npm yet. 
@siimon @zbjornson Any chance this will be released soon? Can I help somehow? :-)

Issue #254, opened 2019-04-12
metric.remove potentially broken


erdii:
First of all: Nice work! `prom-client` is my go-to prometheus exporter lib for node-based services! ❤

I think the `metric.remove` method added in #242 needs to be reworked because there is no safe way to hand over multiple label/value pairs. Or maybe I'm missing something.

Given a gauge metric, that I used to track this value:
```js
metric.set({
	first: "foo",
	second: "bar",
}, 1337)
```

I call `metric.remove("foo", "bar")` and I have to rely on js object-key ordering, which works in V8, but is not guaranteed in the language spec.

What I'd like to call is:
```js
metric.remove({
	first: "foo",
	second: "bar",
});
```

Edit: Formatting

siimon:
PR welcome ☺️

erdii:
@siimon assign me - I'll see what I can do :)

erdii:
uh uh... i was not assigned and totally forgot about this one. will tackle it in the next weeks

pepe79:
That bit me today. The remove function is also completely left out in the README.

boris-chernysh:
> I call metric.remove("foo", "bar") and I have to rely on js object-key ordering, which works in V8, but is not guaranteed in the language spec.

When the delete method is called, an object is formed from labelNames and the passed values. So in fact, the call you make and the one you need are identical at some point.

The hash is calculated for key and valuBut I like the proposed interface because it is more consistent with the format when calling many methods and allows you to work with "holes" in labels.e pairs. When generating the hash, the keys are sorted by the built-in javascript sort method. the specification does not oblige it to be stable. So even if the interface changes, the problem remains :) 

But I like the proposed interface because it is more consistent with the format when calling many methods and allows you to work with "holes" in labels.

alexey-sh:
is it resolved?

erdii:
i guess so: https://github.com/siimon/prom-client/commit/3a86d05c49addc8294e9dcf16600226d6b970385

closing this

PR #387, opened 2020-08-08
Added the ability to pass labels as an object to labels() and remove()

boris-chernysh:
Based on the issue #254 

But additionally solves another problem:
```javascript
const c = new Counter({ name: 'test', help: 'test', labelNames: ['foo', 'bar','baz'] })
c.inc({ foo: 1, baz: 2 })
c.remove(1, undefined, 2) // dont work
c.remove(1, null, 2) // dont work
c.remove({ foo: 1, baz: 2 }) // work
```
Also now calls can be visually consistent.

I don't think this is a big problem, so I didn't remove the old behavior to avoid breaking changes.

(The deleted test was a duplicate. See above)


SimenB (changes requested):
I like it 🙂 Can you update the TS types as well?

siimon (approved):
No other comments than the type definitions as @SimenB mentioned 👍

SimenB (approved):

README.md
```diff
@@ -297,7 +297,7 @@ const gauge = new client.Gauge({
 // 1st version: Set value to 100 with "method" set to "GET" and "statusCode" to "200"
 gauge.set({ method: 'GET', statusCode: '200' }, 100);
 // 2nd version: Same effect as above
-gauge.labels('GET', '200').set(100);
```
SimenB, line 300: I think we should still keep this in the docs, but perhaps below the object one?

test/counterTest.js
```diff
@@ -80,13 +90,6 @@ describe('counter', () => {
 				expect(fn).toThrowErrorMatchingSnapshot();
 			});
 
-			it('should throw error if label lengths does not match', () => {
```
SimenB, line 83: it should still throw an error unless the argument is a plain error

SimenB (approved):

test/counterTest.js
boris-chernysh: I removed these lines because this test is a complete duplicate of the one above it

README.md
boris-chernysh: yes I agree, I will do it soon

boris-chernysh: done

stavalfi:
can we merge this? 👍

zbjornson (approved):

zbjornson:
Landed in 3a86d05. Sorry for the delay and thanks for the contribution!

Issue #390, opened 2020-08-13
Pushgateway call registry async metrics methods


zD98:
https://github.com/siimon/prom-client/blob/25f9c4059d8a0f2d02ccfec7acfb64157c8b72bf/lib/pushgateway.js#L78

`metrics` is an async method. 

So should calling `metrics` with `async/await` ?
```javascript
const metrics = await this.registry.metrics();
req.write(metrics);
```

zbjornson:
Yes, I think you're right. I missed two high-level functions evidently :(. PR welcome, otherwise I'll fix it in the next few days. Thanks for finding that.

rassokhin-s:
Hey guys @zD98 @zbjornson. I have just met the same bug. Do you have any plans to merge PR and release a new version?

neyio:
This issue still occurs at version 13.Please publish a new sub-version to resolve it, thanks.

zbjornson:
This was fixed in #419 (thanks @adorin). Until the new version is published, you can install from github using `yarn add siimon/prom-client#83f7867` or `npm i siimon/prom-client#83f7867`.

zbjornson:
v13.1.0 is now published with this fix!

PR #392, opened 2020-08-24
fix(pushgateway): use async/await gather metrics in pushgateway

zD98:
#390 
registy `metrics` method is an async function.
So update `useGateway` to an async method and use `async/await` to gather  metrics.

### DIFF
--- a/lib/pushgateway.js
+++ b/lib/pushgateway.js
@@ -39,7 +39,7 @@ class Pushgateway {
 		useGateway.call(this, 'DELETE', params.jobName, params.groupings, callback);
 	}
 }
-async function useGateway(method, job, groupings, callback) {
+function useGateway(method, job, groupings, callback) {
 	// `URL` first added in v6.13.0
 	// eslint-disable-next-line node/no-deprecated-api
 	const gatewayUrlParsed = url.parse(this.gatewayUrl);
@@ -75,8 +75,14 @@ async function useGateway(method, job, groupings, callback) {
 	});
 
 	if (method !== 'DELETE') {
-		const metrics = await this.registry.metrics();
-		req.write(metrics);
+		this.registry
+			.metrics()
+			.then(metrics => {
+				req.write(metrics);
+			})
+			.catch(e => {
+				callback(e);
+			});
 	}
 	req.end();
 }
--- a/lib/pushgateway.js
+++ b/lib/pushgateway.js
@@ -39,7 +39,7 @@ class Pushgateway {
 		useGateway.call(this, 'DELETE', params.jobName, params.groupings, callback);
 	}
 }
-function useGateway(method, job, groupings, callback) {
+async function useGateway(method, job, groupings, callback) {
 	// `URL` first added in v6.13.0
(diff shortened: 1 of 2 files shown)


lib/pushgateway.js
```diff
@@ -75,7 +75,8 @@ function useGateway(method, job, groupings, callback) {
 	});
 
 	if (method !== 'DELETE') {
-		req.write(this.registry.metrics());
+		const metrics = await this.registry.metrics();
+		req.write(metrics);
```
zbjornson, line 79: I'd rather not mix a callback with async/await, although you could do it if you add a `try/catch` block here and pass the caught exception to the callback. Coincidentally, I started working on this issue a bit before you opened this PR. This is what I had so far using Promise/then/catch:

```diff
        if (method !== 'DELETE') {
-               req.write(this.registry.metrics());
+               this.registry
+                       .metrics()
+                       .then(metrics => {
+                               req.write(metrics);
+                               req.end();
+                       })
+                       .catch(err => {
+                               callback(err);
+                       });
+       } else {
+               req.end();
        }
-       req.end();
```

(I was trying to see if I could easily write a test for it, so didn't open a PR.)

zD98: It's ok to use promise instead of async/await.
I have updated mr.

lib/pushgateway.js
```diff
@@ -75,7 +75,14 @@ function useGateway(method, job, groupings, callback) {
...
+		this.registry
+			.metrics()
+			.then(metrics => {
+				req.write(metrics);
+			})
+			.catch(e => {
+				callback(e);
+			});
 	}
 	req.end();
```
danielkatz, line 87: In the current implementation, there is a subtle bug, `req.end()` might happen before `req.write(metrics)` is called because the async version of `registry.metrics()` is non-blocking.
I suggest making `useGateway()` async. This is the only easy way of keeping the existing flow semantics.
```javascript
async function useGateway(method, job, groupings, callback) {

        ...

	if (method !== 'DELETE') {
		try {
			const metrics = await this.registry.metrics();
			req.write(metrics);
		} catch (error) {
			callback(error);
		}
	}
	req.end();
}
```

zbjornson: @danielkatz I don't really want to mix callbacks and `async/await`.

@zD98 danielkatz is correct about the bug though. Below is the fix I was thinking of:

```js
	if (method !== 'DELETE') {
		this.registry
			.metrics()
			.then(metrics => {
				req.write(metrics);
				req.end();
			})
			.catch(err => {
				callback(err);
			});
	} else {
		req.end();
	}
```

zbjornson:
@zD98 do you think you'll have time in the next few days to finish this? A few people have commented on open issues that they're eager to have a new release out, and I think this is the last change we need. I'm happy to take over this PR if you don't have time.

cah4a:
Hi

Any progress here?

zbjornson:
Replaced by #419. Thanks for starting this, your name was included in the commit for attribution!

PR #413, opened 2020-12-14
Allow readonly `labelNames` in metric configuration

GabrielCastro:
This allows using `as const` to infer the type of labels


zbjornson:
Thanks for this! I needed to rebase this before merging but didn't want to push to your master branch, so I landed it directly as 487611c.

PR #419, opened 2021-01-11
Fix push gateway attempting to write promise instead of awaiting

adorin:
Added tests for push gateway awaiting on registry.metrics() #390 

Includes the fix proposed in https://github.com/siimon/prom-client/pull/392/files#r487438436

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- fix: push client attempting to write Promise
+
 ### Added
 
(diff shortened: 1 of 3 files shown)


adorin:
@zbjornson I took the liberty to commit your proposed fix and add tests, hoping this would expedite the process.

zbjornson (approved):
Yay, thanks so much!

Will give the other maintainers a chance to review, but for the sake of time will merge tomorrow if there are no other comments.

zbjornson:
Thanks again!

SleepWalker:
Hello, when this will be released?

PR #422, opened 2021-01-23
Set aggregation method for newer event loop metrics

### LINKED ISSUE #418: Advanced event loop monitoring not aggregated correctly across cluster
Hi everybody, 

could it be the the cluster support (#147) was overlooked when introducing the advanced event loop monitoring in #278?

The old event loop metric has an `aggregator` set, whereas all the new onen don't: 
https://github.com/siimon/prom-client/blob/c63689bbbb6a0395ce9d806b0cc17cf5d380f4a8/lib/metrics/eventLoopLag.js#L37

Was looking at some metrics of a node cluster today and couldn't quite make sense of it. It looks to me like something else than the default `sum` aggregator should be used for most the other lag metrics, right?

zbjornson:
These are not entirely accurate: they're the "mean of the mean" and "mean of percentiles," but that's as good as we can get.

No tests added - doesn't seem worth asserting that these properties are set.

Fixes #418

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Don't add event listener to `process` if cluster module is not used.
 - fix: set labels for default memory metrics on linux
 - fix: fix DEP0152 deprecation warning in Node.js v16+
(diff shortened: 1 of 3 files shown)


mkreidenweis-schulmngr:
As you're saying, the aggregation method isn't ideal. For the mean it should actually be fine, as the sample size should actually be the same for all cluster nodes, right? Maybe you want to add a short note about these concerns to the respective Readme section to set the right expectations.

Issue #410, opened 2020-12-08
Adopt OpenMetrics


SnosMe:
https://tools.ietf.org/html/draft-richih-opsawg-openmetrics-00

- gaugehistogram
- exemplars
- changes for Counter (`_total` suffix)
- `_created`
- `# EOF`
etc...

samarara:
@SnosMe `_total` suffix can be provided by the `name` property can it not?

gavnate:
Hi everyone, 

I have implemented prom-client in a typescript project and noticed that the latest helm for prometheus does not scrape JSON but opts for the openmetrics format instead. Does anyone on this issue know of any workarounds while this work is in progress?

Thanks in advance!

PR #432, opened 2021-04-01
feat: add support for exemplars

samarara:
With exemplar support coming to Prometheus version 2.26.0 I thought I could start on this. Also addresses [#410](https://github.com/siimon/prom-client/issues/410). 

Looking for some direction to see if this the right approach to take. I started with Counter and I'm going to add another function to registry to call depending on `register.contentType`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - feat: added `zero()` to `Histogram` for setting the metrics for a given label combination to zero
+- feat: add support for exemplars
 
 ## [13.1.0] - 2021-01-24
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -62,7 +62,7 @@ export class Registry {
 	getSingleMetricAsString(name: string): Promise<string>;
 
 	/**
-	 * Gets the Content-Type of the metrics for use in the response headers.
+	 * Gets or Sets the Content-Type of the metrics for use in the response headers.
 	 */
 	contentType: string;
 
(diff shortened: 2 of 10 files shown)


samarara:
@siimon @zbjornson I implemented this similar to labels. It publishes metrics in a similar fashion as prometheus uses to parse them. If `content-type` is `application/openmetrics-text` the metrics are published with exemplars and the OpenMetrics parser is used in prometheus to parse them. 

Let me know if the direction I'm taking is okay. So far it satisfies everything in the [OpenMetrics exemplar spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#exemplars) except for the optional timestamp. I'll add that shortly.

samarara:
@zbjornson @siimon any advice on this?

shyimo:
@samarara thanks for this MR. i would love to help if you need.
We are using jaeger tracing system and really looking forward for this node.js client support

zbjornson:
Sorry for the delay. I need to read up on exemplars and OpenMetrics to review this, but am in favor of adding support. If someone else who's familiar with these concepts wants to help review, that'd be much appreciated!

It would also be helpful if someone can provide links to docs or code for the equivalent functionality in the other Prometheus client libraries (go/java/.NET/etc.).

samarara:
This MR is a bit out dated as the counter logic has changed since I authored this. Any help would be greatly appreciated. Here is the spec:
https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#exemplars

For the Java client it's implemented for both counters and histograms: https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#exemplars-1
implementation: https://github.com/prometheus/client_java/blob/7a29af04821cd51592c45ce4682c3132372f4c5d/simpleclient/src/main/java/io/prometheus/client/exemplars/Exemplar.java#L10

SimenB:
This landed in #544 I believe - sorry about the radio silence!

PR #440, opened 2021-05-12
Fix DEP0152 / deprecated PerformanceEntry.kind usage

### LINKED ISSUE #435: DeprecationWarning when running Node 16
Hi!

It seems starting with Node 16 there's a deprecating warning stemming from prom-client.

We're using a simple setup of `prom.collectDefaultMetrics();`

With node@16.0.0, npm@7.5.6 and using `--trace-deprecation` to run our server, we get the following (replaced irrelevant paths with `[...]`):

```
(node:22429) [DEP0152] DeprecationWarning: Custom PerformanceEntry accessors are deprecated. Please use the detail property.
    at PerformanceObserver.<anonymous> (/[...]/node_modules/prom-client/lib/metrics/gc.js:47:38)
    at PerformanceObserver.[kDispatch] (node:internal/perf/observe:261:34)
    at Immediate.<anonymous> (node:internal/perf/observe:94:25)
    at processImmediate (node:internal/timers:464:21)
/[...]/node/16.0.0/bin/node[22429]: ../src/env-inl.h:1052:void node::Environment::AddCleanupHook(node::Environment::CleanupCallback, void*): Assertion `(insertion_info.second) == (true)' failed.
 1: 0xb12b00 node::Abort() [/[...]/node/16.0.0/bin/node]
 2: 0xb12b7e  [/[...]/node/16.0.0/bin/node]
 3: 0xb7a58c  [/[...]/node/16.0.0/bin/node]
 4: 0xd5f70b  [/[...]/node/16.0.0/bin/node]
 5: 0xd60bac  [/[...]/node/16.0.0/bin/node]
 6: 0xd61226 v8::internal::Builtin_HandleApiCall(int, unsigned long*, v8::internal::Isolate*) [/[...]/node/16.0.0/bin/node]
 7: 0x160c579  [/[...]/node/16.0.0/bin/node]
 ```

Looking at the changelog for Node 16, it seems this is most likely related to [this change](https://github.com/nodejs/node/pull/37136) that overhauled perf_hooks.

### PR DESCRIPTION
**Pull request by sirreal**
Fix #435

Node 16 deprecated `PerformanceEntry.kind`, moving it to `PerformanceEntry.detail.kind`.

Use `entry.detail.kind` if available. Fall back to deprecated `entry.kind`.

See https://nodejs.org/docs/latest-v16.x/api/deprecations.html#deprecations_dep0152_extension_performanceentry_properties

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 - Don't add event listener to `process` if cluster module is not used.
 - fix: set labels for default memory metrics on linux
+- fix: fix DEP0152 deprecation warning in Node.js v16+
 
 ### Added
 
--- a/lib/metrics/gc.js
+++ b/lib/metrics/gc.js
@@ -41,16 +41,16 @@ module.exports = (registry, config = {}) => {
 
 	const obs = new perf_hooks.PerformanceObserver(list => {
 		const entry = list.getEntries()[0];
+		// Node < 16 uses entry.kind
+		// Node >= 16 uses entry.detail.kind
+		// See: https://nodejs.org/docs/latest-v16.x/api/deprecations.html#deprecations_dep0152_extension_performanceentry_properties
+		const kind = entry.detail ? kinds[entry.detail.kind] : kinds[entry.kind];
 
 		// Convert duration from milliseconds to seconds
-		gcHistogram.observe(
-			Object.assign({ kind: kinds[entry.kind] }, labels),
-			entry.duration / 1000,
-		);
+		gcHistogram.observe(Object.assign({ kind }, labels), entry.duration / 1000);
 	});
 
-	// We do not expect too many gc events per second, so we do not use buffering
-	obs.observe({ entryTypes: ['gc'], buffered: false });
+	obs.observe({ entryTypes: ['gc'] });
 };
 
 module.exports.metricNames = [NODEJS_GC_DURATION_SECONDS];

### CODE REVIEW

**Comment by analytik** [comment]
While I'm grateful for this, maybe we should also fix the deprecated format and an ignored parameter, see the [node.js bug comment](https://github.com/nodejs/node/issues/38412#issuecomment-826854944).

If think `if (entry.detail.kind)` passes, we should assume a newer Node.js version and also instead of `{ entryTypes: ['gc'], buffered: false }` use `{ type: 'gc' }` - it will likely be more future-proof.

**Comment by sirreal** [comment]
Sure, `buffered: false` appears to be completely redundant as it [was the default value](https://nodejs.org/docs/latest-v15.x/api/perf_hooks.html#perf_hooks_performanceobserver_observe_options) and has now been removed. I'll push a change for that.

Regarding `type`/`entryTypes`, I understand these are synonymous, where the second is compatible with older versions:

```js
obs.observe({ type: 'gc' });
obs.observe({ entryTypes: ['gc'] });
```

`{ type: 'gc' }` may be the preferred version for observing a single type, but `{ entryTypes: ['gc'] }` remains a valid way of observing one or more types according to [the v16 documentation](https://nodejs.org/docs/latest-v16.x/api/perf_hooks.html#perf_hooks_performanceobserver_observe_options).

I don't believe checking `event.detail` for this will work. We would need to know whether to use `type` or `entryTypes` before we start observing, and we only see `event` after we've started observing. We may have to resort to inspecting `process.version`. I'd leave this as-is.

It may be worth asking for clarification of https://github.com/nodejs/node/issues/38412#issuecomment-826854944 around `type`/`entryTypes`. Would you like me to do that?

sirreal:
I dug a bit into the specs because https://github.com/nodejs/node/issues/38412#issuecomment-826854944 mentioned aligning with them.

Both [MDN](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/observe#parameters) and [the w3c spec](https://w3c.github.io/performance-timeline/#dom-performanceobserver-observe) describe `type` and `entryTypes`. The difference seems to set the _observer type_ to `"single"` or `"multiple"`, respectively. It's unclear to me what the implications are or if there are downsides to a "multiple" type observer observing a single type.

The spec also mentions `buffered`, so perhaps I'm not looking in the right place.

zbjornson (approved):
Excellent, thank you. Sorry for the slow review.

I pushed to update the changelog and get CI to run.

gruselhaus:
I'm running node `v16.6.1` along with prom-client `v13.1.0` and still get this deprecation warning.

Bildschirmfoto 2021-08-08 um 16 45 02

Is anyone able to reproduce this?

zbjornson:
@gruselhaus sorry, I just published [v13.2.0](https://github.com/siimon/prom-client/releases/tag/v13.2.0) with the fix.

PR #441, opened 2021-05-15
added histogram example with observe method using labels

waqashsn:
Added examples showing how to use `observe `method on histogram when dealing with labels. Such examples existed for other types of metrics but not for histograms.


zbjornson (approved):
Thanks!

PR #442, opened 2021-06-10
Merge master registries to cluster metrics (#183)

rsnigel:
Fixes issue #183 by merging the registries configured in the master AggregatorRegistry with the worker registries.
This fixes as well the error that incomplete node cluster metrics are returned, because default metrics collected via `collectDefaultMetrics` were missing from the master process.
From now on, AggregatorRegistry.setRegistries can be used to configure master process registries as well, that will be merged to the cluster metrics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ label. (See `example/server.js` for an example using
 Metrics are aggregated from the global registry by default. To use a different
 registry, call
 `client.AggregatorRegistry.setRegistries(registryOrArrayOfRegistries)` from the
-worker processes.
+master or worker processes.
 
 ## API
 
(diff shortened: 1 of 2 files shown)


rsnigel:
Is there anything missing to merge this PR?

zbjornson:
Sorry for the delay.

I don't think merging the master registry by default is a good idea because the master is usually unlike the workers, so aggregated metrics could be skewed. (For example, the master probably has a tiny amount of heap space compared to workers.) #280 achieves the same goal as this PR via opt-in behavior, but it needs tests. If you're up for writing tests for it, that would be super helpful!

(Aside, the company I work at doesn't monitor the cluster master at all. The master does so very little that it's not useful to monitor.)

zbjornson:
Closing per above, but happy to discuss further.

PR #443, opened 2021-06-11
fix: set labels for default memory metrics on linux

jamsinclair:
## The Problem

I noticed that the following default metrics were missing labels that were configured in the `collectDefaultMetrics` method
- `process_resident_memory_bytes`
- `process_virtual_memory_bytes`
- `process_heap_bytes`

## The Fix

Looks like `osMemoryHeapLinux.js` missed out in #374 😢 . We just need to pass through the labels from the config to the metrics.

--- 

Thank you so much for this library! It's amazing that in a few simple steps we can all get amazing metrics from our Node.js servers. Cheers 🙇 😊

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Don't add event listener to `process` if cluster module is not used.
+- fix: set labels for default memory metrics on linux
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


jamsinclair:
I'll follow up and see why it's not being captured by the Ubuntu tests 🤔

zbjornson:
> Will the tests pass on every platform if we remove this?

If you want to remove it, we can try!

It'd probably be better to not mock that value, and instead conditionally skip tests that fail.

jamsinclair:
Thanks, @zbjornson! I've removed the platform process, unfortunately, you'll need to manually approve the CI run 🙇 

From my own manual testing it's working with:
- Linux (using the debian node:14 docker image) ✅ 
- MacOS (Big Sur) ✅

zbjornson (approved):
Looks great, thank you.

I rebased to fix the conflict in CHANGELOG.md.

PR #446, opened 2021-07-06
Use Nullish Coalescing Operator in inc()

dapplion:
Calling `.inc(value)` with a dynamic value that may be 0 or 1 will always cause to increase metrics by 1. This is very problematic and may distort the metrics in a non obvious way.

Closes https://github.com/siimon/prom-client/issues/447

### DIFF
--- a/lib/gauge.js
+++ b/lib/gauge.js
@@ -146,14 +146,16 @@ function startTimer(startLabels) {
 function dec(labels) {
 	return value => {
 		const data = convertLabelsAndValues(labels, value);
-		this.set(data.labels, this._getValue(data.labels) - (data.value || 1));
+		if (data.value === undefined) data.value = 1;
+		this.set(data.labels, this._getValue(data.labels) - data.value);
 	};
 }
 
 function inc(labels) {
 	return value => {
 		const data = convertLabelsAndValues(labels, value);
-		this.set(data.labels, this._getValue(data.labels) + (data.value || 1));
+		if (data.value === undefined) data.value = 1;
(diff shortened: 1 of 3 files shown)


zbjornson:
Good find. We can't use the `??` operator though because we currently support older versions of Node. Can you use an alternative fix please?

dapplion:
> Good find. We can't use the `??` operator though because we currently support older versions of Node. Can you use an alternative fix please?

Sure, let me know if this approach is more suitable

https://github.com/siimon/prom-client/blob/21406a0a43455ce8e6a4b705afd9333ce7398114/lib/gauge.js#L149

zbjornson:
That works! Can you also add a test and update the changelog please?

lib/gauge.js
```diff
@@ -146,14 +146,16 @@ function startTimer(startLabels) {
 function dec(labels) {
 	return value => {
 		const data = convertLabelsAndValues(labels, value);
-		this.set(data.labels, this._getValue(data.labels) - (data.value || 1));
+		const value = data.value === undefined ? 1 : data.value
```
zbjornson, line 149: Need a different variable name since `value` is also a parameter name.
Also need a semicolon at the end.

(ditto for changes below)

zbjornson (changes requested):
Sorry about the delay.

Two little changes to make, then should be set! Thanks.

zbjornson (approved):
Thanks!

I changed the impl a little bit so the linter passes, and added another test.

PR #449, opened 2021-07-10
Lazily add process event listener

### LINKED ISSUE #448: Listener added to process even if not using clusters
Hi,

I encountered a memory leak because of a listener added on `process` in the `lib/cluster.js` file:
https://github.com/siimon/prom-client/blob/c7b9a9ddd048f6a1921fa1b33c56a5296713aa2b/lib/cluster.js#L186

It happens after requiring several times a module which requires `prom-client` in my tests while using [`jest.resetModules()`](https://jestjs.io/docs/jest-object#jestresetmodules) (which resets the module registry).
When doing this I encounter this error:
`(node:12735) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 message listeners added to [process]. Use emitter.setMaxListeners() to increase limit`

This problem was actually reported in #155 and fixed in #159 by adding the event listeners only if the cluster module is used.

However this behavior was then partially modified in #294 for #293. In this PR, an event is attached to `process` even if the cluster module isn't used. This was not detected by the test added in #159 because it is only ensuring that no listener is added to `cluster` (and not to `process`):
https://github.com/siimon/prom-client/blob/c7b9a9ddd048f6a1921fa1b33c56a5296713aa2b/test/clusterTest.js#L6-L18

I think that one way to fix this would be to add the listener to `process` within the [`addListeners()`](https://github.com/siimon/prom-client/blob/c7b9a9ddd048f6a1921fa1b33c56a5296713aa2b/lib/cluster.js#L153) function which is already used to add the listener to `cluster`. By doing so, the listener won't be added if `AggregatorRegistry` was not instantiated.
I opened a PR (#449) doing this and the tests are passing, but I'm not familiar with the way the cluster mode is working. So even if it seems ok to me according to the way it's used in [`example/cluster.js`](https://github.com/siimon/prom-client/blob/c7b9a9ddd048f6a1921fa1b33c56a5296713aa2b/example/cluster.js#L7), you might want to double check that I didn't break anything there :slightly_smiling_face:

FabienChaynes:
Only attach a listener to `process` when the cluster mode is used to prevent a memory leak after multiple requires.

The tests are passing, but I'm not familiar with the way the cluster mode is working. So even if it seems ok to me according to the way it's used in [`example/cluster.js`](https://github.com/siimon/prom-client/blob/c7b9a9ddd048f6a1921fa1b33c56a5296713aa2b/example/cluster.js#L7), you might want to double check that I didn't break anything there :slightly_smiling_face:

Fixes #448.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- Don't add event listener to `process` if cluster module is not used.
+
 ### Added
 
 - feat: added `zero()` to `Histogram` for setting the metrics for a given label combination to zero
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -180,27 +180,29 @@ function addListeners() {
 			}
 		});
 	}
-}
 
-// Respond to master's requests for worker's local metrics.
-process.on('message', message => {
-	if (cluster().isWorker && message.type === GET_METRICS_REQ) {
-		Promise.all(registries.map(r => r.getMetricsAsJSON()))
-			.then(metrics => {
-				process.send({
-					type: GET_METRICS_RES,
-					requestId: message.requestId,
-					metrics,
-				});
-			})
-			.catch(error => {
-				process.send({
-					type: GET_METRICS_RES,
-					requestId: message.requestId,
-					error: error.message,
-				});
-			});
+	if (cluster().isWorker) {
+		// Respond to master's requests for worker's local metrics.
+		process.on('message', message => {
+			if (message.type === GET_METRICS_REQ) {
+				Promise.all(registries.map(r => r.getMetricsAsJSON()))
+					.then(metrics => {
+						process.send({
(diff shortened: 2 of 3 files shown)


zbjornson:
Thanks for opening this. Initial 👍 , will review fully this weekend.

zbjornson (approved):
Great, thanks!

zbjornson:
I just realized that this actually broke cluster metrics. `addListeners()` is only called when `new AggregatorRegistry()` is called, and that's not supposed to be instantiated on workers, only the master. (I'll fix up the example so it follows that pattern.)

I don't see a way to do what you're trying to do here without requiring users to call a setup function explicitly. Being explicit is probably best though.

Working on a patch for v13 and v14 now.

> prevent a memory leak after multiple requires.

Ugh, jest.

jdmarshall:
The person filing this PR doesn’t understand how `require()` works. And it’s even more difficult to load a single file multiple times with `import`.

PR #452, opened 2021-08-01
Reduce memory allocations/improve perf of Gauge and Counter

zbjornson:
No need to create and return new functions each time the gauge methods are called. Also saves ~40 LOC.


zbjornson:
Benchmark is noisy, but change is consistently an improvement.
```
✓ gauge ➭ inc is 57.04% faster.
✓ gauge ➭ inc with labels is 5.850% faster.
```

siimon (approved):

Issue #179, opened 2018-03-26
De-emphasise resetting metrics to avoid confusion


brian-brazil:
The main readme mentioned resetting metrics in several places. While there are rare valid use cases for this involving unittesting, most users get confused by this and think they need to reset at every scrape or have other misunderstandings. I'd suggest not mentioning this functionality.

SimenB:
Yeah, I've never used it myself. I have cleared all metrics for testing purposes, but I've never cleared a metric.

👍 for removing its mention from the readme (at least from every single metric type).

zbjornson:
Fixed in #243 🎉

PR #456, opened 2021-08-19
Add example of using .reset()

toymachiner62:


zbjornson:
Thanks for the PR, but `reset()` isn't really supposed to be used except for unit testing (see #179), so I don't think we should include an example of this.

zbjornson:
However, if you want to change the JSDoc code comments to indicate that `reset` is not intended for normal use, I'd be happy to accept that change!

Issue #370, opened 2020-05-04
Improve Event Loop Lag Metric


ChristianBoehlke:
Hello,

Thank you for all the work with this library!
We are using our own implementation of the event loop lag metric because the [default metric](https://github.com/siimon/prom-client/blob/master/lib/metrics/eventLoopLag.js) did not work well for us and I was wondering whether this is something we could contribute back? Here are the two things we have changed:

**Reset libuv-Histogram**
The `mean`/`max`/`percentiles` that libuv provides (through `perf_hooks.monitorEventLoopDelay`) are never reset. After a couple of days all numbers are pretty much set and do not change anymore. There is no way to even distinguish between high load times during the day and quiet times during the night. I think it would be better to expose a moving average/min/max instead of the total-average/min/max.

By default the `resolution` that libuv uses is 10ms which means that libuv generates ~100 measurements per second. It would be possible to reset libuv's histogram every second (or so) and record its `mean` and `max` in circular buffers. The length of the buffers should be a bit longer than Prometheus' scape interval and get averaged when collected.
Instead of the circular buffer, a histogram could be used as well (proposed in #309).
There is some discussion about this in #278.

**Subtract `resolution` from values**
libuv uses [a timer](https://github.com/nodejs/node/blob/bde5f9baf7af4efeec1a0ea1d53b70152e23428f/src/node_perf.cc#L540-L543) and records [the time that has passed since the last invocation](https://github.com/nodejs/node/blob/bde5f9baf7af4efeec1a0ea1d53b70152e23428f/src/node_perf.cc#L513-L534). This means that all values that libuv provides have the `resolution` value "added" to them. I think it would be better if this library could subtract this number before exposing the metrics. This could also be done in Prometheus but this is a rather specific implementation detail and could vary from process to process because it is configurable.
There is some discussion about changing how libuv measures the lag but this didn't get merged: https://github.com/nodejs/node/pull/32018 and https://github.com/nodejs/node/pull/32102. There is also some future work in https://github.com/libuv/libuv/pull/2725.

I am happy to open a PR for this so that you can have a look.

sam-github:
Hi, appreciate the feedback, and reading through your comments in detail is on my list.

Any interest in making a PR?

Also, pending me looking more closely, do you think some of this should be reported directly to nodejs as issues with our core APIs?

jomi-se:
[...]

Issue #309, opened 2019-12-28
Event loop lag should be a summary, not a gauge


dhet:
Hi,
as stated in the title, I believe the event loop lag metric should be a summary, not a gauge. The reason is that event loop lag can fluctuate quite substantially in a very short time. Think of a web server: when a web server handles a request, and we take a lag sample at exactly that moment then the lag value is is probably high. If we take a sample in idle state on the other hand, lag should be close to zero. Now when we plot the gauge metric in a line chart we can see seemingly random lag spikes. The spike might be there or not depending on whether a sample was taken during a request or during idle time. A summary would produce much more meaningful results as it records not only a single point in time but also considers all previous samples. 

Have a look at the official docs: [`monitorEventLoopDelay`](https://nodejs.org/api/perf_hooks.html#perf_hooks_perf_hooks_monitoreventloopdelay_options) also uses a histogram. 

I can open a PR if you want, otherwise #278 would be a good opportunity to make this change.

dhet:
This is what it looks like currently:
<img width="1140" alt="image" src="https://user-images.githubusercontent.com/2668621/71549080-d3f4a100-29b7-11ea-860e-f1ad380f6a8a.png">
Quite random I would say


Here's what I think it should look like:
<img width="1129" alt="image" src="https://user-images.githubusercontent.com/2668621/71549076-c2ab9480-29b7-11ea-89bb-505d96ff018f.png">

sam-github:
Shouldn't it be a histogram?

siimon:
Yeah, histogram is the way to go

dhet:
Care to explain why? Isn't a summary just a histogram with percentiles as buckets?

sam-github:
Histograms have fixed buckets, with counts in them, so its possible to sum the histograms across all the instances (pods for example). Summaries the data is analyzed on the fly, each pod has different values for the percentiles, and they can't be summed. If you google around you'll find a bunch of info on this. Summaries look great until someone tries to aggregate into a dashboard, then it goes badly.

There is a case to be made that the user should be able to provide either a histogram or summary object to be used for the observation of the values (go does this with some metrics it collects), but I haven't seen the Node.js client do this, so if there is only one, it should be histogram.

zbjornson:
[...]

PR #459, opened 2021-08-21
Reset internal histogram of monitorEventLoopDelay after each collect() invocation

yarsky-tgz:
This PR addresses problem described at #370. 

I've read attentively that issue and propositions about the solution. Anyway after own investigation of the code I've decided that simple calling `histogram.reset()` after each `collect()` invocation will be simply enough.

Event loop lag metrics is important for us because we want to see how our application handles spikes and how internal event loop lag increases, to determine easier, there is hiding the bottleneck. Currently we just cannot see that, because `nodejs_eventloop_lag_seconds` represents just single measurement result (so it is pretty random, as noticed already at #309) and `nodejs_eventloop_lag_mean_seconds` shows mean value for overall process lifetime and I cannot imagine, how to use such information in our case and for what cases it can be useful at all.

If all collected now event loop lag values will represent only scrape interval period that will be simply enough for us and I think that for everyone too.

### DIFF
--- a/lib/metrics/eventLoopLag.js
+++ b/lib/metrics/eventLoopLag.js
@@ -60,6 +60,8 @@ module.exports = (registry, config = {}) => {
 			lagP50.set(labels, histogram.percentile(50) / 1e9);
 			lagP90.set(labels, histogram.percentile(90) / 1e9);
 			lagP99.set(labels, histogram.percentile(99) / 1e9);
+
+			histogram.reset();
 		};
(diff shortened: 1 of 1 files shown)


zbjornson (approved):
Seems reasonable, left some additional notes in https://github.com/siimon/prom-client/issues/370#issuecomment-922402070.

I'll update the changelog separately.

Thanks for the contribution!

PR #464, opened 2021-09-19
fix cluster metrics when AggregatorReg isn't created on workers

zbjornson:
When using the cluster module, you must now call `promclient.setupClusterWorker()` from each cluster worker.

v13.2.0 introduced a change from v13.1.0 (#449) that broke cluster metrics if an `AggregatorRegistry` was not instantiated on each cluster worker. The example in `examples/cluster.js` shows instantiation of an `AggregatorRegistry` in the workers, so users following that example were not affected by this change. However, the example was not written as intended: `new AggregatorRegistry()` should only be called on the cluster master. `examples/cluster.js` has been updated to show the full, correct usage.

---

This also unblocks a new approach to #182.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,17 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Breaking
 
+- When using the cluster module, you must now call
+  `promclient.setupClusterWorker()` from each cluster worker.
+
+  Long explanation: v13.2.0 introduced a change from v13.1.0 that broke cluster
+  metrics if an `AggregatorRegistry` was not instantiated on each cluster
+  worker. The example in `examples/cluster.js` shows instantiation of an
+  `AggregatorRegistry` in the workers, so users following that example were not
+  affected by this change. However, the example was not written as intended:
+  `new AggregatorRegistry()` should only be called on the cluster master.
+  `examples/cluster.js` has been updated to show the full, correct usage.
+
 ### Changed
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -481,6 +481,8 @@ register
   });
 ```
 
+Note that you must call `client.setupClusterWorker()` in each worker.
+
 ### Pushgateway
 
 It is possible to push metrics via a
--- a/example/cluster.js
+++ b/example/cluster.js
@@ -2,17 +2,20 @@
 
 const cluster = require('cluster');
 const express = require('express');
-const metricsServer = express();
-const AggregatorRegistry = require('../').AggregatorRegistry;
-const aggregatorRegistry = new AggregatorRegistry();
+const prometheus = require('../');
 
 if (cluster.isMaster) {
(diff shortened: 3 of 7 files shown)


lucianodltec:
Desperated need this fix! I'm getting some errors running metrics on nodejs cluster. The metrics are not being aggregated, it´s simply does not work on cluster. Is this a fix for it?

CHANGELOG.md
```diff
@@ -9,6 +9,17 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Breaking
 
+- When using the cluster module, you must now call
+  `promclient.setupClusterWorker()` from each cluster worker.
+
+  Long explanation: v13.2.0 introduced a change from v13.1.0 that broke cluster
```
SimenB, line 15: sounds like we should revert that change, release a patch, then a new major with that and this?

SimenB (approved):

CHANGELOG.md
zbjornson: That's a good idea, I can release a v13.2.1 with it reverted.

zbjornson:
@lucianodltec check out #280. That PR just needs tests. If you're willing to work on them, that would be awesome!

theoilie:
Wondering if there are plans to merge this PR? It looks super helpful!

lebrak:
Hi thank you !! 
I need this PR too, when this PR should be merge plz ?

SimenB:
@zbjornson could you target `next` with this?

zbjornson:
@SimenB will do shortly. I want to revisit this PR to see if I can make both patterns work so it's not a breaking change in either direction.

ethan-burrell:
@zbjornson 

This would be great to unblock some work, looking forward to this being merged!

MichaelSitter:
👋 👋 Hey team!! Any movement on this issue? We are in need of this fix, hoping it's landing soon 🤞

jasongornall-dd:
Bumping this! Would love to have this merged.

morgaan:
@zbjornson Any update on this one? We are stuck with `13.1.0` and would be happy to migrate to latest. Thank you in advance 🤞

SimenB:
@zbjornson I plan to release a stable v15 this week. Will you be able to work on the two issues in https://github.com/siimon/prom-client/issues?q=is:open+label:semver-major? If yes, I'll hold off - if not I'll just roll it out 🙂  Been more than half a year since exemplars landed without a release 😅

CHANGELOG.md
NeoPhi: It looks like v13.2.1 was never released so would it be possible to get this MR merged? The cleaner API is nice.

jdmarshall:
I don’t think this is the way to go. Particularly now since worker support has replicated this issue. I think it would be better to either make the current example canonical - include the cluster file before forking children rather than after, or make sure redundant reloads of the file unregister existing instances of the listener.

PR #471, opened 2021-10-01
Support for getting clustered metrics as object

### LINKED ISSUE #470: Please consider adding a getClusterMetricsAsJSON funtion
At the moment it seems the cluster metrics are only available in the string format and not the JSON format.
I seems like it would be trivial to also add the JSON variant to the clusterMetrics function as well.
This way it would mirror how the non clustered metrics Registry interface work.

PatrLind:
As requested in issue #470, here is a PR
New functions:
getClusterMetricsAsJSON()
getClusterMetricsAsArray()

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -93,6 +93,20 @@ export class AggregatorRegistry extends Registry {
 	clusterMetrics(): Promise<string>;
 
 	/**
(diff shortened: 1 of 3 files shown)


jdmarshall:
There's new work to do this in progress, also this should now need to cover worker.js as well.

PR #474, opened 2021-10-18
Add eventloop idle time metric

alex3d:


SleepWalker:
Wow. I've thought that was already merged. Interesting why it wasn't? :thinking: It seems to me like a much better alternative to CPU usage

PR #476, opened 2021-11-02
chore: test using node 17

SimenB:
Also use built-in caching of `actions/setup-node`


SimenB:
Bah, builtin cache doesn't work without a lockfile

PR #477, opened 2021-11-10
fix getMetricsAsArray type

thierrylamarre:
This fixes the registry's `getMetricsAsArray()` function type.
It appears to be currently incorrect as it does not return a promise:
https://github.com/siimon/prom-client/blob/4e6aacd4921a3791e8f01ac6ab2fd6bb421b0dc0/lib/registry.js#L21-L23

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Breaking
 
+- types: fixed type for `registry.getMetricsAsArray()`
+
 - changed: `linearBuckets` does not propagate rounding errors anymore.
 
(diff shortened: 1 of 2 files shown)


zbjornson (approved):
Thanks!

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Breaking
 
+- types: fixed type for `registry.getMetricsAsArray()`
```
zbjornson, line 26: Since this isn't a breaking change, can you move it from here to `### Changed` please?

thierrylamarre: erf yes sorry about that. It's not in the right version either, moving it.

zbjornson (approved):
Thanks :)

PR #485, opened 2021-12-24
Use `process.getActiveResourcesInfo()` in process metrics

RaisinTen:
We should use `process.getActiveResourcesInfo()` here because it is a
public alternative of the private APIs `process._getActiveHandles()` and
`process._getActiveRequests()`.

Refs: https://nodejs.org/api/process.html#processgetactiveresourcesinfo

Signed-off-by: Darshan Sen <darshan.sen@postman.com>


RaisinTen:
cc @SimenB

lib/metrics/processResources.js
```diff
@@ -0,0 +1,58 @@
...
+		registers: registry ? [registry] : undefined,
+		collect() {
+			const resources = process.getActiveResourcesInfo();
+
+			const data = {};
+
+			for (let i = 0; i < resources.length; i++) {
+				const resource = resources[i];
+
+				if (Object.hasOwnProperty.call(data, resource)) {
```
zbjornson, line 32: Since this requires Node.js v17.3.0, you could use `Object.hasOwn(data, resource)` here.

zbjornson (changes requested):
Looks great, but the tests need to be skipped in Node.js <v17 for CI to pass.

It would be great to provide upgrade guidance in the changelog since the metric name (and maybe labels?) change. (Also noting that this PR is semver major for that reason.)

lib/metrics/processResources.js
RaisinTen: @zbjornson Thanks for the review! Updated accordingly, PTAL.

test/metrics/processResourcesTest.js
```diff
@@ -0,0 +1,38 @@
...
+		register.clear();
+	});
+
+	afterEach(() => {
+		register.clear();
+	});
+
+	it('should add metric to the registry', async () => {
+		if (typeof process.getActiveResourcesInfo !== 'function') {
+			it.skip('process.getActiveResourcesInfo() is not a function');
```
zbjornson, line 17: Oops, CI failed with:

```
Missing second argument. It must be a callback function. Perhaps you want to use `test.todo` for a test placeholder.

      15 | 	it('should add metric to the registry', async () => {
      16 | 		if (typeof process.getActiveResourcesInfo !== 'function') {
    > 17 | 			it.skip('process.getActiveResourcesInfo() is not a function');
         | 			   ^
      18 | 			return;
      19 | 		}
      20 | 
```

Can probably just delete this line.

RaisinTen: @zbjornson 🤦 thanks, deleted that. Hopefully we'll get a green CI this time!

zbjornson (approved):
Awesome, thank you!

zbjornson:
@RaisinTen - @cjihrig suggested that we actually keep both of these sets of metrics for now: https://twitter.com/cjihrig/status/1481433837255155717. What do you think?

RaisinTen:
@zbjornson yes, I agree that we should keep both the set of metrics for now. I sent https://github.com/siimon/prom-client/pull/488 to do so. :)

Issue #480, opened 2021-11-19
Performance Issues related with high request call rates to T _host_get_clock_service


sky-nunomarcos:
I have been troubleshooting a node implementation that is currently using your lib to expose prometheus metrics and using clinicJS tool notices that almost 80% of total cpu usage time is spend in T _host_get_clock_service call. Already updated lib version to 14.0.1, but issues continues to be present.

Clinic Flamechart is this one:
![imagem](https://user-images.githubusercontent.com/23661699/142622794-94564d68-95fc-4bb1-8904-a34e2656268b.png)

Looking at your source code it can be related with Date.now(); requests performed in lib/timeWindowQuantiles.js or  lib/metrics/processStartTime.js.

Do you have any suggestion on how to troubleshoot these issues and try to mitigate them?

Many Thanks for your help

zbjornson:
Thanks for reporting this. I'm hoping to look into this this month, but please don't let that stop anyone else from taking a look if they have time before I do.

timo-klarshift:
Any update on this? Due to the high resource usage of this library we cannot use it in production.

jdmarshall:
Other than the already mentioned issues and improvements coming in v0.16, my other concern is that somehow you're getting multiple copies of prom-client loaded and what you're actually seeing here is all of them gathering their defaultMetrics, rather than the one instance attached to the scraping logic.

PR #486, opened 2022-01-01
WIP: potential fix for #480

zbjornson:
@sky-nunomarcos can you give this branch a try and see if it resolves the performance issue you see?

Am I correct in assuming that you're using Summaries? Are you setting `maxAgeSeconds` and `ageBuckets`?

---

It sounds like `maxAgeSeconds` and `ageBuckets` *should* be set normally. They weren't set when they were added to prom-client to avoid a breaking change. I don't think the rotation interval has to be exact though, so we can batch rotations up at least.

### DIFF
--- a/lib/timeWindowQuantiles.js
+++ b/lib/timeWindowQuantiles.js
@@ -43,18 +43,17 @@ class TimeWindowQuantiles {
 }
 
 function rotate() {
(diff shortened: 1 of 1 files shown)


jdmarshall:
@zbjornson You should be able to benchmark this now, if it's still an issue.

jdmarshall:
@zbjornson I'd invite you to rebase this on main so the benchmarks run.

PR #487, opened 2022-01-11
fix: always use sum aggregator for sum & count

raghuram-periaswamy:
sum & count metrics of summary & histogram should always be summed irrespective of the custom aggregator that's mentioned as part of the metric definition.

Previously, aggregating the following two sample sets 

**set-1**
```
http_duration_seconds{quantile="0.95"} 4
http_duration_seconds_sum 10
http_duration_seconds_count 5
```

**set-2**
```
http_duration_seconds{quantile="0.95"} 6
http_duration_seconds_sum 20
http_duration_seconds_count 5
```

using `average` yields the following where every metric is `averaged`.
```
http_duration_seconds{quantile="0.95"} 5
http_duration_seconds_sum 15
http_duration_seconds_count 5
```

The fix will always perform `sum` for the `sum` & `count` yielding,
```
http_duration_seconds{quantile="0.95"} 5
http_duration_seconds_sum 30
http_duration_seconds_count 10
```

### DIFF
--- a/lib/metricAggregators.js
+++ b/lib/metricAggregators.js
@@ -3,6 +3,56 @@
 const { Grouper, hashObject } = require('./util');
 
 /**
+ * Returns the sum of all label values passed
+ * @param {Object} v Label values
+ * @return {Number} sum of values
+ */
+function sum(v) {
+	return v.reduce((p, c) => p + c.value, 0);
+}
+
+/**
+ * Returns the first label value
+ * @param {Object} v Label values
+ * @return {Number} first value.
+ */
+function first(v) {
+	return v[0].value;
+}
+
+/**
+ * @return {undefined} Undefined; omits the values.
+ */
+function omit() {}
+
+/**
+ * Returns the mean of all label values
+ * @param {Object} v Label values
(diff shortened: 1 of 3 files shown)


raghuram-periaswamy:
@SimenB @zbjornson Requesting for the review.

lib/metricAggregators.js
```diff
@@ -29,12 +79,20 @@ function AggregatorFactory(aggregatorFn) {
...
 			}
+			// Should always `sum` if it's `sum` or `count` that's part of histogram & summary
+			if (
+				values[0].metricName === `${result.name}_sum` ||
+				values[0].metricName === `${result.name}_count`
+			) {
+				valObj.value = sum(values);
+			} else {
+				valObj.value = aggregatorFn(values);
+			}
```
zbjornson, line 95: This is a good find, thanks and sorry for the year-long delay reviewing it.

I think this can wrongly identify custom metrics that are happen to be named `something_sum`/`something_count` though. Can you use the `type` instead? (`metrics[0].type`)

zbjornson (changes requested):

PR #488, opened 2022-01-16
Preserve the `processRequests` and `processHandles` metrics

RaisinTen:
The `processResources` metrics depend on the
`process.getActiveResourcesInfo()` API which is still experimental, so
we shouldn't remove the existing metrics in favour of that quite yet.

Signed-off-by: Darshan Sen <raisinten@gmail.com>

cc @zbjornson


zbjornson (approved):
Great, thank you!

PR #491, opened 2022-01-23
Update the sample output in `example/default-metrics.js`

RaisinTen:
This updates the outdated sample output with the latest one we get when we
run the script using Node.js v18.12.1 on an x86_64 Ubuntu 20.04.

Signed-off-by: Darshan Sen <raisinten@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- changed: updated the sample output in `example/default-metrics.js`
+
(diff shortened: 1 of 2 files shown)


RaisinTen:
cc @zbjornson

RaisinTen:
Gentle ping @zbjornson, in case you missed ^

SimenB:
Should we rerun with 18 as it's LTS?

RaisinTen:
@SimenB good idea, done!

SimenB (approved):
👍

Issue #494, opened 2022-02-25
EventLoop leak


justinmchase:
This appears to be a leak:

https://github.com/siimon/prom-client/blob/master/lib/metrics/eventLoopLag.js#L47-L50

You're not exposing a way to call `histogram.disable()` and so there is no way to turn it off and it will permanently hang open the process.

Ideally `collectDefaultMetrics` would return something that I could call stop on at a later time.

``` 
const register = new Registry()
const collect = collectDefaultMetrics({ register: this.register })

export function stop() {
  collect.stop()
}
```

I got this info from `why-is-node-running` module:

```
# ELDHISTOGRAM
node:internal/async_hooks:201                                                   
node:internal/perf/event_loop_delay:81                                          
node:internal/perf/event_loop_delay:78                                          
/var/app/example/node_modules/prom-client/lib/metrics/eventLoopLag.js:47 - const histogram = perf_hooks.monitorEventLoopDelay({
/var/app/example/node_modules/prom-client/lib/defaultMetrics.js:43       - metric(config.register, config);
```

SimenB:
`collectDefaultMetrics` should deffo return a function that tears down whatever it set up. Wanna send a PR?

justinmchase:
@SimenB I'll give it a go.

justinmchase:
That would be much better, I can't think of a time you'd want this to keep it open.

SimenB:
yeah, that makes sense!

zbjornson:
@justinmchase can you provide a full repro for this please? Neither of these are keeping Node.js (16.13.1) alive for me (see also https://github.com/nodejs/node/issues/42132#issuecomment-1053935605):

```js
const pc = require("."); // in the prom-client directory
const register = new pc.Registry()
const collect = pc.collectDefaultMetrics({ register })
```
```js
const perf_hooks = require('perf_hooks');
const histogram = perf_hooks.monitorEventLoopDelay({
	resolution: 10,
});
histogram.enable();
```

justinmchase:
Unfortunately I have a fairly complex setup of course and when I run `why-is-node-running` I have a number of things showing up in the list and this was just one of them, so I reported it here.

zbjornson:
[...]

PR #495, opened 2022-02-25
add the ability to stop the event loop

justinmchase:
Fixes https://github.com/siimon/prom-client/issues/494

Adds an optional "stop" function which when called will ultimately disable the histogram which will allow the node process to exit.

I also fixed a bug in the tests where a failed assertion in a setTimeout wasn't properly rejecting the test promise.

Screen Shot 2022-02-25 at 8 52 49 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,12 +27,16 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 - The `processResources` metric was added, which keeps a track of all sorts of
   active resources. It consists of the following gauges:
+
   - `nodejs_active_resources` - Number of active resources that are currently
     keeping the event loop alive, grouped by async resource type.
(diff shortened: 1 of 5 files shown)


SimenB:
>I also fixed a bug in the tests where a failed assertion in a setTimeout wasn't properly rejecting the test promise.

Since we're waiting for node to respond, maybe that could be pulled into a separate PR so we can land it? 🙂

justinmchase:
Merged with the other PR that broke out the change to the tests.

Also, you probably don't need to wait for the node "unref" response since, even if they agree and add it, you'll still probably want a way to stop this as its reasonable to want to stop it without stopping the process. Also, if they do accept it it will likely go into node 17, in which case I'll have to wait for node 18 (lts) which could be years frankly.

PR #496, opened 2022-02-26
chore(tests): async await instead of setTimeout

justinmchase:
added a util/wait function and modified a test to use that instead of setTimeout.

This change is extracted from PR #495

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
     async resources that these metrics do not keep a track of (like timers).
 
 - Added an async `wait` function which can be used in asynchronous methods
-  instead of setTimeout. Updated a test so failed expects would get
+  instead of `setTimeout`. Updated a test so failed expects would get
   propagated correctly.
 
 ## [14.0.0] - 2021-09-18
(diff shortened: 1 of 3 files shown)


lib/util.js
```diff
@@ -44,6 +44,10 @@ exports.getLabels = function (labelNames, args) {
 	return acc;
 };
 
+exports.wait = async function (ms) {
```
SimenB, line 47: I don't think we need to expose this out, it's enough to use it in the single test

lib/util.js
```diff
@@ -44,6 +44,10 @@ exports.getLabels = function (labelNames, args) {
...
+	return await new Promise(resolve => setTimeout(resolve, ms), ms);
```
SimenB, line 48:
```suggestion
	await new Promise(resolve => setTimeout(resolve, ms), ms);
```

SimenB (changes requested):

CHANGELOG.md
```diff
@@ -27,13 +27,18 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 - The `processResources` metric was added, which keeps a track of all sorts of
   active resources. It consists of the following gauges:
+
```
justinmchase, line 30: I didn't add this line, for some reason the commit hook is adding it.

SimenB (approved):
thanks!

CHANGELOG.md
SimenB: probably prettier thinking the newline is... prettier 😀

PR #503, opened 2022-06-15
perf: improve performance of gauge inc() dec()

tuyennhv:
**Motivation**
- Right now when using `gauge.inc()` or `gauge.dec()`, we call `hashObject()` 2 times
- It's better to call `hashObject()` once as in `counter.inc()`

**Description**
- New function `setValueDelta()` with a `hash` param
- `gauge.inc()` and `gauge.dec()` to call `setValueDelta()` via `setDelta()` function


SimenB (approved):
thanks!

PR #504, opened 2022-06-21
chore(deps): upgrade to latest dependencies

RomanYar:
Upgrading this project to the latest dependencies


SimenB:
We support node 10, so there's a bunch of these we cannot do without dropping a node version or 2

RomanYar:
No problem, it's just that Node dropped support for node 12, so it's pretty legit :) but no problem though
Thanks!

SimenB:
Done in #533

PR #505, opened 2022-06-21
feat(types): add default types to metrics types

RomanYar:
Currently all the `Metric` types must receive a generic argument which must extend a string.

Adding a default value `string` so that users who just want to type `<string>` there, won't have to


SimenB (approved):
thanks!

PR #508, opened 2022-08-21
Support gzipped pushgateway requests

vpalmisano:
This PR allows to use the gzip Content-Encoding introduced with https://github.com/prometheus/pushgateway/pull/477.

### DIFF
--- a/lib/pushgateway.js
+++ b/lib/pushgateway.js
@@ -3,6 +3,8 @@
 const url = require('url');
 const http = require('http');
 const https = require('https');
+const { Readable } = require('stream');
+const { createGzip } = require('zlib');
 const { globalRegistry } = require('./registry');
 
 class Pushgateway {
@@ -79,8 +81,23 @@ async function useGateway(method, job, groupings) {
 			this.registry
 				.metrics()
 				.then(metrics => {
-					req.write(metrics);
-					req.end();
+					if (
+						options.headers &&
+						options.headers['Content-Encoding'] === 'gzip'
+					) {
+						Readable.from([metrics])
+							.pipe(createGzip())
+							.pipe(req)
+							.on('error', error => {
+								reject(error);
+							})
+							.on('finish', () => {
+								req.end();
+							});
+					} else {
+						req.write(metrics);
+						req.end();
+					}
 				})
 				.catch(err => {
 					reject(err);
--- a/lib/pushgateway.js
+++ b/lib/pushgateway.js
@@ -3,8 +3,7 @@
 const url = require('url');
 const http = require('http');
 const https = require('https');
-const { Readable } = require('stream');
-const { createGzip } = require('zlib');
+const { gzipSync } = require('zlib');
 const { globalRegistry } = require('./registry');
 
 class Pushgateway {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
 		"index.d.ts"
 	],
 	"engines": {
-		"node": ">=10"
+		"node": ">=12.3.0"
```
SimenB, line 12: this is a breaking change - is it needed?

SimenB (approved):
thanks!

could you include a test or two?

package.json
vpalmisano: Yep, it is required for using Readable class.

vpalmisano: I'm checking if there is an alternative.

vpalmisano: This is the eslint error when using node=10: 
```
The 'stream.Readable.from' is not supported until Node.js 12.3.0 (backported: ^10.17.0). The configured version range is '>=10'.
```

SimenB:
An alternative might be using `readable-stream`.

However, dropping node 10 which has been EOL since April 2021 is probably fine. 12 is also EOL now, so probably fine to drop that as well

package.json
SimenB: An alternative might be using `readable-stream`.

However, dropping node 10 which has been EOL since April 2021 is probably fine. 12 is also EOL now, so probably fine to drop that as well

lib/pushgateway.js
```diff
@@ -79,8 +81,23 @@ async function useGateway(method, job, groupings) {
 			this.registry
 				.metrics()
 				.then(metrics => {
-					req.write(metrics);
-					req.end();
+					if (
+						options.headers &&
+						options.headers['Content-Encoding'] === 'gzip'
+					) {
+						Readable.from([metrics])
```
zbjornson, line 88: I don't think you need to use `Readable.from` here since it's just one string. That will be faster and avoid the node 12 requirement issue.

vpalmisano: What is the alternative? Using gzipSync will keep the output buffer in memory.

zbjornson: Just [`zlib.gzip(metrics, cb)`](https://nodejs.org/dist/latest-v16.x/docs/api/zlib.html#zlibgzipbuffer-options-callback), optionally put in a Promise if you want.

SimenB: we already have the entire string in memory, does passing it to `gzipSync` cause any real issue?

lib/pushgateway.js
```diff
@@ -79,8 +81,23 @@ async function useGateway(method, job, groupings) {
...
+							.pipe(createGzip())
+							.pipe(req)
+							.on('error', error => {
+								reject(error);
+							})
+							.on('finish', () => {
+								req.end();
+							});
```
SimenB, line 96: can we use `stream.pipeline` instead of manual `pipe` and event listening?

lib/pushgateway.js
vpalmisano: Probably it would be fine simply using `req.write(gzipSync(metrics))`.

SimenB:
Thanks! Can you update the changelog as well?

SimenB (approved):
👍

PR #515, opened 2022-08-31
Fix incompatibility with perf_hooks accross nodejs versions

Fadoli:
When using rollup to build an all-in-one esm module, rollup fails to properly handle require in try{}catch{} code.

It causes perf_hooks to not be defined ( that would happens if perf_hooks isn't implemented in current nodejs version) and breaks code accessing perf_hooks constants fields :

```js
perf_hooks.constants.NODE_PERFORMANCE_GC_MAJOR
```

It'd be nice to either require it (and as such require recent nodejs version) or make sure the codebase is compatible when it is not present.

This change is for enabling perf_hooks to not exists. It would also be possible to make it a hard requirement.

Note : `perf_hooks.constants` is defined only on nodejs version starting with the following :

1. >13.9.0
1. >12.17.0

Because of this I'm unsure the rest of the gc.js code will work correctly with nodejs >= 10 & <= 12.17

(Nodejs 10 requirement is from your package.json `engines` field)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- Increase compatibility with external build system such as `rollup` by making perf_hooks optional in gc.js
+
 ### Added
 
 ## [14.1.0] - 2022-08-23
--- a/lib/metrics/gc.js
+++ b/lib/metrics/gc.js
@@ -16,10 +16,10 @@ const DEFAULT_GC_DURATION_BUCKETS = [0.001, 0.01, 0.1, 1, 2, 5];
 const kinds = [];
 
 if (perf_hooks && perf_hooks.constants) {
-    kinds[perf_hooks.constants.NODE_PERFORMANCE_GC_MAJOR] = 'major';
-    kinds[perf_hooks.constants.NODE_PERFORMANCE_GC_MINOR] = 'minor';
-    kinds[perf_hooks.constants.NODE_PERFORMANCE_GC_INCREMENTAL] = 'incremental';
-    kinds[perf_hooks.constants.NODE_PERFORMANCE_GC_WEAKCB] = 'weakcb';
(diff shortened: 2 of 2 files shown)


SimenB (approved):
thanks 👍

SimenB:
We could also just make a major release dropping node 10, 12 and 17

mireiaangles1109:
Is there any update on this PR? We are facing the same issue and would be awesome to have it ready. 

Thanks!

zbjornson:
I agree that it's reasonable to just drop 10, 12 and 17. v12 went EOL in April. @mireiaangles1109 can you use a current version of Node.js, or are you still running into issues with v14+?

mireiaangles1109:
Thanks for your reply @zbjornson 
We are trying to upgrade vite to the latest version (4.0.0) that uses rollup3. This is causing perf_hooks to not be defined as explained in this PR. 

In our project we are using Node 16, so my guess is that the issues are coming with the rollup3 upgrade

Thanks!

felipesantiago5555:
It would be awesome to get this merged, we are also facing similar issue with rollup.

zbjornson:
It looks like Rollup has supported optional catch bindings since 2018. Do you need to set rollup's parser to the right ES version? https://github.com/rollup/rollup/pull/2455

gasparearmato6991:
We're having the same problem here using SvelteKit and Vite, it would great to merge this PR

zbjornson:
@gasparearmato6991 do you have Vite configured for the right target? I have no idea if it will work to use Vite with this Node.js-only library, either.

gasparearmato6991:
@zbjornson yes we're pointing the the right ES target, as mentioned before the problem here is that rollup does not handle the require in a try-catch block

SimenB:
I think we can land this now and just revert when it's non-conditional in a future major version

SimenB:
https://github.com/siimon/prom-client/releases/tag/v14.1.1

PR #522, opened 2022-11-11
Reset metrics when send to pushgateway

imsamurai:
Hi, currently metrics are not automatically reset when sent to pushgateway, instead they are accumulated. 
In this PR, I add resetting the metrics right before the http request (if we do this after the request is made, we lost all the metrics that was changed during the request)

### DIFF
--- a/lib/pushgateway.js
+++ b/lib/pushgateway.js
@@ -80,6 +80,7 @@ async function useGateway(method, job, groupings) {
 			this.registry
 				.metrics()
 				.then(metrics => {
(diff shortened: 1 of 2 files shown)


zbjornson:
Do you have a reference (or example in another language's prometheus library) showing that metrics should be reset? I don't think they're supposed to be, otherwise I'm not sure how counters would work.

imsamurai:
i'm sorry, I'm migrating from statsd and seems was not fully understand how prometheus works) You right, reset is not needed!

PR #530, opened 2023-01-12
feat: allow arrow functions for Gauge::collect

d0ctr:
As explained in the change for README.md, this will allow the usage of arrow functions as `Gauge::collect()`
```js
class Foo {
  constructor() {
    this.gauge_value = 0;
    this.gauge = new client.Gauge({
      name: 'metric_name',
      help: 'metric_help',
      collect: (metric) => {
        // Invoked when the registry collects its metrics' values.
        // This can be synchronous or it can return a promise/be an async function.
        metric.set(this.gauge_value);
      },
    });
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,8 +182,26 @@ new client.Gauge({
 });
 ```
 
-Note that you should not use arrow functions for `collect` because arrow
-functions will not have the correct value for `this`.
+If you want `this` to point to you object instead of the metric, you can pass arrow function as `collect`:
+
+```js
+const client = require('prom-client');
+class Foo {
+  constructor() {
+    this.gauge_value = 0;
+    this.gauge = new client.Gauge({
+      name: 'metric_name',
+      help: 'metric_help',
+      collect: (metric) => {
+        // Invoked when the registry collects its metrics' values.
+        // This can be synchronous or it can return a promise/be an async function.
+        metric.set(this.gauge_value);
+      },
+    });
+  }
+}
+
+```
 
 ##### Utility Functions
 
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ new client.Gauge({
 });
 ```
 
-If you want `this` to point to you object instead of the metric, you can pass arrow function as `collect`:
+If you want `this` to point to the outer scope instead of the metric, you can pass arrow function as `collect`:
 
 ```js
 const client = require('prom-client');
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -182,8 +182,26 @@ new client.Gauge({
...
+      collect: (metric) => {
+        // Invoked when the registry collects its metrics' values.
+        // This can be synchronous or it can return a promise/be an async function.
+        metric.set(this.gauge_value);
+      },
+    });
+  }
+}
+
+```
```
zbjornson, line 204: Thanks for the PR!

I'm not sure if this change is a good idea. That the readme says "functions will not have the correct value for `this`" is debatable: `this` is the metric, which is correct for some use cases.

In this example, wouldn't this work as well?
```js
collect() {
  this.gauge.set(this.gauge_value);
}
```

If we do make this change, it should be made to all metric types uniformly.

What do others think?

README.md
```diff
@@ -182,8 +182,26 @@ new client.Gauge({
 });
 ```
 
-Note that you should not use arrow functions for `collect` because arrow
-functions will not have the correct value for `this`.
+If you want `this` to point to you object instead of the metric, you can pass arrow function as `collect`:
```
zbjornson, line 185:
```suggestion
If you want `this` to point to the outer scope instead of the metric, you can pass arrow function as `collect`:
```

SimenB (commented):
I agree - if you want `this`, then don't use an arrow function.

If we _stopped_ using `this` and _just_ passed `metric` as first arg, I'd be more agreeable. But 2 ways of doing an identical thing seems more like bloat to me.

README.md
d0ctr: Your example wouldn’t work, `this` is reassigned when using Object.assign(), which is used here: https://github.com/siimon/prom-client/blob/8d6456c05f0708fed34628da5bd6f492a0ab43f8/lib/metric.js#L15

As for other metrics, my mistake, I was under the impression that the usage of the `collect` function is only available for the gauge.

zbjornson: Sorry -- my example should have used an arrow function:

```js
collect: () => {
  this.gauge.set(this.gauge_value);
}
```

d0ctr: This one would work, I suppose. Maybe it should be added to the README.md as an example?

PR #531, opened 2023-01-17
fix: add get method to type definitions of metric classes

nflaig:
**Current Limitation**

At the moment there is no clean way to get a single metric value programmatically while using typescript.

The following is not possible as it causes type issues as the getters are implemented in javascript but are missing from type definitions.

```ts
const metric = metrics.register.getSingleMetric("process_cpu_user_seconds_total");

// Property 'get' does not exist on type 'Metric<string>'
const {value} = (await metric.get()).values[0]
```

**Solution**

Add `get` method to type definitions of each metric classes and define `MetricObjectWithValues` type which is returned by getters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Add `get` method to type definitions of metric classes
+
 ## [14.1.1] - 2022-12-31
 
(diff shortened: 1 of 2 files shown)


nflaig:
@SimenB sorry for the ping but is there anything I can do to get this merged and released? 🙏

SimenB:
Thanks!

PR #541, opened 2023-02-17
docs(README): remove await of getSingleMetric

ric113:
### Summary

`getSingleMetric` is a synchronous function, don't need to `await` it 😅 
ref: https://github.com/siimon/prom-client/blob/master/lib/registry.js#L132


zbjornson (approved):
Thanks!

PR #542, opened 2023-02-26
feat: improve performance of getMetricAsPrometheusString

shappir:
Use Array.prototype.join instead string concatenation for constructing Prometheus metric string. This results in improved performance and reduced memory consumption, especially when there's a large amount of dimensions (labels and label values). In addition uses array iteration methods instead of explicit loops.

### DIFF
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -27,40 +27,27 @@ class Registry {
 		const name = escapeString(item.name);
 		const help = `# HELP ${name} ${escapeString(item.help)}`;
 		const type = `# TYPE ${name} ${item.type}`;
-		const defaultLabelNames = Object.keys(this._defaultLabels);
+		const defaultLabels =
+			Object.keys(this._defaultLabels).length > 0 ? this._defaultLabels : null;
 
-		let values = '';
-		for (const val of item.values || []) {
-			val.labels = val.labels || {};
+		const values = [help, type];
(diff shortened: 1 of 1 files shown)


SimenB:
Thanks! Do you have any benchmarks before/after?

https://github.com/siimon/prom-client/tree/master/benchmarks

shappir:
This change primarily impacts the performance of metrics that have a large number of rows in the Prometheus metrics output, e.g. histograms with several labels. In our application I saw metric generation time reduced by ~50%. This is not surprising since the number of times string values are copied as result of concatination is the same order as the number of rows.

In the benchmark test I got 50% faster and 50% "acceptably slower". Every once in a while I would get one entry or another as slower, but it was inconsistent.

SimenB:
Hey @shappir! I just released this in 14.2.0.

However, I'm having some issues trying to rebase `next` on `master` on top of #544. You wouldn't by any chance be able to send a forward-porting PR to `next` applying the changes in this PR?

shappir:
Is this still an issue? Do you need me to do anything?

SimenB:
If it's not too much trouble, a PR against `next` applying the same optimizations would be very helpful 🙂

shappir:
Created a pr for `next`: https://github.com/siimon/prom-client/pull/548

Issue #543, opened 2023-03-01
Very long NodeJS event loop lag


shappir:
We are using `prom-client` with NestJS (together with `@willsoto/nestjs-prometheus`). We have several custom metrics - in particular histograms. As a result, the overall metrics size grows up to 8MB. This results in the NodeJS event loop lag getting to be as large as 1 second!

What we see:
- Reducing the amount of data collected, e.g. by removing some metrics, improves the situation (however we would like to keep the metrics we have)
- Explicitly invoking `register.resetMetrics()` causes the lag to go down almost to 0. It then starts increasing again
- CPU usage is constant at around 15%
- CPU profile shows long time spent in Node internal network functions, such as `getPeerCertificate` and `destroySSL`

Has anyone encountered this sort of behavior? Any suggestions?

ngavalas:
I've noticed this as well; there are long, blocking calls when converting the metrics to the string that gets output to the client. I saw you landed a PR that improved this a bit (thanks); I'm benchmarking another fix on top of yours to see if it helps.

Would love an update if you figure this out.

zbjornson:
Is something inflating the number of metrics, e.g. using request paths as label values? What types of metrics are you using?

14.2.0 has https://github.com/siimon/prom-client/pull/540 that might help if you're using summaries.

#480 is another known perf issue with summaries, and I started a fix in #486.

ngavalas:
@zbjornson From my traces, the other major offender when we have a slow call is rendering histograms. In particular, this function seems quite slow: https://github.com/siimon/prom-client/blob/master/lib/histogram.js#L210-L217

I believe we could improve the performance of this by internally having a concept of "shared labels" across an entire set of buckets so that we don't need to copy all the labels N times. In particular, `histogram.get()` would return this set of shared labels and we'd merge in the specific entries labels on top of that.

For histograms, this means we only build `{ le: upperBound }` per bucket and everything else is shared. I would need to go poke around in the code a bit and see what else this would affect, but it seems like a relatively straightforward win.

nflaig:
@ngavalas  `<metric>.get()` is an external API and can be consumed by anyone who want to get a single metric value

ngavalas:
Ah. Thoughts on exposing an internal version for `getMetricAsPrometheusString` and have `get` flatten out the labels for external callers so the existing API is undisturbed?

zbjornson:
[...]

PR #546, opened 2023-03-08
perf: swap order of escape regexes to avoid lookahead

ngavalas:
When looking at some CPU profiles of metric -> string serialization, I noticed that escaping all of the strings and labels was showing up in the flamegraph more than I would expect. This isn't a massive win by any means, but it's pretty consistently ~20% faster on a microbenchmark: https://jsbench.me/qklez5ombn/1. I'll take 20%.

Some notes:
1) The benchmarks literally don't test this case, as far as I can tell. None of the registry benchmarks have quotes or newlines or anything in the labels? An A/A test (master vs. 14.2.0) was producing inconsistent results, so I'm not quite sure I trust that benchmark anyway.
2) Correctness-wise, this should work. Replacing the `\` before replacing the `\n` fixes the case that made you need the `.replace(/\\(?!n)/g, '\\\\')` regex in the first place, so now we can just do `\\` -> `\\\\` and `\n` -> `\\n` and `"` -> `\\"` in that order without lookaheads.

Relevant to #543.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Breaking
 
 ### Changed
+- Refactor `escapeString` helper in `lib/registry.js` to improve performance and
+  avoid an unnecessarily complex regex.
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Breaking
 
 ### Changed
+
 - Refactor `escapeString` helper in `lib/registry.js` to improve performance and
   avoid an unnecessarily complex regex.
 
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -2,7 +2,7 @@
 const { getValueAsString } = require('./util');
 
 function escapeString(str) {
-	return str.replace(/\n/g, '\\n').replace(/\\(?!n)/g, '\\\\');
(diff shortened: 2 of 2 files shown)


zbjornson:
Nice. prom-client still support Node.js 10 though, and `String.prototype.replaceAll` wasn't added until Node.js 15.0.0. I think the other maintainers are fine with bumping the minimum supported version, but we should maybe wait until [Apr 30 when Node.js v14.x is EOL](https://github.com/nodejs/release#release-schedule).

Did you benchmark just reversing the two regexps so that a lookbehind isn't needed? I'm curious if that would give a similar speedup.

ngavalas:
Totally forgot that we used to live in a world without `replaceAll`. Dark times.

From the microbenchmark, it seems like the swapped regex is as fast, if not faster (!!!). Somewhat surprising given that `replaceAll` is like the simplest case of regex matching, so I'd expect it to be as fast as the equivalent regex, but...

https://jsbench.me/qklez5ombn/1

I'm happy to just swap the regex order and call it a day given that it seems to be the winner anyway. The thing that really matters is that these operations (original regex vs. swapped w/o lookahead) are equivalent; am I missing anything?

ngavalas:
I'll update the changelog after all the checks come back green to avoid an extra roundtrip if something's still wrong. My bad. :)

zbjornson:
Awesome, easy win.

Yes, the behavior from just swapping the regexps looks correct.

zbjornson (approved):
Thanks!

PR #548, opened 2023-03-08
perf: improve performance of getMetricAsString in next

shappir:
Use `Array.prototype.join` instead string concatenation for constructing Prometheus metric string. This results in improved performance and reduced memory consumption, especially when there's a large amount of dimensions (labels and label values). In addition uses array iteration methods instead of explicit loops.

**note**: there are no tests for exemplar so I could not verify that this functionality works as expected.

### DIFF
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -2,22 +2,6 @@
 
 const { getValueAsString } = require('./util');
 
-function escapeString(str) {
-	return str.replace(/\n/g, '\\n').replace(/\\(?!n)/g, '\\\\');
-}
-function escapeLabelValue(str) {
-	if (typeof str !== 'string') {
-		return str;
-	}
-	return escapeString(str).replace(/"/g, '\\"');
-}
-function standardizeCounterName(name) {
-	if (name.endsWith('_total')) {
-		return name.replace('_total', '');
-	}
-	return name;
-}
-
 class Registry {
 	static get PROMETHEUS_CONTENT_TYPE() {
 		return 'text/plain; version=0.0.4; charset=utf-8';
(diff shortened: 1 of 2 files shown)


SimenB:
Wonderful, thank you!

lib/registry.js
```diff
@@ -250,5 +201,24 @@ class Registry {
...
+}
+
+function escapeLabelValue(str) {
+	if (typeof str !== 'string') {
+		return str;
+	}
+	return escapeString(str).replace(/"/g, '\\"');
+}
+function escapeString(str) {
+	return str.replace(/\n/g, '\\n').replace(/\\(?!n)/g, '\\\\');
```
SimenB, line 217: #546 as well?

zbjornson:
@SimenB I didn't realize you had a "next" branch. Should I be landing all PRs against that? Is that just to keep the README aligned with the current npm release?

I can move the commits I landed from master to next tonight.

SimenB:
@zbjornson yeah, landed removal of old nodes and such. I think landing changes there and just releasing a new major would be nice

zbjornson:
@SimenB got it, I will fix up the mess I made and do that going forward.

SimenB (approved):

SimenB:
Thanks again @shappir!

---

@zbjornson I'll just land `next` on `master`, then the issue should resolve itself

SimenB:
@budarin happy to take a PR to that effect 🙂

budarin:
[preset-perf](https://github.com/budarin/babel-preset-perf) 🙂

SimenB:
I don't think we wanna add a compilation step at this time. Cool plugin, tho!

PR #551, opened 2023-03-09
fix: Disable content encoding for pushgateway delete requests

vpalmisano:
When using gzip `Content-Encoding` the DELETE requests sent to a pushgateway server fail with error code 400. With this PR we disable `Content-Encoding` option, if found, for DELETE requests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


SimenB:
Can you add a changelog entry?

zbjornson (approved):
I think a compliant server is supposed to ignore this and anything else related to a body, but we can definitely remove it here. 👍 

Looks good % changelog entry.

PR #553, opened 2023-03-11
bug: support array argument to labels() again

### LINKED ISSUE #499: Breaking changes v13 - labels as arguments considered as object
Hi,
When I'm setting labels as arguments and the first argument is an array:
```
someCounter.labels(arr, string1, string2)
```
it is considered as an object and throws `Added label "0" is not included in initial labelset`.
I found that there is a [condition](https://github.com/siimon/prom-client/blob/721829cc593bb7da28ae009985caeeacb4b59e05/lib/util.js#L32) `if (typeof args[0] === 'object')` but it might also check that it is not an array.

I've upgraded from v12 to v14 (but it happens in v13 also).

Thanks

zbjornson:
This was undocumented and untested, and broke in 13.1.0.

<sub>I'm not sure why we provide 4 ways to set labels.</sub>

Fixes #499

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
   avoid failures from the server when using `Content-Encoding: gzip` header.
 - Refactor `escapeString` helper in `lib/registry.js` to improve performance and
   avoid an unnecessarily complex regex.
+- Support array argument to `metric.labels()`. This was supported prior to
+  version 13.1 (undocumented). [#553](https://github.com/siimon/prom-client/pull/553)
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


zbjornson:
> I'm not sure why we provide 4 ways to set labels

@siimon @SimenB I'm somewhat inclined to not land this. This was undocumented and untyped, and it took 14 months for one user to notice this broke (between the time 13.1 was released and issue was opened). On the other hand, it's easy to support and the work is done. WDYT?

SimenB:
I'm happy with reducing the API surface

jdmarshall:
getLabels isn't the hottest path in the code but it's kinda hot.

If I'm reading this correctly, the index.d.ts signatures already exist in master now. Another variation on this is:

```
  args[0].constructor === Object
```


This has been sitting this way for a long time now, do we still care?

PR #555, opened 2023-03-12
perf: improve perf by caching unchanged histogram strings between gets

shappir:
`Histogram.get()` returns the content of the histogram as JSON that gets converted into a string representation by `Registry.getMetricsAsString()`. This conversion can be expensive, especially when there are many labels / label values / buckets, because each combination needs to be converted. This overhead can be reduced by caching the string representations for each entry in the histogram hash. This is because when there are many combinations, it's likely that most won't change between getting the metrics.

In order to perform the caching, the conversion to string needs to happen inside `Histogram.get()`. To accomplish this `Registry.getMetricsAsString()` will now pass into it an optional value-to-string conversion function. If such a function is provided then the caching will happen, otherwise the same behavior as before.

It's possible to easily extend this behavior to the other metrics types. However they won't get such a significant improvement as the conversion operation for them is less expensive per value.

**note**: registry benchmark has been modified to perform an update of an eighth if a histogram between getting the metrics string. This is so that it's not all cached. This is because full caching is so fast the benchmark explodes.

### DIFF
--- a/benchmarks/registry.js
+++ b/benchmarks/registry.js
@@ -16,15 +16,23 @@ function setupRegistrySuite(suite) {
 	labelSetups.forEach(({ name, counts }) => {
 		suite.add(
 			`getMetricsAsJSON#${name}`,
-			(client, registry) => registry.getMetricsAsJSON(),
-			{ setup: setup(counts) },
+			(client, { registry }) => registry.getMetricsAsJSON(),
+			{ teardown, setup: setup(counts) },
 		);
 	});
 
 	labelSetups.forEach(({ name, counts }) => {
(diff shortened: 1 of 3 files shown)


SimenB:
This conflicts with #549 - but the optimisations here might've been applied there?

shappir:
Is that what you prefer? That I apply my optimizations on top of that PR? Or to wait until it's merged and then review/merge this PR separately?

Also, can you review and provide feedback on this in any event, even if you don't want to merge yet?

SimenB:
#549 has landed on master, which is why this PR has conflicts. If the optimisations in this PR are still applicable, I'd love to land them 🙂

shappir:
Checking to determine if still applicable. In the interim closing this PR

PR #559, opened 2023-03-14
perf: improve perf by refactoring code to be cleaner and more efficient

shappir:
Cleans up and compacts the code.

### DIFF
--- a/lib/histogram.js
+++ b/lib/histogram.js
@@ -56,8 +56,8 @@ class Histogram extends Metric {
 			this.hashMap = {
 				[hashObject({})]: createBaseValues(
 					{},
-					Object.assign({}, this.bucketValues),
-					Object.assign({}, this.bucketExemplars),
+					this.bucketValues,
+					this.bucketExemplars,
 				),
 			};
 		}
@@ -84,16 +84,17 @@ class Histogram extends Metric {
 
 	updateExemplar(labels, value, exemplarLabels) {
 		const hash = hashObject(labels);
-		const b = findBound(this.upperBounds, value);
-		if (!isObject(this.hashMap[hash].bucketExemplars[b])) {
-			this.hashMap[hash].bucketExemplars[b] = new Exemplar();
+		const bound = findBound(this.upperBounds, value);
+		const { bucketExemplars } = this.hashMap[hash];
+		let exemplar = bucketExemplars[bound];
+		if (!isObject(exemplar)) {
+			exemplar = new Exemplar();
+			bucketExemplars[bound] = exemplar;
 		}
-		this.hashMap[hash].bucketExemplars[b].validateExemplarLabelSet(
-			exemplarLabels,
-		);
-		this.hashMap[hash].bucketExemplars[b].labelSet = exemplarLabels;
-		this.hashMap[hash].bucketExemplars[b].value = value;
-		this.hashMap[hash].bucketExemplars[b].timestamp = nowTimestamp();
+		exemplar.validateExemplarLabelSet(exemplarLabels);
+		exemplar.labelSet = exemplarLabels;
+		exemplar.value = value;
+		exemplar.timestamp = nowTimestamp();
 	}
 
 	async get() {
@@ -134,8 +135,8 @@ class Histogram extends Metric {
 		const hash = hashObject(labels);
 		this.hashMap[hash] = createBaseValues(
 			labels,
-			Object.assign({}, this.bucketValues),
-			Object.assign({}, this.bucketExemplars),
+			this.bucketValues,
+			this.bucketExemplars,
 		);
 	}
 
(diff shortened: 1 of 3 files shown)


lib/registry.js
```diff
@@ -212,10 +205,10 @@ class Registry {
...
 }
 
-function formatLabels(labels) {
-	return Object.entries(labels).map(
-		([n, v]) => `${n}="${escapeLabelValue(v)}"`,
-	);
+function formatLabels(labels, filter) {
+	const entries = Object.entries(labels);
+	const filtered = filter ? entries.filter(filter) : entries;
+	return filtered.map(([n, v]) => `${n}="${escapeLabelValue(v)}"`);
```
SimenB, line 211: do we need to iterate the labels 3 times?

shappir: Do you mean why `formatLabels` is called from three places? This is because:
1. for the shared labels, to generate the formatted string that is also saved in the map (line 221)
2. for the labels that aren't shared (line 61). This one includes a filter to prevent the shared labels to be included twice
3. for `exemplar.labelSet` (line 75)
All these iterations existed before. I consolidated them to use the same function instead of repeating the code.

zbjornson: Simen was maybe asking--or if not, I'll ask--about the three iterations from `Object.entries`, `filter` and `map`. Might get some speed boost by using one plain loop.

zbjornson (commented):
Benchmarks so far:

```
✓ registry ➭ getMetricsAsJSON#1 with 64 is 68.92% faster.
✓ registry ➭ getMetricsAsJSON#2 with 8 is 110.6% faster.
✓ registry ➭ getMetricsAsJSON#2 with 4 and 2 with 2 is 72.82% faster.
✓ registry ➭ getMetricsAsJSON#2 with 2 and 2 with 4 is 481.3% faster.
✓ registry ➭ getMetricsAsJSON#6 with 2 is 127.6% faster.
✓ registry ➭ metrics#1 with 64 is 23.72% faster.
✓ registry ➭ metrics#2 with 8 is 53.88% faster.
✓ registry ➭ metrics#2 with 4 and 2 with 2 is 72.04% faster.
✓ registry ➭ metrics#2 with 2 and 2 with 4 is 71.24% faster.
✓ registry ➭ metrics#6 with 2 is 98.86% faster.
✓ histogram ➭ observe#1 with 64 is 1.422% faster.
✓ histogram ➭ observe#2 with 8 is 2.263% faster.
✓ histogram ➭ observe#2 with 4 and 2 with 2 is 5.482% faster.
⚠ histogram ➭ observe#2 with 2 and 2 with 4 is 1.523% acceptably slower.
✓ histogram ➭ observe#6 with 2 is 0.7490% faster.
✓ gauge ➭ inc is 11.81% faster.
✓ gauge ➭ inc with labels is 3.459% faster.
⚠ summary ➭ observe#1 with 64 is 3.768% acceptably slower.
⚠ summary ➭ observe#2 with 8 is 1.771% acceptably slower.
✓ summary ➭ observe#2 with 4 and 2 with 2 is 3.571% faster.
✓ summary ➭ observe#2 with 2 and 2 with 4 is 4.680% faster.
✓ summary ➭ observe#6 with 2 is 0.01691% faster.
```

prom-client benchmarks are super noisy... I don't think there's any reason gauge's perf should have changed, so these numbers are probably +/- at least 11.8%.

lib/registry.js
shappir: I refactored to use a single explicit loop (`for...of`) and it is indeed noticeably faster so I pushed this change.

lib/histogram.js
```diff
@@ -239,43 +240,39 @@ function observe(labels) {
...
+		  };
 }
 
 function extractBucketValuesForExport(histogram) {
+	const name = `${histogram.name}_bucket`;
 	return bucketData => {
-		const buckets = [];
 		let acc = 0;
-		for (const upperBound of histogram.upperBounds) {
+		const buckets = histogram.upperBounds.map(upperBound => {
```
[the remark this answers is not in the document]
shappir: See my previous comment: I checked changing the implementation of `formatLabels` to an explicit loop using `for...of` and it's indeed faster so I pushed this change. In the other cases `map` was either slightly faster for me, or the there was no noticeable difference, and so I took the shorter and more explicit implementation.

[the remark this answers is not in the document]
shappir: > Usually plain loops + `arr.push()` are faster than `map()`

I dispute this as a general statement: with `map` JS engine can allocated a properly sized array to begin with. OTOH when using `push` the engine may need to expand the size of the target array. Each such expansion means new allocation + copying over the values from the previous target array instance.

shappir:
Any more thoughts / comments about this PR? Any chance that it will be merged?

SimenB (approved):
thanks!

Issue #565, opened 2023-05-06
ContentTypes in index.d.ts are mixed up


FreeRiderBysik:
export type **Prometheus**ContentType =
	`${OpenMetricsMIME}; version=${OpenMetricsVersion}; charset=${Charset}`;

export type **OpenMetrics**ContentType =
	`${PrometheusMIME}; version=${PrometheusMetricsVersion}; charset=${Charset}`;

PR #571, opened 2023-05-13
Fix typescript definitions for using with Registry<OpenMetricsContentType>

FreeRiderBysik:
1. Content types was mixed up -  #565 
2. Methods like collectDefaultMetrics without setting generic type does not allows to work with Registry<OpenMetricsContentType>


SimenB (approved):
thanks!

PR #574, opened 2023-06-06
fix: pushgateway reject with timeout

FLNacif:
As explained in [README.md](https://github.com/siimon/prom-client/blob/master/README.md?plain=1#L575) the timeout option should timeout the request but it is not doing it.

It was missing listen to the timeout event from `node HTTP` and rejecting the promise (destroying the request).


SimenB (approved):
thanks!

PR #578, opened 2023-07-17
fix: do not ignore error if request fails

kdeberk:
Throw an error if the gateway rejected the pushed metrics.

### DIFF
--- a/lib/pushgateway.js
+++ b/lib/pushgateway.js
@@ -76,9 +76,9 @@ async function useGateway(method, job, groupings) {
 					reject(
 						new Error(`push failed with status ${resp.statusCode}, ${body}`),
 					);
+				} else {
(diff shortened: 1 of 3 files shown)


lib/pushgateway.js
```diff
@@ -72,6 +72,12 @@ async function useGateway(method, job, groupings) {
 				body += chunk;
 			});
 			resp.on('end', () => {
+				if (resp.statusCode >= 400) {
+					reject(
+						new Error(`push failed with status ${resp.statusCode}, ${body}`),
+					);
+				}
+
 				resolve({ resp, body });
```
SimenB, line 81:
```suggestion
				} else {
					resolve({ resp, body });
				}
```

kdeberk: I've added your suggestion

PR #585, opened 2023-08-31
Fixes description for setToCurrentTime to seconds instead of milliseconds

### LINKED ISSUE #568: `Gauge.setToCurrentTime()` : TypeScript definition says the time is set in milliseconds, but implementation sets the time in epoch seconds
TypeScript doc says it sets the value to epoch milliseconds:

https://github.com/siimon/prom-client/blob/1df34ed58e83914244e8e6c99edc1e086697af5a/index.d.ts#L361-L365

Actual implementation sets the value to epoch seconds:

https://github.com/siimon/prom-client/blob/1df34ed58e83914244e8e6c99edc1e086697af5a/lib/gauge.js#L74-L86

I guess it was not updated properly after changed to seconds (#167)

alarobric:
Fixes #568 
This was broken since the change in #167 in v11.0.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
   avoid an unnecessarily complex regex.
 - Cleanup code and refactor to be more efficient
 - Correct TS types for working with OpenMetrics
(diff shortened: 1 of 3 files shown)


SimenB:
Thanks! Could you update the changelog as well?

alarobric:
Sure, done now.

SimenB (approved):
thanks!

PR #586, opened 2023-09-07
chore: use caching of actions/setup-node

SimenB:
No need to use `actions/cache`


SimenB:
Ah, that action only supports caching when there's a lockfile present. Fair enough 😀

PR #592, opened 2023-10-14
perf: refactored structureOutput with single reduce

Connormiha:
A little simplification with single `reduce` instead of `filter`+`forEach`. And `startsWith` instead of `indexOf`.


zbjornson:
Looks great, thanks. CI will complain in a moment that there's no changelog entry. Can you add one please?

SimenB (approved):
thanks!

SimenB:
https://github.com/siimon/prom-client/releases/tag/v15.1.0

PR #593, opened 2023-10-24
feat: startTimer with exemplar

Pigrabbit:
When using `Histogram.startTimer()` with 'enableExemplars' option set to true in order to record duration, it throws an error as following.
It supposed to work normally as it does when 'enableExemplars' option set to false.

image

Plus, it would be better if we could provide adding exemplarLabels just like we could do with `Histogram.observe()`.

I have added the feature with regarding test cases. Please feel free to give me any feedback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Allow Pushgateway to now require job names for compatibility with Gravel Gateway.
+- Allow `histogram.startTime()` to be used with exemplars.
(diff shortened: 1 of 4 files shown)


zbjornson (approved):
Thanks for the PR, this looks perfect!

I just rebased (and added a changelog entry along the way) to try to get CI to run. I'm missing the button to authorize it for some reason, but all the checks pass locally.

zbjornson:
Heh, there it goes.

SimenB:
https://github.com/siimon/prom-client/releases/tag/v15.1.0

PR #600, opened 2023-11-26
add `defaultLabels` option to metric classes

### LINKED ISSUE #598: Set default labels on a single metric
I would like to propose an option to be added to each metric type, which would allow specifying default label values for that metric only. Those labels would then be applied to each metric value (unless the value is overridden when recorded, similar to how default labels already work at the client level in `prom-client`).

It would look something like this:
```
const counter = new client.Counter({
  name: 'metric_name',
  help: 'metric_help',
  labelNames: ['my_label'],
  defaultLabels: {
    my_label: 'my_value'
  }
});

counter.inc(); // recorded with my_label: my_value
counter.inc({ my_label: 'my_other_value' }); // recorded with my_label: my_other_value
```

If this is a valid use case / not already possible, I would love to contribute.

**More use case details**:
My team uses prom-client to expose metrics which are scraped by prometheus and ultimately pushed to newrelic via a [prometheus remote write](https://prometheus.io/docs/concepts/remote_write_spec/) integration.

We have an issue occasionally where if a metric is created that does not follow the [naming convention expected by newrelic](https://docs.newrelic.com/docs/infrastructure/prometheus-integrations/install-configure-remote-write/set-your-prometheus-remote-write-integration/#mapping), newrelic interprets the metric as an inappropriate type (most often, counters are interpreted as gauges). I'd like to use this API to add the `newrelic_metric_type` label in a universal way to prevent this.

bienzaaron:
adds a `defaultLabels` option to the base Metric class and updates each metric class to use the value where appropriate.

Closes #598

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -10,7 +10,7 @@
 		"es6": true
 	},
 	"parserOptions": {
-		"ecmaVersion": 2019
+		"ecmaVersion": 2020
 	},
 	"rules": {
 		"no-underscore-dangle": "off",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 - Allow Pushgateway to now require job names for compatibility with Gravel Gateway.
 - Allow `histogram.startTime()` to be used with exemplars.
+- added `defaultLabels` option to metric classes
 
 ## [15.0.0] - 2023-10-09
 
--- a/README.md
+++ b/README.md
@@ -393,6 +393,53 @@ Default labels will be overridden if there is a name conflict.
 
 `register.clear()` will clear default labels.
 
+##### Per-metric default label values
+
+Static labels values may also be applied per-metric:
+
+```js
+const counter = new client.Counter({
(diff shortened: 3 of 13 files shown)


BourgoisMickael:
Instead of applying it during metric usage, maybe it could be added at metric scrape time https://github.com/siimon/prom-client/blob/c1d76c5d497ef803f6bd90c56c713c3fa811c3e0/lib/registry.js#L55-L57

jdmarshall:
There is default label support in the registry, which you can add each metric to.

https://github.com/siimon/prom-client/blob/4c6421b2253c9780a845889004b9a58dac646a1c/lib/registry.js#L17

Given the number of open tickets about memory and CPU pressure, and how much of that is down to the fact of labels, I think it would probably be good to use what is already there instead of adding more ways to do the slow things.

jdmarshall:
A lot of very difficult work has been done in the last year to make the cost of calling `inc()` competitive with other stats frameworks. Adding bookkeeping on every single call is not a good solution. Add your defaults to a registry, so they are only taxed at scrape time. Or move them to your sidecar if at all possible, where they will not be run inside of a JIT with a global interpreter lock

PR #606, opened 2024-01-11
perf: improve the memory usage of histogram

xsbchen:
We have a use case where it generates a huge amount of histograms, each has more than 10 labels, we found the memory usage is quite high, under this scenario, we did an experiment and found that removing bucketExemplars field from histograms could save about `27%` memory usage for each histogram instance, hope the maintainers could review and consider it
SeaTalk_IMG_20240111_195229

### DIFF
--- a/lib/histogram.js
+++ b/lib/histogram.js
@@ -26,13 +26,6 @@ class Histogram extends Metric {
 		this.defaultExemplarLabelSet = {};
 		this.enableExemplars = false;
 
-		if (config.enableExemplars) {
-			this.enableExemplars = true;
-			this.observe = this.observeWithExemplar;
-		} else {
-			this.observe = this.observeWithoutExemplar;
-		}
-
 		for (const label of this.labelNames) {
 			if (label === 'le') {
 				throw new Error('le is a reserved label keyword');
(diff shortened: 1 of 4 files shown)


SimenB:
Should we do this for the other metric types as well?

xsbchen:
> Should we do this for the other metric types as well?

I checked only `Counter` and `Histogram` support the `enableExemplars` option, and `Counter` already handled this case.

SimenB:
@voltbit thoughts on this? Seems reasonable to me, but I have to admit I'm a bit unsure how this code works 😅

lib/histogram.js
```diff
@@ -45,13 +38,19 @@ class Histogram extends Metric {
 			return acc;
 		}, {});
 
-		this.bucketExemplars = this.upperBounds.reduce((acc, upperBound) => {
-			acc[upperBound] = null;
-			return acc;
-		}, {});
+		if (config.enableExemplars) {
+			this.enableExemplars = true;
```
SimenB, line 42: should we set this to `false` in the `else`? I know it was not done before, but still

voltbit:
Looks good to me. The data struct being there when exemplars are disabled was an overlook on my side I think.

lib/histogram.js
xsbchen: hi SimenB, `this.enableExemplars` already set to `false` in [line 27](https://github.com/siimon/prom-client/pull/606/files/5b58cd4f34ac3f2518a1f6daa5d861b883107329#diff-4cacad18c9a9f40e8b66ee6f58741215e06d9e1bcd886452f61e74daee8ea387R27)

xsbchen:
@zbjornson @siimon master, plz help to review it if you are free, thanks.

SimenB:
https://github.com/siimon/prom-client/releases/tag/v15.1.1

PR #608, opened 2024-01-29
Implement BigInt support for gauges and counters

hudikm:
Add BigInt support for gauges and counters.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -1,7 +1,9 @@
 'use strict';
 
 exports.getValueAsString = function getValueString(value) {
-	if (Number.isNaN(value)) {
+	if (typeof value === 'bigint') {
+		return `${value}`;
+	} else if (Number.isNaN(value)) {
 		return 'Nan';
 	} else if (!Number.isFinite(value)) {
 		if (value < 0) {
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -53,7 +53,8 @@
 	"types": "./index.d.ts",
 	"jest": {
 		"testEnvironment": "node",
-		"testRegex": ".*Test\\.js$"
+		"testRegex": ".*Test\\.js$",
+		"maxWorkers": 1
```
SimenB, line 57: Revert this

lib/util.js
```diff
@@ -1,7 +1,9 @@
 'use strict';
 
 exports.getValueAsString = function getValueString(value) {
-	if (Number.isNaN(value)) {
+	if (typeof value === 'bigint') {
+		return `${value}`;
```
SimenB, line 5: Is the `n` suffix valid for both openmetrics and prom?

SimenB (changes requested):

lib/util.js
hudikm: In this case, the string representation of BigInt does not include the n suffix.

package.json
hudikm: Reverted to the previous version

hudikm:
BigInt is not supported in the Prometheus server itself, so it doesn't make sense to add support to the prom client library.

PR #613, opened 2024-02-07
feat: enhance error output on not valid name and label names

ManuelLG92:
While working I was facing an error performing some testing because of no valid metric name and no valid label names. 
I decided to dive into the code to get more output in the error.

The changes are the following:

- Show the invalid name in the error. 
  - Original:
  `Invalid metric name`
  - Improvement:
  `Invalid metric name: ${this.name}`
- Show the invalid label names instead of a generic message
  - Change the check of every to filter by the regex, and return the `invalid` labels. So it returns an array instead of a boolean
  - Check if the array length is higher than 0, then the error is thrown.
    - Original:
     `throw new Error('Invalid label name');`
    - Improvement:
    `Invalid metric name: ${this.name}`
    
  The changes have been tested by unit testing inside the gauge test case. It could be in any other metric that inherits from the[ metric abstract class](https://github.com/siimon/prom-client/blob/master/lib/metric.js#L8).


jdmarshall:
LGTM. @SimenB @zbjornson ?

SimenB (approved):
Agreed, this is great 👍

SimenB:
@siimon could you enable this setting? (And/or give myself and @zbjornson admin access to the repository, that way I can update that setting myself)

Screenshot_20250713-193815

I'm kinda afraid to merge a PR I don't know passes CI 🙈

SimenB:
I _also_ think you should invite @jdmarshall as a collaborator - he's been doing some awesome work in the last few weeks (months?) 😃

ManuelLG92:
Hello @SimenB and @siimon

I am seeing and git says it can be safely merged

. If you need something else from me, just let me know.

Screenshot_20250716-143808

SimenB:
Yes, but the PR is out of date and CI logs have been purged. Can you merge in master?

SimenB:
Thanks! 😀

Issue #616, opened 2024-02-26
Exemplar still gets updated(to one with empty labels) when providing no exemplar labels with the sample


zkx5xkt:
I have a setup where I only upload a small percentage of my traces.
This means that for majority of my samples in a histogram there won't be a corresponding trace.

### Current library implementation: 
When presented with a new sample with no exemplar labels the library overrides the bucket's exemplar with an empty object essentially purging the existing exemplar

Example of metric after sample is published with an exemplar labels: `example_duration_bucket{le="5"} 1 # {traceID="9899898"} 3.0005617 1708951967.457`
Example of metric after sample is published with no exemplar labels afterwards: `example_duration_bucket{le="5"} 1 # {} 3.0005617 1708952967.457`

### Desired library implementation:
When presented with a new sample with no exemplar the library should retain the existing exemplar so it can be scraped with the metric.

Example of metric after sample is published with an exemplar labels:  `example_duration_bucket{le="5"} 1 # {traceID="9899898"} 3.0005617 1708951967.457`
Example of metric after sample is published with no exemplar labels afterwards: `example_duration_bucket{le="5"} 1 # {traceID="9899898"} 3.0005617 1708951967.457`

### Example 
```typescript
import { Span, TraceFlags } from "@opentelemetry/api";
import { setTimeout } from "node:timers/promises";
import { Histogram } from "prom-client";

const histogram = new Histogram({
    name: "example_duration",
    help: "example duration",
    buckets: [0.5, 5],
    labelNames: [],
    enableExemplars: true,
});

async function example(span: Span) {
    const spanContext = span.spanContext();

    const exemplarLabels = spanContext.traceFlags & TraceFlags.SAMPLED ? {
        traceID: spanContext.traceId,
        spanID: spanContext.spanId
    } : undefined;

    const histogramTimer = histogram.startTimer(undefined, undefined);

    await setTimeout(3_000);

    histogramTimer(undefined, exemplarLabels);
}
```

currently for `prom-client@15.1.0` i'm applying the following hotpatch for Histogram (do not overwrite exemplar values when no exemplar labels given):
```patch
--- node_modules/prom-client/lib/histogram.js	2024-02-26 06:50:35.132071300 +0200
+++ node_modules/prom-client/lib/histogram.js	2024-02-26 07:09:28.088071300 +0200
@@ -89,14 +89,16 @@
 		const bound = findBound(this.upperBounds, value);
 		const { bucketExemplars } = this.hashMap[hash];
[...]

PR #620, opened 2024-03-26
fix: avoid updating exemplar values during subsequent metric changes

psimk:
## Problem

The current exemplars implementation expects that every time we update a metric, we will have fresh exemplar labels to set together with the metric. However, setups that have a low sampling rate for their exemplar labels (e.g. trace ids), this means that this implementation would be reseting exemplar labels.

This also creates a lot of noise in graphing solutions like Grafana as there are lots of exemplars displayed without any contextual data attached to them.

## Solution

Before setting an exemplar, or even creating a brand new one, we first will check whether the the exemplar labels were configured with anything other that isn't the default empty object. If there are no exemplar labels, then we simply short circuit the function.

Note, this likely means that we could get rid off the `defaultExemplarLabelSet` field and just handle the `undefined` value, but I wanted to have the least amount of changes in the PR.

_fixes https://github.com/siimon/prom-client/issues/616_

---

@SimenB

### DIFF
--- a/lib/counter.js
+++ b/lib/counter.js
@@ -85,6 +85,7 @@ class Counter extends Metric {
 	}
 
 	updateExemplar(exemplarLabels, value, hash) {
+		if (exemplarLabels === this.defaultExemplarLabelSet) return;
 		if (!isObject(this.hashMap[hash].exemplar)) {
 			this.hashMap[hash].exemplar = new Exemplar();
 		}
--- a/lib/histogram.js
+++ b/lib/histogram.js
@@ -84,6 +84,7 @@ class Histogram extends Metric {
 	}
 
 	updateExemplar(labels, value, exemplarLabels) {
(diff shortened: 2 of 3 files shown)


SimenB:
Sweet, thanks! Gauge and Histograms are not affected?

psimk:
> Sweet, thanks! Gauge and Histograms are not affected?

Gauge's don't seem to have exemplars and I made the change for histograms in [lib/histogram.js](https://github.com/siimon/prom-client/pull/620/files#diff-4cacad18c9a9f40e8b66ee6f58741215e06d9e1bcd886452f61e74daee8ea387R87).

SimenB:
👍 

(I meant summary, but I see there's no exemplar support there either).

Could you add a changelog entry? (also, CI is failing)

psimk:
> Could you add a changelog entry? (also, CI is failing)

Added.

SimenB (approved):
thanks!

Issue #623, opened 2024-03-27
Provide exclusions for default metrics collected


LukasDeco:
Hello!

I am trying to utilize `prom-client` in a `bun` application and unfortunately bun does not yet support the `eventLoopLag` metric. I just want to be able to exclude this but there doesn't seem to be a way, can I just add some kind of parameter for excluding metrics?

LukasDeco:
I actually also need to exclude heap space.

Issue #570, opened 2023-05-12
Is it possible to use in Bun/Deno/other runtime?


krutoo:
Hi, currently i use prom-client in nodejs app and want to use it in other runtime such as Deno or Bun

Is it possible today?

SimenB:
I dunno. We use a bunch of node specific APIs, but mostly in the default metrics. I guess you could omit them?

I've never used either Deno or Bun, so you'd have to tell us and potentially send PRs for changes needed 🙂

krutoo:
Yes, node and v8 specific default metrics can be omitted

zbjornson:
@krutoo if you want to test my PR, that would be helpful: `bun install siimon/prom-client#zb/bun`.

dawidsowardx:
> Yes, node and v8 specific default metrics can be omitted

@krutoo How? Is it documented anywhere

krutoo:
> > Yes, node and v8 specific default metrics can be omitted
> 
> @krutoo How? Is it documented anywhere? I want to omit metric by metric

@dawidsowardx I mean in my case I skip call `PromClient.collectDefaultMetrics()` and collect only custom metrics

SimenB:
https://github.com/siimon/prom-client/releases/tag/v15.1.2

PR #599, opened 2023-11-25
feat: allow running in Bun, add to CI

zbjornson:
Adds feature testing to more default metrics so that Bun (and possibly Deno) work. The metrics are all called `node_something_something` /shrug.

I think this is all that's required, but our tests didn't catch any failures without these changes. Appreciate any beta-testing from the community.

~For some reason, the tests are way slower in Bun (80 vs 5 seconds). idk how to use Bun's profiler with jest.~ evidently only on my WSL2 environment

https://github.com/oven-sh/bun/issues/4708 will obviate these changes.

Fixes #570

PR #624, opened 2024-04-02
fix: enable bun.js by catching NotImplemented error

owlcode:
Catch [error thrown by bun.js](https://github.com/oven-sh/bun/blob/801e475c72b3573a91e0fb4c3afdd53b437770e1/src/js/node/perf_hooks.ts#L158) when calling `perf_hooks.monitorEventLoopDelay`

Following up on https://github.com/siimon/prom-client/pull/599 - bun.js already introduced some of the missing APIs (or created stubs) so to the best of my knowledge this is the last thing which is required to make default gc stats work for bun (at least it worked for me with this change).

Error is also reported in bun's repository - https://github.com/oven-sh/bun/issues/4708#issuecomment-2030436691

Fixes https://github.com/siimon/prom-client/issues/570
Closes https://github.com/siimon/prom-client/issues/623

### DIFF
--- a/lib/metrics/eventLoopLag.js
+++ b/lib/metrics/eventLoopLag.js
@@ -63,8 +63,12 @@ module.exports = (registry, config = {}) => {
 
 				histogram.reset();
 			};
-		} catch {
-			// bun has `perf_hooks.monitorEventLoopDelay` defined but throws "NotImplemented" when called
+		} catch (e) {
+			if (e.code === 'ERR_NOT_IMPLEMENTED') {
+				return; // Bun
+			}
+
+			throw e;
 		}
 	}
 
--- a/lib/metrics/eventLoopLag.js
+++ b/lib/metrics/eventLoopLag.js
@@ -37,32 +37,35 @@ module.exports = (registry, config = {}) => {
 	const labelNames = Object.keys(labels);
 	const registers = registry ? [registry] : undefined;
 
-	let collect;
-	if (!perf_hooks || !perf_hooks.monitorEventLoopDelay) {
-		collect = () => {
-			const start = process.hrtime();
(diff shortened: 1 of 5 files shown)


lib/metrics/eventLoopLag.js
```diff
@@ -37,32 +37,35 @@ module.exports = (registry, config = {}) => {
...
+				lagStddev.set(labels, histogram.stddev / 1e9);
+				lagP50.set(labels, histogram.percentile(50) / 1e9);
+				lagP90.set(labels, histogram.percentile(90) / 1e9);
+				lagP99.set(labels, histogram.percentile(99) / 1e9);
+
+				histogram.reset();
+			};
+		} catch {
+			// bun has `perf_hooks.monitorEventLoopDelay` defined but throws "NotImplemented" when called
+		}
```
SimenB, line 68: the `catch` should check for the specific error bun throws and rethrow any other

lib/metrics/heapSpacesSizeAndUsed.js
```diff
@@ -11,6 +11,11 @@ METRICS.forEach(metricType => {
 });
 
 module.exports = (registry, config = {}) => {
+	try {
+		v8.getHeapSpaceStatistics();
+	} catch {
+		return; // Bun
+	}
```
SimenB, line 18: ditto

SimenB (changes requested):

SimenB:
Can you include the CI stuff from #599?

owlcode:
> Can you include the CI stuff from https://github.com/siimon/prom-client/pull/599?

Done

SimenB:
https://github.com/siimon/prom-client/releases/tag/v15.1.2

PR #625, opened 2024-04-16
fix: correctly get and set top level `contentType` property

SimenB:
It currently just reads the default prop, and then no longer affect or reflect reality

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,10 +17,8 @@ Object.defineProperty(exports, 'contentType', {
 		exports.register.setContentType(value);
 	},
 });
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -7,11 +7,18 @@
 
 exports.register = require('./lib/registry').globalRegistry;
 exports.Registry = require('./lib/registry');
-exports.contentType = require('./lib/registry').globalRegistry.contentType;
```
SimenB, line 10: this into getter/setter is the fix (first commit of PR)

Issue #628, opened 2024-04-23
Prom-Client aggregation causing master to choke in cluster mode


ssg2526:
I implemented a small cluster module based code. I have one master and eight workers running. I am using AggregatorRegistry class like:
```js
const AggregatorRegistry = client.AggregatorRegistry;
const aggregatorRegistry = new AggregatorRegistry();

const defaultLabels = { 'application': 'test_app' };
masterRegistry.setDefaultLabels(defaultLabels);

const externalMethodTimings = new client.Histogram({
    name: 'external',
    labelNames: ['status', 'method', 'url'],
    help: 'None'
});
```
The worker has one end point which responds with a 30ms latency and also a middleware to increase the number of metrics to simulate the production like behaviour.
```js
let express = require('express');
let app = express();
app.use(function(req, res, next)
{
    req.startTime = new Date().getTime();
    req.custom = req.url+"/"+counter%700
    counter++;
    next();
});

app.use(function(req, res, next){
    var startAt = req.startTime;
    res.on('finish', ()=> {
        try {
        var diff = new Date().getTime() - startAt;
        externalMethodTimings.labels({'method': req.method, 'url': req.custom}).observe(diff/1000.0);
        return;
        } catch(err){
        
        }
    });
    next();
});

app.get('/worker', function(req, res){
    setTimeout(()=>{
        return res.status(200).send('worker Server is running');
    },30);
});

app.listen(port, function () {
    console.log('worker server listening on port %d', port);
});
```
Now I have a master that has the following code
```js
app.get('/master', async (req, res) => {
    try {
        var start = new Date().getTime();
        const workerMetrics = aggregatorRegistry.clusterMetrics();
		            
        workerMetrics.then(values => {
            var allMetrics = values[0] + values[1]; 
            console.log(new Date().getTime() - start);
            res.set('Content-Type', aggregatorRegistry.contentType);
            res.send(allMetrics);
        });
    } catch (e) {
	    res.statusCode = 500;
	    res.send(e.message);
    }
});

app.listen(master_port, function(){
    console.log('master server listening on port %d', master_port);
});
```
When I run this application and put load using apache benchmark using the below configs. Also I ran a watcher to hit the cluster metric endpoint on master at every 1 second to simulate prometheus scraping. The command is given below -
```
ab -n 60000 -c 50 "http://localhost:8080/worker"

watch -n 1 "curl http://localhost:3000/master > /dev/null";
```
[...]

PR #631, opened 2024-05-20
perf: fix for cluster aggregation performance

ssg2526:
The changes include performance improvements for cluster mode aggregations flow. The changes are 3 fold -
* The hashing of each metric is moved to workers. Now each worker hashes the data and adds that in a key named hash in the metrics json. 
  * This is distributing the hashing task among workers rather than master doing the CPU task
* The Map which was being built on every request freshly, is made global and only if a new metric comes then that will be put in the map else we will get the value and push the item in the array.
  * The insertion of large keys is very slow in Map as it is maintaining insertion order and this will minimize the insertions
* A major choking point was when master asks workers to provide metrics and workers sending those metrics over IPC causing the choking in IPC which was hindering the request routing from master to workers. The change here is that workers don't send metrics over IPC rather they are writing them in a file and sending master the file name over IPC. Master is then reading the metrics from the files and deleting the files.
  * This is removing the congestion on IPC and requests routing from master to workers not getting hampered.

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -30,10 +30,17 @@ let requestCtr = 0; // Concurrency control
 let listenersAdded = false;
 const requests = new Map(); // Pending requests for workers' local metrics.
 
+let enablePerformanceOptimizedVarient = false;
+
 class AggregatorRegistry extends Registry {
-	constructor(regContentType = Registry.PROMETHEUS_CONTENT_TYPE) {
+	constructor(regContentType = Registry.PROMETHEUS_CONTENT_TYPE, options = {}) {
 		super(regContentType);
 		addListeners();
+
+		if (options.enablePerformanceOptimizedVarient !== undefined) {
+			enablePerformanceOptimizedVarient =
+				options.enablePerformanceOptimizedVarient;
+		}
 	}
 
 	/**
@@ -146,9 +153,11 @@ class AggregatorRegistry extends Registry {
 	 * use a registry/registries other than the default global registry.
 	 * @param {Array<Registry>|Registry} regs Registry or registries to be
 	 *   aggregated.
+	 * @param {Object} options Configuration options for cluster optimizations
+	 * @param {boolean} options.enablePerformanceOptimizedVarient Enable performance optimizations
 	 * @return {void}
 	 */
-	static setRegistries(regs) {
+	static setRegistries(regs, options = {}) {
 		if (!Array.isArray(regs)) regs = [regs];
 		regs.forEach(reg => {
 			if (!(reg instanceof Registry)) {
@@ -156,6 +165,11 @@ class AggregatorRegistry extends Registry {
 			}
 		});
 		registries = regs;
+
+		if (options.enablePerformanceOptimizedVarient !== undefined) {
+			enablePerformanceOptimizedVarient =
+				options.enablePerformanceOptimizedVarient;
+		}
 	}
 }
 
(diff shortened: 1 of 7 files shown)


ssg2526:
@SimenB @siimon kindly approve for workflows. I have added the prettier fixes because of which the earlier workflows failed.

lib/cluster.js
```diff
@@ -198,10 +210,32 @@ function addListeners() {
...
+								const key = `${value.metricName}_${hash}`;
+								value.hash = key;
+							});
+						});
+						// adding request id in file path to handle concurrency
+						const filename = path.join(
+							os.tmpdir(),
+							`metrics-${process.pid}-${message.requestId}.json`,
+						);
+						fs.writeFile(filename, JSON.stringify(metrics), err => {
```
zbjornson, line 225: I'm really hesitant to involve files, since that suddenly involves the page cache, file system, more syscalls and possibly other surprises like filesystem permissions, running out of disk space and cleaning up. How much of an improvement did this yield? Did you benchmark before or after Node.js v18.6.0, which has [this optimization](https://github.com/nodejs/node/pull/42931)? How much disk I/O load did your system have when benchmarking? I think IPC is "supposed" to be faster than involving the filesystem.

zbjornson (commented):
Thanks for working on this!

Do you have any benchmarks you can show? (both the code and the results)

lib/cluster.js
ssg2526: The benchmarks that I have performed are on our own application. The screenshots of the results are shared in the
this [Issue Link](https://github.com/siimon/prom-client/issues/628). The benchmarks were done on a 16 core machine and approximately 250-300 metrics (total size ~500 Kbs). At P99 and above levels we are getting ~10x improvement in app performance. The major choking point was with IPC. The detailed bifurcation that I have done on my local machine which is 8 core machine is given In the screenshot below. and the code used is in below given zip file. Have added some consoles in existing promc-client in this zip. Here if we see the worst IPC time is always higher. I had run multiple iterations and same.similar results are found. haven't tested with node v18.6.0 yet. will test the same code with that as well and share the results in the same thread
[cluster_test.zip](https://github.com/user-attachments/files/15531600/cluster_test.zip)
Screenshot 2024-06-03 at 2 27 48 PM

ssg2526: In these benchmarks ([issue link](https://github.com/siimon/prom-client/issues/628)) the scraping interval used was every 5 seconds and the throughput on the app was about 1100-1200 RPS.

ssg2526: Hi @zbjornson 

- I also tried the code in the zip file with node v18.6.0. I am getting the similar results without any improvements in the logged timings.
- I think we shouldn't use IPC for this communication because that can create a problem whenever the sizes of metrics go beyond a certain point and that will start blocking the requests routing at high throughputs. 
- We can ask user to provide permissions for /tmp folder. From prom-client performance POV i didn't see much of a difference with files but again I am only hitting it once every 5 seconds.
- If not files then we should think of other ways of communication to solve this completely

ssg2526: Hi @zbjornson @SimenB we went live in production with these changes about 3 months ago and everything is working fine for us. We are not seeing any additional latencies post this change. We are service at more than 100,000 RPS at peak for this service. If required I can share the production results as well.
I understand that introducing file system can pose different challenges for the library, But we can explore communication with HTTP or some other mode of communication (Not IPC) when we have larger size of the metrics as the current solution will cause very high tail latencies (>P99) at high throughput and larger metrics sizes. 
Requesting your inputs on this, In terms of how we can take this forward as we don't want to diverge from the library maintaining our own custom version.

mrnonz:
Any concern in filesystem anymore? Should we consider to release it?

ssg2526:
> Any concern in filesystem anymore? Should we consider to release it?

@mrnonz we haven't faced any issues till now related to filesystem. This change would require the users to provide the access to /tmp path for prom-client

lib/cluster.js
```diff
@@ -175,19 +178,28 @@ function addListeners() {
...
-				if (request.pending === 0) {
-					// finalize
-					requests.delete(message.requestId);
-					clearTimeout(request.errorTimeout);
-
-					const registry = AggregatorRegistry.aggregate(request.responses);
-					const promString = registry.metrics();
-					request.done(null, promString);
-				}
+				fs.readFile(message.filename, 'utf8', (err, data) => {
```
BourgoisMickael, line 181: ## About IPC

The IPC congestion is because of the number of request or the payload size ?
___
Can we have it as a customizable option and keep the default behavior instead ?
So users could pick if they want file vs ipc communication and there would be no breaking change.
___
Instead of sending filename to worker, could the worker keep a permanent file (named with its pid) and use IPC only to notify the master to read the file (you could use the 1st line of the file to reconcile requestId) or use `fs.watchFile`
___
Instead of file, could using unix socket or named pipes (if available) give better perf ?


## About hashing

The master process should be used only to initiate and restart workers and aggregate metrics. So it should not need that much CPU compared to workers that handle the workload. And this will put a little more load to all the workers at constant interval.

I guess you could have bad performance when you have a lot of workers and short scrape interval then.

Thus same for this feature, could it be an option to configure prom-client with hashing either in master or distributed on workers, and by default keep the current behavior to introduce no breaking changes.

ssg2526: @BourgoisMickael sorry for late reply. I agree with making both file based communication and the hashing configurable. I'll keep a single config for both of these so that we can keep the same changes.
About the file name. The workers are sending the file name to the master process and not the other way around. We need to keep the request Id in the file name to avoid concurrency cases.
I'll update the PR with the requested changes for configurability

ssg2526:
@BourgoisMickael kindly approve the workflows for the suggested changes

jdmarshall:
So after staring at this code for the better part of a week, and spending a good bit of time at my previous job trying to write something like this for opentelemetry, these are my thoughts on both this PR and the code:

- If process.send() is not fast enough, the solution is not going through the filesystem, it's handing over a socket via `send()` and streaming the data. Neither Prometheus nor Opentelemetry are built on reliable delivery, but IPC through the filesystem is considered one of the classic blunders. Filesystems are at least 10x more insane than you can possibly guess.
- Rather than sending the hashObject() response back over the link, we would get some utility from sorting the keys on the sending end. Right now the code fights this in a few places (there are tests that insist that labels and default labels should merge in a specific order that is definitely *not* sorted).  
- I have made some PRs that improve the performance of the label processing and Grouper, which should help a bit on both ends of the connection. I have another PR almost ready and one more after that, which should modify the ratios that are driving this PR. It would be worth someone with a larger corpus than I to weigh in on what things look like now.

jdmarshall:
https://github.com/siimon/prom-client/pull/692 answers most of these concerns.

The main thing I have not done is recycling the per-stat collectors between aggregation calls. There are functions which support it, but I'm not calling them at present. That can be done as a follow-on.

PR #633, opened 2024-06-21
Improve error message on labels

leftieFriele:
This fix ensures consumers of the library get more insights into which labels are causing an error to occur. 

With the new error message it is easier for a developer to debug what happened in the case where you accidentally pass in a bad argument.

If `getLabels` gets called like this:
```js
getLabels(['label1', 'label2'], ['arg1']);
```
it used to throw...
```
Error: Invalid number of arguments
    at exports.getLabels 
```
now it throws this:
```
Error: Invalid number of arguments: "arg1" for label names: "label1,label2".
    at exports.getLabels 
```

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -54,7 +54,7 @@ exports.getLabels = function (labelNames, args) {
 
 	if (labelNames.length !== args.length) {
 		throw new Error(
-			`Invalid number of arguments: "${args.join()}" for label names: "${labelNames.join()}".`,
+			`Invalid number of arguments (${args.length}): "${args.join(', ')}" for label names (${labelNames.length}): "${labelNames.join(', ')}".`,
 		);
 	}
 
--- a/lib/util.js
+++ b/lib/util.js
@@ -54,7 +54,7 @@ exports.getLabels = function (labelNames, args) {
 
 	if (labelNames.length !== args.length) {
 		throw new Error(
-			`Invalid number of arguments. Got ${args.join()}. Expected ${labelNames.join()}.`,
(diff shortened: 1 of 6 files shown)


lib/util.js
```diff
@@ -53,7 +53,9 @@ exports.getLabels = function (labelNames, args) {
 	}
 
 	if (labelNames.length !== args.length) {
-		throw new Error('Invalid number of arguments');
+		throw new Error(
+			`Invalid number of arguments: "${args.join()}" for label names: "${labelNames.join()}".`,
```
SimenB, line 57: I wonder if something like this is even more helpful?

```suggestion
			`Invalid number of arguments (${args.length}): "${args.join(', ')}" for label names (${labelNames.length}): "${labelNames.join(', ')}".`,
```

?

SimenB (approved):
thanks! just a quick suggestion, happy to merge after that 🙂

lib/util.js
leftieFriele: 💪 I like this suggestion, I'll add it.

leftieFriele:
> thanks! just a quick suggestion, happy to merge after that 🙂 

👍 I didn't squash it on the branch, so squash and merge. thanks for the quick reply ❤️

SimenB (approved):

stalet:
When is this scheduled to be released ?

SimenB:
I'm currently away at a festival, but I can make a release on Sunday (unless @zbjornson gets to it before me)

SimenB:
had some time this morning, so published in https://github.com/siimon/prom-client/releases/tag/v15.1.3

PR #635, opened 2024-06-22
chore: update prettier and husky

SimenB:
Why not

Also migrates from `eslint-plugin-node` to `eslint-plugin-n` as the former is unmaintained and replaced

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
 		"@typescript-eslint/parser": "^6.21.0",
 		"eslint": "^8.32.0",
 		"eslint-config-prettier": "^9.1.0",
-		"eslint-plugin-node": "^11.0.0",
+		"eslint-plugin-n": "^16.0.0",
 		"eslint-plugin-prettier": "^5.0.1",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -38,14 +39,14 @@
 		"@typescript-eslint/parser": "^6.21.0",
 		"eslint": "^8.32.0",
 		"eslint-config-prettier": "^9.1.0",
-		"eslint-plugin-node": "^11.0.0",
+		"eslint-plugin-n": "^16.0.0",
```
SimenB, line 42: `17` is the latest, but that dropped support for Node 16

package.json
```diff
@@ -58,9 +59,9 @@
 		"testRegex": ".*Test\\.js$"
 	},
 	"lint-staged": {
-		"*.js": "eslint --fix",
-		"*.{ts,md,json,yml}": "prettier --write",
-		".{eslintrc,travis.yml}": "prettier --write"
```
SimenB, line 63: oh, oops - I deleted the wrong entry 😅

PR #636, opened 2024-06-22
Update README.md - fix wrong variable name

korengal:
new client.Counter should initiate counter variable and not gauge


SimenB (approved):
good catch, thanks!

PR #638, opened 2024-08-08
Fix the awaitable collect() calls.

farcaller:
In case if non-native promises are used, the Promise object won't be instanceof Promise. That's still valid in JS, as Promise is something that implements then(). It's always safe to cast a maybe-promise into native Promise with Promise.resolve().

### DIFF
--- a/lib/counter.js
+++ b/lib/counter.js
@@ -109,7 +109,7 @@ class Counter extends Metric {
 	async get() {
 		if (this.collect) {
 			const v = this.collect();
-			if (v instanceof Promise) await v;
+			if ((v instanceof Promise) || (v.then instanceof Function)) await v;
 		}
 
 		return {
--- a/lib/gauge.js
+++ b/lib/gauge.js
@@ -108,7 +108,7 @@ class Gauge extends Metric {
 	async get() {
 		if (this.collect) {
 			const v = this.collect();
-			if (v instanceof Promise) await v;
+			if ((v instanceof Promise) || (v.then instanceof Function)) await v;
 		}
 		return {
 			help: this.help,
--- a/lib/histogram.js
+++ b/lib/histogram.js
@@ -108,7 +108,7 @@ class Histogram extends Metric {
 	async getForPromString() {
 		if (this.collect) {
 			const v = this.collect();
-			if (v instanceof Promise) await v;
+			if ((v instanceof Promise) || (v.then instanceof Function)) await v;
 		}
 		const data = Object.values(this.hashMap);
 		const values = data
--- a/lib/summary.js
+++ b/lib/summary.js
@@ -51,7 +51,7 @@ class Summary extends Metric {
 	async get() {
 		if (this.collect) {
 			const v = this.collect();
-			if (v instanceof Promise) await v;
+			if ((v instanceof Promise) || (v.then instanceof Function)) await v;
 		}
 		const hashKeys = Object.keys(this.hashMap);
 		const values = [];


lib/counter.js
```diff
@@ -108,8 +108,7 @@ class Counter extends Metric {
 
 	async get() {
 		if (this.collect) {
-			const v = this.collect();
-			if (v instanceof Promise) await v;
+			const v = await Promise.resolve(this.collect());
```
zbjornson, line 111: The original code was intended to avoid entering the microtask queue unless necessary, for performance (i.e. don't `await` unless necessary).

What about duck-typing for a [thenable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#thenables)?

zbjornson (changes requested):
Is anyone actually using non-native promises now? They've been available in Node since Feb 2015 and this library requires native Promise support. I'm kind of inclined to say we shouldn't make this change and should instead tell users implementing custom `collect()` functions to return a native Promise.

farcaller:
> Is anyone actually using non-native promises now?

Funnily, yes. ClojureScript's Promesa uses non-native Promises (and that's how I ended up going down this rabbit hole).

lib/counter.js
farcaller: That's an interesting point I didn't consider. I suppose duck-typing is fair. Updated the PR.

zbjornson:
What do you think about doing this on the consumer side though? i.e. in your custom `collect()` function, `return Promise.resolve(/*the Promesa promise */)`.

farcaller:
I'm giving this another thought and I wonder: is this actually the use case where the code must be extremely performant? I'd expect that metrics page would be scraped once in a while, or pushed once in a while, where "a while" is dozens of seconds. At that scale, optimizing for microtask scheduling seems a bit futile, is it not?

zbjornson:
> is this actually the use case where the code must be extremely performant?

prom-client gets a surprising number of [PRs to improve its performance](https://github.com/siimon/prom-client/pulls?q=is%3Apr+perf+). I think it's valid to want observability frameworks to be as low of overhead as possible. If you have 100s of metrics with `collect()` functions, that could have an impact.

I'm +1 on a doc-only change that says the return type needs to be a native Promise. The typescript types are already correct fwiw: they say `Promise` and not `PromiseLike`.

jdmarshall:
@farcaller This is a programming language with a single thread and a high barrier to sending structured data to another thread for data processing.  

Every time the collection interval hits, your p95 times will spike. And if you use cluster mode, it will spike on all threads on the same server at exactly the same time, which will definitely hit your p95 times if you have less than 50 servers in your cluster.

jdmarshall:
I noticed this code earlier today when working on #683 

I have a different proposal here.

Sometime around Node 16 or 18 (my memory is fuzzy and I no longer work for the company where I could check commit history to verify), an optimization to v8 allowed awaiting of non-promises to be more efficient.

It changed `await("foo")` to eliminate triggering nextTick(), which I discovered when a couple of tests started throwing bizarre error messages that did not make sense. It turned out that we were mocking an async function with a sync mock, and then part of the setup code started running _after_ the test condition instead of before as it previously had - because v8 was no longer yielding to look at the promise resolution queue if the await was on a synchronous call.

tl;dr: has anyone tried just non-conditionally awaiting the results and benchmarking how much the JIT optimizes that code?

lib/counter.js
```diff
@@ -109,7 +109,7 @@ class Counter extends Metric {
 	async get() {
 		if (this.collect) {
 			const v = this.collect();
-			if (v instanceof Promise) await v;
```
jdmarshall: Rather than changing this code, the correct solution here is to suggest to people who are still using third party promises (Native promises have been faster since Node 16 at the latest, but I believe Node 12), is for them to change their collect method to be async and let node sort it out.


If you have an application that is half async, and half Promises, you can pick boundaries where all code is required to be async and you can call into the 'old code' any time you want, or have the old code call your new code any time you want.

If a function is async, then it returns a promise.

```
   promises.push(fancyNewAsyncFunction(important, arguments));
   ...
```

If a function is async, you can return a Promise.

```
  return oldPromiseBasedFunction();
```

Here's a gist showing Bluebird working with async/await:

https://gist.github.com/jdmarshall/2632b368bd89f844ba653c15132b9724


You don't need to manually thunk, and in fact it IMO makes the code harder to read if you do and likely slower.

So, long story short, we don't need any of this code. It can all go, and save some ticks in the process.

jdmarshall:
Mark the collect() function in your custom metric as async and let V8 fix this problem for you.

PR #639, opened 2024-08-10
fix: avoid duplicate buckets in histogram

erlang23:
Hello,

this is an user-introduced bug that drove me crazy debugging Prometheus itself.

prom-client allows you to define the same LE bucket label multiple times. This leads to wrong values and makes the histograms useless.

With this commit I remove the possible duplicates before setting the buckets to the upperBounds. Also included an unit  test.

### DIFF
--- a/lib/histogram.js
+++ b/lib/histogram.js
@@ -32,7 +32,7 @@ class Histogram extends Metric {
 			}
 		}
 
-		this.upperBounds = this.buckets;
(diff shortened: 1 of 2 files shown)


jdmarshall:
I'm okay with this solution, but I'd also be okay with an error as well. How do we feel about silent errors of this sort? Better to complain or better to let it go?

PR #645, opened 2024-11-18
feat: allow specifying timeout for cluster metrics aggregation

vitush93:
Adds ability to specify cluster metrics aggregation timeout.

Overriding hard-coded timeout as a workaround for #628 where scraping in larger projects end up in a timeout most of the time.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -137,6 +137,7 @@ export class AggregatorRegistry<
 	/**
 	 * Gets aggregated metrics for all workers.
 	 * @param {ClusterMetricsOptions|undefined} options Additional options for cluster metrics aggregation
+	 * @param {number} [options.timeoutMs=5000] Timeout to wait for all workers to respond before error.
 	 * @return {Promise<string>} Promise that resolves with the aggregated
 	 * metrics.
 	 */
@@ -194,6 +195,7 @@ export enum MetricType {
 }
 
 export type ClusterMetricsOptions = {
+	/** Timeout to wait for all workers to respond before error. */
 	timeoutMs?: number;
 };
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -136,10 +136,11 @@ export class AggregatorRegistry<
 > extends Registry<T> {
 	/**
 	 * Gets aggregated metrics for all workers.
+	 * @param {ClusterMetricsOptions|undefined} options Additional options for cluster metrics aggregation
 	 * @return {Promise<string>} Promise that resolves with the aggregated
 	 * metrics.
 	 */
-	clusterMetrics(): Promise<string>;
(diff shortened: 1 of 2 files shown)


lib/cluster.js
```diff
@@ -36,10 +36,11 @@ class AggregatorRegistry extends Registry {
 	/**
 	 * Gets aggregated metrics for all workers. The optional callback and
 	 * returned Promise resolve with the same value; either may be used.
+	 * @param {ClusterMetricsOptions|undefined} options Additional options for cluster metrics aggregation.
```
BourgoisMickael, line 39:
```suggestion
	 * @param {ClusterMetricsOptions|undefined} options Additional options for cluster metrics aggregation.
	 * @param {number} [options.timeoutMs=5000] Timeout to wait for all workers to respond before error.
```

index.d.ts
```diff
@@ -136,10 +136,11 @@ export class AggregatorRegistry<
 > extends Registry<T> {
 	/**
 	 * Gets aggregated metrics for all workers.
+	 * @param {ClusterMetricsOptions|undefined} options Additional options for cluster metrics aggregation
```
BourgoisMickael, line 139:
```suggestion
	 * @param {ClusterMetricsOptions|undefined} options Additional options for cluster metrics aggregation
	 * @param {number} [options.timeoutMs=5000] Timeout to wait for all workers to respond before error.
```

index.d.ts
```diff
@@ -192,6 +193,10 @@ export enum MetricType {
 	Summary,
 }
 
+export type ClusterMetricsOptions = {
+	timeoutMs?: number;
```
BourgoisMickael, line 197:
```suggestion
	/** Timeout to wait for all workers to respond before error. **/
	timeoutMs?: number;
```

This can be documented as well

BourgoisMickael (approved):
:+1: Yes please we need this !
I hope author @SimenB @siimon can merge this soon

jdmarshall:
@vitush93 see #692

jdmarshall:
I believe this is fundamentally the wrong solution. 

 - You cannot fire a timeout event unless the event loop is unblocked. So you cannot abort out of a tight loop because you will never see the signal
 - Timeouts are necessary but vastly insufficient. Metrics should not cause you to blow your SLAs. So what all of the tight inner loops in this module needs are interruptions to allow the event loop to cycle and REAL WORK to progress while the metrics are being processed.
 - The real test of exceeding your time budget is if a second metric request shows up while you are still processing the first. 
 - The aggregation thread cannot send another message, or do anything else like route requests, if it itself is blocking its own event loop.

I would strongly recommend that the team take a step back and consider the 'niceness' problem first, before entertaining PRs like this. 

First make the steps as fast as you can (see my PRs) and then split up what's left to make them disappear into the background.

PR #647, opened 2024-11-22
fix: Add missing properties to metrics types from abstract class Metric

kouak:
The typescript definitions for all 4 metric classes only include methods.

This PR adds properties in the typescript definitions to the 4 metric classes, which come from the abstract `Metric` class.


jdmarshall:
#678

jdmarshall:
Resubmitted

PR #654, opened 2025-04-25
ResetMetric

sc-crescendo:
Added function reset metric which allows resetting single metric by name

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -37,7 +37,12 @@ export class Registry<
 	/**
 	 * Reset all metrics in the registry
 	 */
-	resetMetrics(): void;
+	resetMetrics(name: string): void;
+
+	/**
+	 * Reset one naed metric
+	 */
+	resetMetric(): void;
 
 	/**
 	 * Register metric to register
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -39,6 +39,11 @@ export class Registry<
 	 */
 	resetMetrics(): void;
 
+	/**
+	 * Reset one naed metric
```
mindw, line 43: `naed` seems like a typo? perhaps:
```suggestion
	 * Reset a named metric
	 * @param metric Metric to reset
```

lib/registry.js
```diff
@@ -165,6 +165,11 @@ class Registry {
 			this._metrics[metric].reset();
 		}
 	}
+	resetMetric(name) {
+
```
mindw, line 169: seems like a redundant space
```suggestion
```

mindw:
It seems reasonable a test for this new functionality will be requested.

Not the/a maintainer.

jdmarshall:
Is this doing the same thing as looking up the metric and then resetting it directly?

PR #658, opened 2025-06-24
perf: Utilize Maps for metric collection 

jdmarshall:
As of sometime around the middle of Node 16, the performance of Map for long-lived, mutable objects exceeded that of {}. I believe this gap is more pronounced in Node 18 as well.

The deal is that the representation for Objects that repeatedly change turns into a weird chain that occupies (in my benchmarks at least) ~40% more memory. So once v8 made Map.set/get faster than map[], we all should have switched to Maps and Sets.

Between the slightly faster functions and the lower memory pressure this can create a notable improvement.

Making a change like this on my last project for a long lived cache resulted in much better memory behavior and also about a 5% improvement in TTFB - just from changing a popular lookup table.

Note that the final change in this PR constitutes a change for anyone making their own subclasses, and therefore is probably breaking. Let me know if you want me to split this into two PRs, but that's where the bulk of the changes show up. The Registry just doesn't change frequently enough.

```
✓ registry ➭ getMetricsAsJSON#1 with 64 is 736.6% faster.
✓ registry ➭ getMetricsAsJSON#2 with 8 is 753.0% faster.
✓ registry ➭ getMetricsAsJSON#2 with 4 and 2 with 2 is 416.9% faster.
✓ registry ➭ getMetricsAsJSON#2 with 2 and 2 with 4 is 2829% faster.
✓ registry ➭ getMetricsAsJSON#6 with 2 is 956.4% faster.
✓ registry ➭ metrics#1 with 64 is 440.3% faster.
✓ registry ➭ metrics#2 with 8 is 1283% faster.
✓ registry ➭ metrics#2 with 4 and 2 with 2 is 1282% faster.
✓ registry ➭ metrics#2 with 2 and 2 with 4 is 1512% faster.
✓ registry ➭ metrics#6 with 2 is 107.3% faster.
```

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -14,24 +14,42 @@ exports.getValueAsString = function getValueString(value) {
 	}
 };
 
+/**
+ * @function removeLabels
+ * @param {Map} hashMap
+ * @param {string[]} labels
+ * @param {string[]} sortedLabelNames
+ */
 exports.removeLabels = function removeLabels(
 	hashMap,
 	labels,
 	sortedLabelNames,
 ) {
-	const hash = hashObject(labels, sortedLabelNames);
-	delete hashMap[hash];
+	hashMap.delete(hashObject(labels, sortedLabelNames));
 };
 
+/**
+ * @function setValue
+ * @param {Map} hashMap
+ * @param {*} value
+ * @param {object} labels
+ */
 exports.setValue = function setValue(hashMap, value, labels) {
 	const hash = hashObject(labels);
-	hashMap[hash] = {
+	hashMap.set(hash, {
 		value: typeof value === 'number' ? value : 0,
 		labels: labels || {},
-	};
+	});
 	return hashMap;
 };
 
+/**
+ * @function setValueDelta
+ * @param {Map} hashMap
+ * @param {*} deltaValue
+ * @param {object} labels
+ * @returns {Map}
+ */
 exports.setValueDelta = function setValueDelta(
 	hashMap,
 	deltaValue,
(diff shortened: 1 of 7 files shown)


SimenB:
Thanks for digging into it!

I'm fine with making a major release dropping node 16 and 18 and this API change 👍

package.json
```diff
@@ -12,8 +12,8 @@
 		"node": "^16 || ^18 || >=20"
 	},
 	"scripts": {
-		"benchmarks": "node ./benchmarks/index.js",
-		"test": "npm run lint && npm run check-prettier && npm run compile-typescript && npm run test-unit",
+		"benchmarks": "node --max-old-space-size=8192 ./benchmarks/index.js",
```
SimenB, line 15: landing this in a separate PR makes sense 👍

jdmarshall:
This still works on node 16. It’s just that the numbers are less impressive.

package.json
jdmarshall: Okay

jdmarshall:
I got warnings about a CI error as email, but I'm not seeing them here in the PR.  Shouldn't they be listed here as well?

zbjornson:
I just approved the workflow run. Do you see the errors here now?

Looks good to me once the two failed checks are resolved.

jdmarshall:
Sure. Are you running different eslint settings that “test”? Thought I already caught these.

jdmarshall:
> $ npm run lint
> 
> > prom-client@15.1.3 lint
> > eslint .
> 
> =============
> 
> WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.
> 
> You may find that it works just fine, or you may not.
> 
> SUPPORTED TYPESCRIPT VERSIONS: >=4.3.5 <5.4.0
> 
> YOUR TYPESCRIPT VERSION: 5.8.3
> 
> Please only submit bug reports when using the officially supported version.
> 
> =============
> 

I'm getting absolutely nothing.

lib/util.js
```diff
@@ -14,39 +14,65 @@ exports.getValueAsString = function getValueString(value) {
 	}
 };
 
+/**
+ * @function removeLabels
+ * @param {Map} hashMap
+ * @param {string[]} labels
+ * @param {string[]} sortedLabelNames
+ */
+
```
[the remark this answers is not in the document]
jdmarshall: sure

zbjornson:
One last CI error to fix.

zbjornson (approved):
Awesome, thanks!

PR #668, opened 2025-06-24
Add Code Coverage and make sure benchmarks doesn't go OOM

jdmarshall:
I am doing some perf work on prometheus and I should know what code isn't tested well before I go changing it. These changes were split off of a performance PR at the request of the author.


SimenB (approved):
thanks!

PR #669, opened 2025-06-24
Reduce function indirection in histogram.js for easier debugging

jdmarshall:
Also add missing benchmarks for startTimer(), since I'm mucking about in its innards.

In theory removing call() for arrow functions or straight calls improves performance, but I've never managed a benchmark that proves it. This change also reduces closure creation but GC is also difficult to benchmark. Make enough of these changes however and you can get yourself a little teeny tiny notch in your telemetry data that proves you did something (I think I clocked my test case at about .8%, because the calls were slow and so that added up to 5+ms). 

So these are officially just about improving single step debugging. 

One of the places where Prometheus can outshine OpenTelemetry is in trying to figure out why your metrics are confusing/missing/garbled. If you've ever tried to single-step through OTEL it's a maze of twisty little passages, all the same. prom-client has a fairly straightforward architecture and thus is already much better in this regard but I'm interested in making tools that need to 'just work' more approachable.

### DIFF
--- a/lib/metrics/helpers/processMetricsHelpers.js
+++ b/lib/metrics/helpers/processMetricsHelpers.js
@@ -10,7 +10,7 @@ function aggregateByObjectName(list) {
 			continue;
 		}
 
-		if (Object.hasOwnProperty.call(data, listElement.constructor.name)) {
+		if (Object.hasOwn(data, listElement.constructor.name)) {
 			data[listElement.constructor.name] += 1;
 		} else {
 			data[listElement.constructor.name] = 1;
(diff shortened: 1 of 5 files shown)


lib/metrics/helpers/processMetricsHelpers.js
```diff
@@ -10,7 +10,7 @@ function aggregateByObjectName(list) {
 			continue;
 		}
 
-		if (Object.hasOwnProperty.call(data, listElement.constructor.name)) {
+		if (Object.hasOwn(data, listElement.constructor.name)) {
```
jdmarshall, line 13: Object.hasOwn was introduced in Node 16.9

PR #670, opened 2025-06-24
Reorder number formatting tests to reflect likelihood.

jdmarshall:
It seems to me that the likelihood is that most metrics contain a number, so testing for NaN right out of the box is an odd choice for a function on the hot path.

There's also a triple negative for the happy path and that's not very happy.  If it's a finite number, return immediately. Otherwise faff about with what to send.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -1,16 +1,16 @@
 'use strict';
 
 exports.getValueAsString = function getValueString(value) {
+	if (Number.isFinite(value)) {
+		return `${value}`;
+	}
+
 	if (Number.isNaN(value)) {
 		return 'Nan';
-	} else if (!Number.isFinite(value)) {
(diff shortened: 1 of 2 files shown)


lib/util.js
```diff
@@ -1,16 +1,16 @@
...
-			return '-Inf';
-		} else {
-			return '+Inf';
-		}
+	} else if (value < 0) {
+		return '-Inf';
 	} else {
-		return `${value}`;
+		return '+Inf';
 	}
```
zbjornson, line 14: Fine as-is, but you could also remove all of the `else`s to possibly shorten this fn a bit.

Can you update the changelog please? LGTM otherwise.

zbjornson:
This needs a manual rebase to fix conflicts, then I'm happy to merge it.

Issue #685, opened 2025-07-07
Using prom-client in the browser: Uncaught ReferenceError: process is not defined


alexgleason:
There is an issue when trying to use prom-client in the browser:

```
Uncaught ReferenceError: process is not defined
```

**I am not calling `collectDefaultMetrics`.** I just want to collect some simple metrics like counters in the browser.

I figured out the import itself is what's causing it:

```ts
import { Counter, Histogram, Gauge } from 'prom-client';
```

The issue is here: https://github.com/siimon/prom-client/blob/4c6421b2253c9780a845889004b9a58dac646a1c/lib/metrics/processStartTime.js#L4

(this code is executed at the top-level, at module import time)

I was able to work around this in vite by adding this to my config:

```ts
export default defineConfig(() => ({
  // ...rest of config

  optimizeDeps: {
    include: ['prom-client'],
  },
  define: {
    'process.env': '{}',
    'process.version': '""',
    'process.uptime': '() => 0',
  },
}));
```

Now my crazy in-browser metrics system works:

<img width="1326" height="854" alt="Image" src="https://github.com/user-attachments/assets/3526a3c3-735b-4365-81a6-988f76dacaca" />

<img width="1326" height="854" alt="Image" src="https://github.com/user-attachments/assets/63ea77a9-7f5e-427f-b153-e413a2157a86" />

But it would be good if it "just worked" without needing to know this magic spell

PR #686, opened 2025-07-07
Fix browser compatibility

alexgleason:
Fixes https://github.com/siimon/prom-client/issues/685

Takes care when accessing `process` at the top-level in modules, so that merely importing these modules does not cause an error in browsers. Check the test file at the bottom - this test fails without these changes.

### DIFF
--- a/lib/metrics/version.js
+++ b/lib/metrics/version.js
@@ -1,7 +1,12 @@
 'use strict';
 
 const Gauge = require('../gauge');
-const version = process.version;
+
+const version =
+	typeof process !== 'undefined' && typeof process.version === 'string'
+		? process.version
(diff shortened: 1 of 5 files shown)


jdmarshall:
LGTM

lib/metrics/version.js
```diff
@@ -1,7 +1,12 @@
 'use strict';
 
 const Gauge = require('../gauge');
-const version = process.version;
+
+const version =
+	typeof process !== 'undefined' && typeof process.version === 'string'
+		? process.version
+		: 'v0.0.0'; // Fallback for environments where process.version is not available
```
SimenB: This metric should arguably just self disable

PR #680, opened 2025-07-05
Additional performance improvements for cluster mode

jdmarshall:
Most of these improvements are on the worker side of the connection, but when combined with #679 should provide some
relief for people using this mode.

This PR includes:

- substantial additional code coverage in the benchmarks via addition of more scenarios
- a workaround for benchmark.js not allowing async test bodies (despite supporting async setup and teardown?)
- removal of truthy conditionals in hot paths
- removal of double lookups during map insertions
- reworking hashObject() to optimize for the common case, which is metrics with labels

There's a lot of jitter in these tests, but the common theme is that the benchmarks with no labels show a tiny regression versus the label paths showing 10+% improvements.

Given that the no-label path was previously almost 500x faster, that's a reasonable tradeoff. If prom-client is not fast enough for anyone, it will be someone using a lot of labels.

⚠ counter ➭ inc is 0.06717% acceptably slower.
✓ counter ➭ inc with labels is 25.43% faster.
✓ gauge ➭ inc is 2.639% faster.
✓ gauge ➭ inc with labels is 21.03% faster.
✓ histogram ➭ observe#1 with 64 is 16.87% faster.
✓ histogram ➭ observe#2 with 8 is 10.97% faster.
⚠ histogram ➭ observe#2 with 4 and 2 with 2 is 1.345% acceptably slower.
⚠ histogram ➭ observe#2 with 2 and 2 with 4 is 3.383% acceptably slower.
⚠ histogram ➭ observe#6 with 2 is 3.862% acceptably slower.
✓ registry ➭ getMetricsAsJSON#1 with 64 is 4.661% faster.
✓ registry ➭ getMetricsAsJSON#2 with 8 is 5.511% faster.
✓ registry ➭ getMetricsAsJSON#2 with 4 and 2 with 2 is 5.085% faster.
✓ registry ➭ getMetricsAsJSON#2 with 2 and 2 with 4 is 5.989% faster.
✓ registry ➭ getMetricsAsJSON#6 with 2 is 4.661% faster.
✓ registry ➭ metrics#1 with 64 is 7.167% faster.
✓ registry ➭ metrics#1 with 64 and openMetrics is 4.562% faster.
✓ registry ➭ metrics#2 with 8 is 3.493% faster.
✓ registry ➭ metrics#2 with 8 and openMetrics is 4.773% faster.
✓ registry ➭ metrics#2 with 4 and 2 with 2 is 5.141% faster.
[...]

Issue #561, opened 2023-03-27
Event loop lag metric (gauge) does NOT measure average event lag


shappir:
The built-in metric (gauge) `nodejs_eventloop_lag_seconds` is marked as measuring the average event loop lag: the `aggregator` property value for it is `average`.

However, what it actually reports is the amount of time required to generate and send the previous `metrics` response.

This is equivalent to the maximum event loop lag value rather than the average. To get the actual average you need to use `nodejs_eventloop_lag_mean_seconds` instead (for Node servers that support it). In our case the average event loop lag is under 11ms while the `nodejs_eventloop_lag_seconds` gets as high as 1 second !

I recommend:
1. Fixing `nodejs_eventloop_lag_seconds` to actually report the average event loop lag
2. Introduce a new metric to report the `metrics` response generation time, say `nodejs_metrics_response_time`

**Important**: the amount of time required to generate the `metrics` response remains high, and needs to be improved.

shappir:
This is how `nodejs_eventloop_lag_seconds ` is currently computed:
1. In `eventLoopLag.js` a `Gauge` is created with that name and a [`collect` method](https://github.com/siimon/prom-client/blob/126d4f32afc686b410e1b00ce311526ee3eca645/lib/metrics/eventLoopLag.js#L52)
2. That `collect` method is invoked whenever the metric's [values are requested](https://github.com/siimon/prom-client/blob/126d4f32afc686b410e1b00ce311526ee3eca645/lib/gauge.js#L109)
3. The method measures the high-res time when it is invoked. Since it's a built-in metric, it will be one of the first, and so [records the time at which metrics collection starts](https://github.com/siimon/prom-client/blob/126d4f32afc686b410e1b00ce311526ee3eca645/lib/metrics/eventLoopLag.js#L53)
4. It then uses `setImmediate` to [schedule another measurement](https://github.com/siimon/prom-client/blob/126d4f32afc686b410e1b00ce311526ee3eca645/lib/metrics/eventLoopLag.js#L54). But since that measurement happens in the [next iteration of the event loop](https://nodejs.dev/en/learn/understanding-setimmediate/), it will take place 
**after** the metrics are collected all collected and reported
5. Since the `collect` method doesn't return a promise, the value it measures will only be reported in the next `metrics` invocation.

Since the delta is from when `metrics` starts until after the response is sent, this is what it measures. And it's not updated until the next invocation of `metrics`.

AntunVukicevic:
[...]

PR #687, opened 2025-07-08
Performance: New storage engine for stats, with improved key management

jdmarshall:
This change consolidates a lot of util code into a single object that can guarantee that the functionality is all used consistently, which gets the metrics out of the job of managing labelNames and their sorting.

This results in a 20-45% improvement in benchmark time for most of the metric types. Gauges with no labels act weird, but if you disable the counter tests this mostly goes away, suggesting testing artifacts. The benefits are more substantial for metrics with label (which IME is 'every singe metric you have', but may result in a slight decline in the no-label path.

However, since the no-label path has historically been **nearly 500 times faster** than the label scenario, this is not a real problem. The amortized cost of the entire system is greatly reduced by optimizing for the common case over the rare.

### What's it do?

This PR introduces a data structure that owns both the labels and the metrics that correspond with them. It encapsulates all of the lifecycle operations in a single class, which means internal implementation details can be adjusted in the future to explore further improvements. For example, moving `validateLabels` into the receiving functions and changing the key structure in arbitrary ways, including possibly turning it into an array to reduce string fragment duplication.

The brunt of this PR, besides adding lots of additional benchmarks and culling some other ones to keep the total run time not insane, is this:

- All possible keys for a metric are known at creation time
- new values can show up arbitrarily
- the store becomes a cartesian product of all values for the given keys
 
You can think of this as a matrix or a table

```
| label1        |        label2 |        label3 |
| ------------- | ------------- | ------------- | 
|      -        |      200      |     'get'     |
|    orange     |       -       |       -       |
|     blue      |      403      |       -       |
```
The keys are a sparse list of hits with `''` serving as all empty elements, resulting in the following keys for the example above:
- `|200|get` instead of `label2:200,label3:get,`
- `orange||` instead of `label1:orange,`
- `blue|403|` instead of `label1:blue,label2:403,`

In my experience the tags can be at least as long as the values, and a substantial fraction of tags that can be set will be set, so I believe this will be a net substantial reduction in memory usage.

Structured this way, we can experiment with different solutions. A tree structure? sorting keys by a different order, etc.

#### What about pipes in the data causing key collisions?

The existing system has the same problem with ':', but worse, because both the keys and values can contain colons. which is a legal character in key names in prometheus., For pipes to cause a problem here, you need two consecutive columns with repeating values, such that for instance 'a' and 'a|b' are legal values in one column, and 'c' and 'b|c' are legal values in the consecutive columns. Essentially you need to generate extra pipes with common substrings in two columns to cause the values to collide.

If this is seen as a problem, we could prometheus-escape the values before inserting them, but that will erode a little of the performance gain by causing an encode on every read operation. We could investigate other options at a later date, as the key structure is, as I mentioned already, now an internal implementation detail.

Note that this branch shares some benchmark-related commits in common with #683 and #680. If either of those merge, this PR will get a lot smaller.

### DIFF
--- a/lib/counter.js
+++ b/lib/counter.js
@@ -5,7 +5,6 @@
 
 const util = require('util');
 const { isObject, getLabels, nowTimestamp, LabelMap } = require('./util');
-const { validateLabel } = require('./validation');
 const { Metric } = require('./metric');
 const Exemplar = require('./exemplar');
 
@@ -31,9 +30,7 @@ class Counter extends Metric {
 	 * @returns {void}
 	 */
 	incWithoutExemplar(labels, value) {
-		if (isObject(labels)) {
-			validateLabel(this.labelNames, labels);
-		} else {
+		if (!isObject(labels)) {
 			value = labels;
 			labels = {};
 		}
@@ -47,6 +44,7 @@ class Counter extends Metric {
 
 		if (value === null || value === undefined) value = 1;
 
+		this.store.validate(labels);
 		this.store.setDelta(labels, value);
 	}
 
@@ -120,7 +118,7 @@ class Counter extends Metric {
 
 	remove(...args) {
 		const labels = getLabels(this.labelNames, args) || {};
-		validateLabel(this.labelNames, labels);
+		this.store.validate(labels); //TODO: this isn't really necessary is it?
 		this.store.remove(labels);
 	}
 }
--- a/lib/counter.js
+++ b/lib/counter.js
@@ -4,13 +4,7 @@
 'use strict';
 
 const util = require('util');
-const {
-	hashObject,
-	isObject,
-	getLabels,
-	removeLabels,
-	nowTimestamp,
-} = require('./util');
+const { isObject, getLabels, nowTimestamp, LabelMap } = require('./util');
 const { validateLabel } = require('./validation');
 const { Metric } = require('./metric');
 const Exemplar = require('./exemplar');
@@ -34,12 +28,10 @@ class Counter extends Metric {
 	 * Increment counter.
 	 * @param {object} labels - What label you want to be incremented
 	 * @param {number} value - Value to increment, if omitted increment with 1
-	 * @returns {object} results - object with information about the inc operation
+	 * @returns {void}
 	 */
 	incWithoutExemplar(labels, value) {
-		let hash = '';
 		if (isObject(labels)) {
-			hash = hashObject(labels, this.sortedLabelNames);
 			validateLabel(this.labelNames, labels);
 		} else {
 			value = labels;
@@ -55,9 +47,7 @@ class Counter extends Metric {
 
 		if (value === null || value === undefined) value = 1;
 
-		setValue(this.hashMap, value, labels, hash);
-
-		return { labelHash: hash };
+		this.store.setDelta(labels, value);
 	}
 
 	/**
@@ -79,17 +69,16 @@ class Counter extends Metric {
 		value = this.defaultValue,
 		exemplarLabels = this.defaultExemplarLabelSet,
 	} = {}) {
-		const res = this.incWithoutExemplar(labels, value);
-		this.updateExemplar(exemplarLabels, value, res.labelHash);
+		this.incWithoutExemplar(labels, value);
+		this.updateExemplar(labels, exemplarLabels, value);
 	}
 
-	updateExemplar(exemplarLabels, value, hash) {
+	updateExemplar(labels, exemplarLabels, value) {
 		if (exemplarLabels === this.defaultExemplarLabelSet) return;
 
-		const entry = this.hashMap.get(hash);
-		if (!isObject(entry.exemplar)) {
-			entry.exemplar = new Exemplar();
-		}
+		const entry = this.store.entry(labels);
+
+		entry.exemplar ??= new Exemplar();
 		entry.exemplar.validateExemplarLabelSet(exemplarLabels);
 		entry.exemplar.labelSet = exemplarLabels;
 		entry.exemplar.value = value ? value : 1;
@@ -101,9 +90,9 @@ class Counter extends Metric {
 	 * @returns {void}
 	 */
 	reset() {
-		this.hashMap = new Map();
+		this.store = new LabelMap(this.labelNames);
 		if (this.labelNames.length === 0) {
-			setValue(this.hashMap, 0);
+			this.store.set({}, 0);
 		}
 	}
 
@@ -117,7 +106,7 @@ class Counter extends Metric {
 			help: this.help,
 			name: this.name,
 			type: this.type,
-			values: Array.from(this.hashMap.values()),
+			values: Array.from(this.store.values()),
 			aggregator: this.aggregator,
 		};
 	}
@@ -132,19 +121,8 @@ class Counter extends Metric {
 	remove(...args) {
 		const labels = getLabels(this.labelNames, args) || {};
 		validateLabel(this.labelNames, labels);
-		return removeLabels.call(this, this.hashMap, labels, this.sortedLabelNames);
+		this.store.remove(labels);
 	}
-}
-
-function setValue(hashMap, value, labels = {}, hash = '') {
-	const entry = hashMap.get(hash);
-
-	if (entry !== undefined) {
-		entry.value += value;
-	} else {
-		hashMap.set(hash, { value, labels });
-	}
-	return hashMap;
 }
 
 module.exports = Counter;
--- a/lib/metric.js
+++ b/lib/metric.js
@@ -50,6 +50,7 @@ class Metric {
 			this.sortedLabelNames = [];
 		}
 
+		// TODO: Bad things happen when you call functions on half-initialized objects, yo
 		this.reset();
 
 		for (const register of this.registers) {
--- a/lib/metricAggregators.js
+++ b/lib/metricAggregators.js
@@ -21,23 +21,26 @@ function AggregatorFactory(aggregatorFn) {
 		const byLabels = new Grouper();
 		metrics.forEach(metric => {
 			metric.values.forEach(value => {
-				const key = hashObject(value.labels);
-				byLabels.add(`${value.metricName}_${key}`, value);
+				const key = value.metricName ?? '';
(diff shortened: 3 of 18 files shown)


lib/util.js
```diff
@@ -134,6 +136,203 @@ exports.nowTimestamp = function nowTimestamp() {
...
+ */
+class LabelMap {
+	/** @type {Set<string>} */
+	#labelNames;
+
+	/** @type {Map<string, StatsEntry>} */
+	#map = new Map();
+
+	constructor(labelNames = []) {
+		this.#labelNames = new Set(labelNames.slice().sort());
```
SimenB, line 156:
```suggestion
		this.#labelNames = new Set(labelNames);
```

is there any reason to sort when storing in a `Set`?

lib/cluster.js
```diff
@@ -165,7 +165,7 @@ function addListeners() {
 	if (listenersAdded) return;
 	listenersAdded = true;
 
-	if (cluster().isMaster) {
+	if (cluster().isPrimary) {
```
SimenB, line 168: I like this change, but should it be in this PR?

lib/metricAggregators.js
```diff
@@ -21,22 +21,26 @@ function AggregatorFactory(aggregatorFn) {
...
-			const valObj = {
-				value: aggregatorFn(values),
-				labels: values[0].labels,
-			};
-			if (values[0].metricName) {
-				valObj.metricName = values[0].metricName;
-			}
-			// NB: Timestamps are omitted.
-			result.values.push(valObj);
+		byLabels.forEach(group => {
```
SimenB, line 31: should these be `for-of`?

lib/registry.js
```diff
@@ -124,21 +131,24 @@ class Registry {
...
 
 		const promises = [];
 
 		for (const metric of this.getMetricsAsArray()) {
 			promises.push(metric.get());
 		}
 
-		const resolves = await Promise.all(promises);
+		for (const promise of promises) {
+			const item = await promise;
```
SimenB, line 146: hmm, can't this result in unhandled rejections if one promise rejects while we're `await`ing another?

maybe we should `.then` chain it and keep the `Promise.all` or extract `metrics.get()` into a separate async function that also does the label thing?

SimenB (commented):
I'm a big fan of these changes, thanks for working on them 😀

lib/registry.js
jdmarshall: We will still blow up at the end, and we are forcing the event queue to lock up entirely once the promise.all() resolves. You’ve got a few people in the issues database claiming this library is unusable because of the delays and I tend to agree. 

Degraded stats are degrades stats. You have the same problem on the upstream aggregator. Dropouts make all of the data weird and unrepresentative. The thing I like least about OpenTelemetry is it’s all-or-nothing behavior. It’s not real when you have five hundred processes sending data, and two tiers of aggregation coalescing it. Which in Node might only be a couple dozen machines due to one process per core, one aggregator sidecar/pod per machine.

lib/cluster.js
jdmarshall: I think it’s also in another PR that you just merged.  Let me rebase and see how much of this evaporates.

jdmarshall: Confirmed, this is from #680, but I needed the benchmark changes I made to make parts of that PR testable, and that was buried in the middle and I’ve been doing too much interactive rebasing on this branches as it is.

jdmarshall: @SimenB After rebasing, #683 has only 4 commits and all of them are also in this PR. I did not remember doing that. Maybe I was touching the same code and didn't want merge conflicts?

(I just fixed a couple in this one where I ended up having to touch lines that either were in the previous request or one line below, which annoys git)

jdmarshall:
Also found why the benchmarks got slow - Counter.new test was not giving Counter a Registry, so they were all collecting in the global registry.

lib/registry.js
jdmarshall: Fixed.

lib/registry.js
```diff
@@ -122,21 +129,28 @@ class Registry {
...
 			promises.push(metric.get());
 		}
 
-		const resolves = await Promise.all(promises);
+		Promise.all(promises).catch(err =>
+			console.error('Error happened in metric.get()', err),
+		);
+
+		for (const promise of promises) {
+			const item = await promise;
```
zbjornson, line 148: I'm not sure about this change. The one `await Promise.all()` also allows other work to complete, while avoiding waiting a tick for each promise here. (Or by "other work" are you talking about lines 150R-161R? I'm thinking of other functions.)

---

I regret making *all* `metric.get()` fns async in #361, as many are sync and can avoid Promise overhead entirely.

zbjornson (commented):
Nice improvements to the benchmark quality. 👍

lib/registry.js
jdmarshall: Based on my own experiences with a pair of bizarrely misbehaving unit tests, I believe v8 stopped switching context on `await(3)` either in Node 16 or 18. That’s also about when async await got as fast or faster than Bluebird. 

> I regret making all metric.get() fns async

One thing GitHub is not good at is polling users to see how badly they would be affected by changing this behavior on a major version number. But compared to the amount of Map traversal that goes into these things, I think a promise per metricName is probably the least of your worries here. It only will show up on toy projects and so what. 

I commented elsewhere that I wonder if the `(typeof Promise)` conditionals are redundant. I’m pretty sure they are at this point.

jdmarshall: In the original I thought I could avoid the overhead of the `Promise.all` entirely. I don’t think we want to pull the metric.get() down and combine the loops, as that will cause more clock skew between metrics by introducing more time between them. 

And a try/catch to handle all the leftovers would be more overhead than the promise.all.

jdmarshall: I’m going to work on #688 once this lands, while I still have time between working on other projects which are stalled by Life happening to people. That will definitely answer any concerns about this block of code being slower. 

I’ve been trying to keep too many parallel PRs running and I’m cutting back for sanity.

jdmarshall:
Update: I'm seeing slightly less rosy numbers today and cannot nail down why.  30% has become 25% and 75% has become 55%.  Thought it might have something to do with the lower max-heap-size on the final commit, but it seems not.

lib/registry.js
zbjornson: I understand. I'm talking about the scenario where many or all of the metrics are synchronous, in which case they might or will all have resolved by the time the 150-160 loop runs, and in which case the `await`s in the loop will be slower.

I'm struggling to find/remember why we made `collect()` support returning a promise. There's some discussion in #180, but even after #329 resolved that issue, all of the default metrics in this library are all still sync (except for the semi-async, [incorrect](https://github.com/siimon/prom-client/issues/561) `nodejs_eventloop_lag_seconds` that should probably be removed). https://github.com/siimon/prom-client/pull/337#issuecomment-591657430 mentions supporting async for ~niceness. But right now I can't think of a mainstream use case and see no issues requesting or mentioning it. -- Anyway, I do think async collectors are rare, meaning the promises are likely to be resolved by the time this loop runs.

jdmarshall:
I took the await change out of this PR. It will resurface as a 'niceness' PR for cluster and worker threads later.

lib/counter.js
```diff
@@ -131,20 +118,9 @@ class Counter extends Metric {
...
-}
-
-function setValue(hashMap, value, labels = {}, hash = '') {
-	const entry = hashMap.get(hash);
-
-	if (entry !== undefined) {
-		entry.value += value;
-	} else {
-		hashMap.set(hash, { value, labels });
+		this.store.validate(labels); //TODO: this isn't really necessary is it?
```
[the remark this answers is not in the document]
jdmarshall: You can’t remove what isn’t there.

lib/metric.js
```diff
@@ -50,6 +50,7 @@ class Metric {
 			this.sortedLabelNames = [];
 		}
 
+		// TODO: Bad things happen when you call functions on half-initialized objects, yo
```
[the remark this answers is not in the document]
jdmarshall: Unfortunately this is a deeper architectural problem. It would require modifying the entire inheritance tree. If this were a threaded language the blast radius would be a lot larger, but even here you have re-entrancy problems. 

Which is why the code has to double-initialize the hashMap/storage in most of the Metrics. The reset call expects the storage to be defined already but we are still in the super() call. Calling a method named 'init' from a constructor... happens. But a function called reset? I haven't even finished *setting* yet.

This is partially spit-balling, I probably should think about this longer before answering, but I suspect the solution is extract a function from reset() to handle the no-label case, call it from the constructor and drop the reset call entirely from Metric. 

It might be worth pointing out that several people have taken exception in the issues database for how labelless stats are initialized. But I haven't looked at them long enough to be sure if that's a "change the default value" or "remove the special case entirely". If it's the latter then the correct answer is "delete the code"

lib/counter.js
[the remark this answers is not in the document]
jdmarshall: As an aside, I've gone back and forth on where the responsibility for validate should lie.

As I was writing this PR, I had the thought, "You know as a follow-on maybe I should fold the validate call into LabelMap.set/setDelta". And that lasted until I got to the `labels()` methods, and the timer methods in one/a couple of the metrics.

Strictly speaking, we could validate the labels in the labels call, and then skip them for the functions attached to the object returned by these functions. However the whole inc/incWithExemplars, observe/observeWithExemplars trick would make that difficult to unravel. I think there are a lot of other concerns being juggled here and while there's an opportunity to make correctness and performance not fight by inverting some of the code here. 

It's probably not impossible, but my my head is pretty full trying to keep the PRs I already have ready to go, straight.

zbjornson (approved):
Looks good!

zbjornson:
Sorry @jdmarshall, another PR was merged and created a small conflict. If @siimon is okay with it, I think we should remove the changelog reminder and have mergers maintain the changelog instead.

Once that's fixed and @SimenB is good with the revisions, happy to merge.

jdmarshall:
I fixed the changelog

jdmarshall:
Thank you!

PR #689, opened 2025-07-11
feat: add option to exclude specific default metrics from collection

### LINKED ISSUE #623: Provide exclusions for default metrics collected

kwaszczuk:
Fixes: https://github.com/siimon/prom-client/issues/623 (I believe it was incorrectly marked as fixed)

This PR adds `exclude` option to the `DefaultMetricsCollectorConfiguration` that allows to omit certain group of metrics from being collected.

My own use case is that I use a custom implementation of `eventLoopLag` metrics (though based on the original one), so I need to exclude them to avoid conflicts in registry, but others may want to also use it to remove redundant and unused metrics.

An alternative approach would be to expose the exports from `lib/metrics` in the public interface, allowing users to import and manually register only the specific metrics they need. I’m not sure, though, which approach is better - this one or the `exclude` approach.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -767,6 +767,21 @@ export function exponentialBuckets(
 	count: number,
 ): number[];
 
+type AvailableDefaultMetrics =
+	| 'processCpuTotal'
+	| 'processStartTime'
+	| 'osMemoryHeap'
+	| 'processOpenFileDescriptors'
+	| 'processMaxFileDescriptors'
+	| 'eventLoopLag'
+	| 'processResources'
+	| 'processHandles'
+	| 'processRequests'
+	| 'heapSizeAndUsed'
+	| 'heapSpacesSizeAndUsed'
+	| 'version'
+	| 'gc';
+
 export interface DefaultMetricsCollectorConfiguration<
 	T extends RegistryContentType,
 > {
@@ -775,6 +790,7 @@ export interface DefaultMetricsCollectorConfiguration<
 	gcDurationBuckets?: number[];
 	eventLoopMonitoringPrecision?: number;
 	labels?: object;
+	exclude?: AvailableDefaultMetrics[];
 }
 
 export const collectDefaultMetrics: {
@@ -786,7 +802,7 @@ export const collectDefaultMetrics: {
 		config?: DefaultMetricsCollectorConfiguration<T>,
 	): void;
 	/** All available default metrics */
-	metricsList: string[];
+	metricsList: AvailableDefaultMetrics[];
 };
 
 /**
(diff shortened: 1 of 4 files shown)


jdmarshall (approved):
LGTM.

@zbjornson @SimenB how do we feel about this one?

jdmarshall (changes requested):
Needs a changelog entry.

zbjornson:
LGTM too. Either approach (include or exclude) seems reasonable.

> Needs a changelog entry.

Tangent, but I've wondered if we should remove this requirement and instead make it part of the maintainers' release process. It always creates merge conflicts when any other PR is merged.

cjihrig:
> Tangent, but I've wondered if we should remove this requirement and instead make it part of the maintainers' release process. It always creates merge conflicts when any other PR is merged.

FWIW, the OpenTelemetry project also requires contributors to add changelog entries with PRs and it is a massive pain when trying to contribute. Big +1 to removing that requirement.

jdmarshall:
Not a fan either. I had a lot of PRs running in parallel when I first started contributing and I ended up conflicting with my own PRs based on what order they got landed.  I think I'd successfully convinced SimenB it was an issue but not what to do about it.

jdmarshall:
This will need an additional change now that I landed #785

jdmarshall:
@kwaszczuk are you available for assistance with updating this PR? If not I can take a look at it after 0.16 lands.

kwaszczuk:
It has been a long while, but I'll try in the upcoming days.

jdmarshall:
Cool cool. I believe only #785 needs to be worked in, though a changelog entry would also be good. Thanks.

kwaszczuk:
@jdmarshall I've updated the PR including addition of the changelog entry,

jdmarshall:
Some ideas for additional default metrics are coming up, so I'd like to have this one on the books as a dodge "well you can always turn them off if you hate them." before we start considering them for inclusion. I think matching the ordering in #810 will make this process smoother.

I'd like to land this PR somewhere between 4th and 6th in merge order so you've got a couple days before I will ping you again.

index.d.ts
```diff
@@ -865,6 +865,22 @@ export function exponentialBuckets(
...
+	| 'eventLoopLag'
+	| 'eventLoopUtilization'
+	| 'processResources'
+	| 'processHandles'
+	| 'processRequests'
+	| 'heapSizeAndUsed'
+	| 'heapSpacesSizeAndUsed'
+	| 'version'
+	| 'gc';
+
```
jdmarshall, line 883: Can you order these to match #810?

PR #696, opened 2025-07-31
feat: add constructor with content type param to Registry class in Ty…

us-asad:
…peScript definitions


jdmarshall:
@us-asad you've got a couple linter errors do deal with.

jdmarshall (changes requested):
The linter is angry at you but otherwise I believe this is correct.

jdmarshall (approved):

PR #705, opened 2025-08-20
Performance: simplified checking for reserved label keyword

Connormiha:
`this.labelNames` is array of string, so we can just use `includes` instead of `for of`.
Since this code was still throwing an error, I moved it up to ensure it completes earlier. The class instance still wouldn’t be created anyway, I moved it up.

### DIFF
--- a/lib/summary.js
+++ b/lib/summary.js
@@ -19,6 +19,9 @@ class Summary extends Metric {
 			store: new LabelMap(),
 		});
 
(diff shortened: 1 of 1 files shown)


SimenB (approved):

SimenB:
I'd be surprised if there's a benchmark change beyond microbenchmarks, but early returns before creating `LabelMap` seems like a good idea to me 😀

PR #732, opened 2025-11-28
Update to 1.0 version of faceoff.

jdmarshall:
bench-node landed my PR to expose output formatting without console.log, and I am now using that instead of monkey patching console.


jdmarshall:
I'm working on a 1.1 with some bug fixes in it.

PR #736, opened 2025-12-22
Upgrade Faceoff and enable some more tests on trunk.

jdmarshall:
A number of the benchmarks I added are now merged into trunk and thus can be run to compare PRs to pre-released versions of the code.


jdmarshall:
@SimenB I simplified benchmark naming in a later minor version, and that'll work better here for the rename at any rate.

jdmarshall:
This new version also color codes performance deltas and flags inconclusive results due to mean vs std deviation artifacting.

And is easier to read on narrower screens (much smaller output format so the punchlines don't scroll off the right edge of the screen as much)

jdmarshall:
Hey @zbjornson can we land this?

zbjornson:
Sorry @jdmarshall, I was OOO for the last week.

PR #727, opened 2025-11-19
Perf improvements

mcollina:
This PR does _a lot_ of modernization and updates, and it also vendors critical dependencies because they were bottlenecks; I didn't want to fork them into individual repositories. The goal of those changes is to significantly reduce the overhead of prom-client and make it easier to maintain/test.

```
⚠ counter ➭ new is 8.628% acceptably slower.
⚠ counter ➭ inc is 4.248% acceptably slower.
✓ counter ➭ inc with labels is 60.66% faster.
⚠ gauge ➭ inc is 1.275% acceptably slower.
✓ gauge ➭ inc with labels is 112.6% faster.
✓ histogram ➭ observe#1 with 64 is 91.29% faster.
✓ histogram ➭ observe#2 with 8 is 68.05% faster.
✓ histogram ➭ observe#2 with 4 and 2 with 2 is 35.47% faster.
✓ histogram ➭ observe#2 with 2 and 2 with 4 is 34.35% faster.
✓ histogram ➭ observe#6 with 2 is 40.55% faster.
✓ histogram ➭ startTimer#1 with 64 is 23.84% faster.
✓ histogram ➭ startTimer#2 with 8 is 27.74% faster.
✓ histogram ➭ startTimer#2 with 4 and 2 with 2 is 18.38% faster.
✓ histogram ➭ startTimer#2 with 2 and 2 with 4 is 20.34% faster.
✓ histogram ➭ startTimer#6 with 2 is 25.85% faster.
✓ summary ➭ observe#1 with 64 is 61.16% faster.
✓ summary ➭ observe#2 with 8 is 56.47% faster.
✓ summary ➭ observe#2 with 4 and 2 with 2 is 26.87% faster.
✓ summary ➭ observe#2 with 2 and 2 with 4 is 28.96% faster.
✓ summary ➭ observe#6 with 2 is 41.48% faster.
✓ registry ➭ getMetricsAsJSON() no labels is 29.44% faster.
✓ registry ➭ getMetricsAsJSON() 1 x 64 is 14.81% faster.
✓ registry ➭ getMetricsAsJSON() 2 x 4 is 7.147% faster.
✓ registry ➭ getMetricsAsJSON() 2 x 8 is 12.15% faster.
✓ registry ➭ getMetricsAsJSON() 6 x 2 is 11.09% faster.
✓ registry ➭ getMetricsAsJSON() 2 x 4, 2 defaults is 13.37% faster.
[...]

PR #692, opened 2025-07-29
2.4x improvement in Cluster aggregation performance

jdmarshall:
Use a similar strategy to LabelMap for stats coalescing for cluster.js
    
This has to wrestle with the problem of lack of fore-knowledge of the
canonical set of labels. We achieve this by assuming 6 labels to
begin with, first filling in the empties as we discover new ones, and
tnen expanding the columns by ⌈25%⌉ if they don't, and re-keying the
entire table (by concatenating extra ||'s onto the end)..

```    
cluster ⇒ aggregate()

Summary (vs. baseline):
 ⇒ prom-client@latest            | ██████████─────────────── | 4.18 ops/sec | 11 samples (baseline)
 ⇒ prom-client@trunk             | ██████████▌────────────── | 4.26 ops/sec | 11 samples (1.02x faster)
 ⇒ prom-client@current           | █████████████████████████ | 10 ops/sec | 11 samples (2.40x faster)
```

Issue #671, opened 2025-06-26
On optimizing histogram findBounds()


jdmarshall:
Apologies if this has become word salad. This was the contents of my morning walk and some mental exercises last night.

There's a histogram problem that OTEL figured out a couple years ago, and that's the linear bucket scan in the observe() logic. As you increase the number of buckets per label set you increase hosting costs but also collection time. And if you try to generalize your bucket sizes over different classes of traffic, (eg KV store access and write timing and overall response times with the same bucket sets), you can make a real mess. 

They solved the insertion problem with a binary search. I'm not convinced that is the best solution to the problem. It might be worth considering other options as any changes to observe() result in diminishing returns if another better option comes along later.

The most obvious problem with BST is that we aren't looking for an equality, we are looking for an inequality. So we need to find the leftmost bucket that is greater than the our sample value. We are in effect looking for a 'between' rather than a match. So when you get close you need to look at one bucket and its subsequent and next bucket. A binary search is fine for that but other data structures like sorted heaps less so.

Secondly, insertions are not uniformly distributed, they are normally distributed. Most bucket sets I have thus far encountered represent a geometric progression of values. Most of the time every second bucket is twice the bound, and the bucket between them is the average of the two, resulting in an effective exponent of around 1.4. This geometric distribution of bucket bounds leads to some buckets having most of the entries and the outliers having few or sometimes zero. Most of those extra buckets are about making sure the p95 and p99 statistics don't artifact so badly that they look like hallucinations.

... except under heavy load most of these bets are off. Histogram cost is proportional to the samples per second, so when server load goes up due to traffic rather than request cost, a bigger and bigger slice of the CPU time is spent on telemetry instead of retiring requests. Timers being by far the worst (Little's Law), but bucket lookups can also cost L1 cache.

We would be doing the users a service if findBounds required fewer lookups per sample, and particularly for slightly larger buckets than the long-term mean, because as system contention increases the p50 response times will shift right, looking more like the typical the p75 response times. Or p90. So ideally the most popular bucket should be quickest to find and the next ±2 buckets are nearly as efficient.

That suggests a priority balanced Treap, which would be simple enough to implement given that bucket.count is exactly the priority of each bucket. But I don't know how many users would configure enough buckets for that to win out over binary search. 

The simplest thing that could work would be to track the n/2 bucket, and do a linear scan of the first or second half of the bucket set based on value <= middle.upperBound. Or the n/3 bucket, if the long tail pattern tends to hold.

[...]

PR #738, opened 2026-01-03
Mcollina perf improvements migration

jdmarshall:
This represents a cherry-pick of the changes in #727 with a few modifications to separate the testing changes and account for PRs of mine that already merged. 

In most cases I preserved @mcollina 's commits, in some cases with deletions to split problem code from otherwise serviceable code.

Unfortunately this has removed:

- The Commit to migrate the tests. This turned out to be illuminating because dropping it turned up multiple regressions
- CLAUDE.md
- the tdigest changes have a net negative effect on the most commonly called function: `push()` - false economy
- the findBounds() changes do not satisfy the unit tests on trunk, and are setting the wrong fields (fencepost error?)
- the keyFrom changes are generating '|undefined|' values 
  - this undermines the memory improvements of the previous code in favor of microbenchmarks
  - I only seeing a few % performance improvement in the microbenchmarks, and the space tradeoff will eat those
- osMemoryHeapLinux.js is ignoring the inline comment claiming this is a non-blocking operation, and depends on the order in which the stats are evaluated and thus may be brittle
- The LabelMap key sorting change is breaking a test.
  - If this class is not being exposed in the public API, that may be okay. I need to double check. This should be in its own PR if true. But I'm unclear how much it's actually helping.
  - It does improve constructor time but those are not actually fired that often, and we trade off a potential data sharing problem.  
  - Removing list comprehensions is not a substantial performance improvement and a substantial reduction in code legibility. Dropped or chopped several commits which simply changed to for loops.
 - The async optimization includes code that assumes that collect() calls are made in a certain order. The rest of this could also be a separate PR but I didn't want to tangle with it.

## Conclusion

I think there are two or three commits that warrant their own pull requests to discuss the merits of each as a separate commit. Migrating out of Jest is interesting but not something that should be done by Claude as it is clearly not doing any red-green-refactoring on the tests to validate that they are still testing anything. Also there are 5 more months of Node 20 to deal with before `describe()` works. And asserts have the worst DX of any matcher library. I would use chai.expect instead (and have on other projects). 

What is retained here are the things that have no obvious bugs and improved the code quality, regardless of whether they had much net effect on performance.

### optimizing for empty labels.

This is a bad plan.

Stats with no labels are already many times faster than stats *with* labels (though the benchmarks misreport it as 600x, it's much smaller), and before I did the label code rework that multiplier was much higher. Since prometheus really is fairly useless with stats with zero labels (not even server name or app version??), there is absolutely no value in my mind in making 0-labels 25% faster if it makes stats with labels 1% slower. You will have thousands of the latter and a handful of the former.

I looked at this several times and opted to go the opposite direction to speed up labels, which the release version is IIRC more like 900x slower. Matteo did not quite have all of these changes when he started his. 

## Net effect

Once you factor in some changes that Matteo noticed but which were already in my other pull requests that have since merged, removing code with bugs, and importantly **after you switch to using trunk as the baseline to eliminate cross-branch confusion**,  the net effect here is small. But these are the changes that seemed like they might help in real world scenarios instead of just in microbenchmarks.

Final outcome, with latest removed to reduce confusion. Mostly in the noise floor with a few possible % here and there.

```
counter ⇒ new

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████─▕ 927,137 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 947,534 ops/sec | 14 samples (1.02x faster)

counter ⇒ inc

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏█████████████████████████▕ 15,241,474 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏████████████████████████─▕ 14,653,285 ops/sec | 12 samples (1.04x slower)

counter ⇒ inc with labels

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏█████████████████████████▕ 25,284 ops/sec |  9 samples (baseline)
 ⇒ prom-client@current       ▏████████████████████████▌▕ 25,071 ops/sec | 10 samples (1.01x slower)

gauge ⇒ inc

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏███████████████████████▌─▕ 16,650,650 ops/sec |  8 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 17,447,399 ops/sec | 13 samples (1.05x faster)

gauge ⇒ inc with labels

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏█████████████████████████▕ 140,215 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏████████████████████████▌▕ 139,564 ops/sec | 11 samples (1.00x slower)

histogram ⇒ observe#1 with 64

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏███████████████████████▌─▕ 124,410 ops/sec | 11 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 130,695 ops/sec | 12 samples (1.05x faster)

histogram ⇒ observe#2 with 8

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏█████████████████████████▕ 97,870 ops/sec | 11 samples (baseline)
 ⇒ prom-client@current       ▏████████████████████████─▕ 94,061 ops/sec | 10 samples (1.04x slower)

histogram ⇒ observe#2 with 4 and 2 with 2

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 51,122 ops/sec |  9 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 51,717 ops/sec | 11 samples (1.01x faster)

histogram ⇒ observe#2 with 2 and 2 with 4

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 50,296 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 50,696 ops/sec | 10 samples (1.01x faster)

histogram ⇒ observe#6 with 2

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏█████████████████████████▕ 39,667 ops/sec | 11 samples (baseline)
 ⇒ prom-client@current       ▏████████████████████████▌▕ 39,603 ops/sec | 10 samples (1.00x slower)

histogram ⇒ startTimer#1 with 64

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 60,210 ops/sec | 11 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 61,351 ops/sec | 10 samples (1.02x faster)

histogram ⇒ startTimer#2 with 8

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 52,742 ops/sec | 11 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 53,690 ops/sec | 10 samples (1.02x faster)

histogram ⇒ startTimer#2 with 4 and 2 with 2

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 34,852 ops/sec |  9 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 35,461 ops/sec | 11 samples (1.02x faster)

histogram ⇒ startTimer#2 with 2 and 2 with 4

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 34,517 ops/sec |  9 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 34,907 ops/sec | 11 samples (1.01x faster)

histogram ⇒ startTimer#6 with 2

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████─▕ 29,454 ops/sec | 11 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 30,463 ops/sec |  9 samples (1.03x faster)

util ⇒ hashObject

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏█████████████████████████▕ 4,261,424 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏████████████████████████▌▕ 4,208,851 ops/sec |  8 samples (1.01x slower)

util ⇒ LabelMap.validate()

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏█████████████████████████▕ 16,797,115 ops/sec | 12 samples (baseline)
 ⇒ prom-client@current       ▏████████████████████████▌▕ 16,512,785 ops/sec | 10 samples (1.02x slower)

util ⇒ LabelMap.keyFrom()

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 7,209,639 ops/sec | 13 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 7,227,734 ops/sec | 12 samples (1.00x faster)

summary ⇒ observe#1 with 64

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏█████████████████████████▕ 105,703 ops/sec | 11 samples (baseline)
 ⇒ prom-client@current       ▏████████████████████████▌▕ 103,763 ops/sec | 10 samples (1.02x slower)

summary ⇒ observe#2 with 8

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 80,878 ops/sec |  9 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 81,213 ops/sec | 10 samples (1.00x faster)

summary ⇒ observe#2 with 4 and 2 with 2

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏█████████████████████████▕ 46,491 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏████████████████████████▌▕ 45,926 ops/sec | 10 samples (1.01x slower)

summary ⇒ observe#2 with 2 and 2 with 4

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏█████████████████████████▕ 44,956 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏████████████████████████▌▕ 44,600 ops/sec | 11 samples (1.01x slower)

summary ⇒ observe#6 with 2

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 36,361 ops/sec | 11 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 36,595 ops/sec | 10 samples (1.01x faster)

registry ⇒ getMetricsAsJSON() no labels

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏█████████████████████████▕ 427,999 ops/sec | 11 samples (baseline)
 ⇒ prom-client@current       ▏████████████████████████─▕ 416,444 ops/sec | 11 samples (1.03x slower)

registry ⇒ getMetricsAsJSON() 1 x 64

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 8,138 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 8,211 ops/sec |  9 samples (1.01x faster)

registry ⇒ getMetricsAsJSON() 2 x 4

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏█████████████████████████▕ 25,853 ops/sec | 11 samples (baseline)
 ⇒ prom-client@current       ▏████████████████████████▌▕ 25,838 ops/sec |  9 samples (1.00x slower)

registry ⇒ getMetricsAsJSON() 2 x 8

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏█████████████████████████▕ 7,343 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏████████████████████████▌▕ 7,286 ops/sec | 10 samples (1.01x slower)

registry ⇒ getMetricsAsJSON() 6 x 2

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 4,646 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 4,689 ops/sec | 10 samples (1.01x faster)

registry ⇒ getMetricsAsJSON() 2 x 4, 2 defaults

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 16,838 ops/sec | 11 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 17,069 ops/sec | 10 samples (1.01x faster)

registry ⇒ getMetricsAsJSON() 2 x 2, 4 defaults

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏█████████████████████████▕ 66,602 ops/sec |  9 samples (baseline)
 ⇒ prom-client@current       ▏████████████████████████▌▕ 66,161 ops/sec | 11 samples (1.01x slower)

registry ⇒ metrics() no labels

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 166,510 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 169,287 ops/sec | 10 samples (1.02x faster)

registry ⇒ metrics() no labels and openMetrics

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████─▕ 161,218 ops/sec | 11 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 165,202 ops/sec | 12 samples (1.02x faster)

registry ⇒ metrics() 1 x 64

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████─▕ 2,840 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 2,921 ops/sec |  9 samples (1.03x faster)

registry ⇒ metrics() 1 x 64 and openMetrics

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████─▕ 2,833 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 2,923 ops/sec |  9 samples (1.03x faster)

registry ⇒ metrics() 2 x 4

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 10,901 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 11,082 ops/sec | 10 samples (1.02x faster)

registry ⇒ metrics() 2 x 4 and openMetrics

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏█████████████████████████▕ 10,829 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏████████████████████████▌▕ 10,662 ops/sec |  9 samples (1.02x slower)

registry ⇒ metrics() 2 x 8

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████─▕ 3,110 ops/sec |  9 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 3,218 ops/sec | 10 samples (1.03x faster)

registry ⇒ metrics() 2 x 8 and openMetrics

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 3,139 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 3,157 ops/sec |  9 samples (1.01x faster)

registry ⇒ metrics() 6 x 2

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████─▕ 2,449 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 2,544 ops/sec | 10 samples (1.04x faster)

registry ⇒ metrics() 6 x 2 and openMetrics

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████─▕ 2,443 ops/sec |  9 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 2,528 ops/sec | 10 samples (1.03x faster)

registry ⇒ metrics() 2 x 4, 2 defaults

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████▌▕ 6,914 ops/sec |  9 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 6,961 ops/sec | 10 samples (1.01x faster)

registry ⇒ metrics() 2 x 4, 2 defaults and openMetrics

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████─▕ 6,906 ops/sec | 10 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 7,102 ops/sec | 10 samples (1.03x faster)

registry ⇒ metrics() 2 x 2, 4 defaults

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████─▕ 26,211 ops/sec | 11 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 26,875 ops/sec |  8 samples (1.03x faster)

registry ⇒ metrics() 2 x 2, 4 defaults and openMetrics

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████─▕ 26,328 ops/sec | 11 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 27,376 ops/sec |  9 samples (1.04x faster)

cluster ⇒ aggregate()

Summary (vs. baseline):
 ⇒ prom-client@trunk         ▏████████████████████████─▕ 8.46 ops/sec | 11 samples (baseline)
 ⇒ prom-client@current       ▏█████████████████████████▕ 8.64 ops/sec | 11 samples (1.02x faster)


No significant regressions found.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Expanded benchmarking code
+- Expanded benchmarks and using new benchmarking library
 - new WorkerRegistry to provide equivalent support to AggregatorRegistry
 
 ## [15.1.3] - 2024-06-27
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - ci: switch out deprecated benchmark-regression library for replacement
 - AggregatorRegistry renamed to ClusterRegistry, old name deprecated
 - chore: update faceoff to 1.1
+- perf: Avoid array conversion in getMetricsAsJSON by directly iterating over metric values (~1.3% faster)
+- perf: Optimize string escaping for better metrics serialization
 
 ### Added
 
--- a/benchmarks/defaultMetrics.js
+++ b/benchmarks/defaultMetrics.js
@@ -1,0 +1,58 @@
+'use strict';
+
+const Path = require('path');
+const { createRequire } = require('node:module');
+const { getLabelCombinations } = require('./utils/labels');
+
+let count = 1;
+
+module.exports = function setupSuites(benchmark) {
+	benchmark.suite('osMemoryHeap', suite => {
+		suite.add(
+			'new',
+			(client, { labels, metric: osMemoryHeap, registry }) =>
+				osMemoryHeap(registry, {
+					prefix: `heap${count++}`,
+					help: 'HeapUsed',
+					labels,
+				}),
+			{
+				setup: (client, location) =>
+					loadMetrics(client, location, 'osMemoryHeap'),
+				teardown,
+			},
+		);
+
+		suite.add('collect', async (client, { registry }) => registry.metrics(), {
+			setup: (client, location) => {
+				const ctx = loadMetrics(client, location, 'osMemoryHeap');
+				const { metric: osMemoryHeap, labels, registry } = ctx;
+
+				osMemoryHeap(registry, {
+					prefix: `heap${count++}`,
+					help: 'HeapUsed',
+					labels,
+				});
+
+				return { registry };
+			},
+			teardown,
+		});
+	});
+};
+
+function loadMetrics(client, location, metricName) {
+	const require = createRequire(location);
+	const fromModule = Path.join(location, `./lib/metrics/${metricName}`);
+	const combinations = getLabelCombinations([1], ['region', 'env']);
+
+	return {
+		metric: require(fromModule),
+		labels: combinations[0],
+		registry: new client.Registry(),
+	};
+}
+
+function teardown(client, { registry }) {
+	registry.clear();
+}
--- a/benchmarks/index.js
+++ b/benchmarks/index.js
@@ -24,6 +24,7 @@ benchmarks.suite('histogram', require('./histogram'));
 benchmarks.suite('util', require('./util'));
 benchmarks.suite('summary', require('./summary'));
 benchmarks.suite('registry', require('./registry'));
+benchmarks.suite('default metrics', require('./defaultMetrics'));
 benchmarks.suite('cluster', require('./cluster'));
 
 benchmarks
--- a/benchmarks/util.js
+++ b/benchmarks/util.js
@@ -5,7 +5,7 @@ const Path = require('path');
 module.exports = setupUtilSuite;
 
 function setupUtilSuite(suite) {
-	const skip = ['prom-client@latest', 'prom-client@trunk'];
+	const skip = ['prom-client@latest'];
 
 	suite.add(
 		'hashObject',
--- a/index.js
+++ b/index.js
@@ -38,4 +38,3 @@ exports.ClusterRegistry = require('./lib/cluster');
 exports.WorkerRegistry = require('./lib/worker');
 /** @deprecated */
 exports.AggregatorRegistry = exports.ClusterRegistry;
-exports[Symbol('util')] = require('./lib/util');
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -137,7 +137,7 @@ class Registry {
 
 		const promises = [];
 
-		for (const metric of this.getMetricsAsArray()) {
+		for (const metric of this._metrics.values()) {
 			promises.push(metric.get());
 		}
 
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -293,11 +293,46 @@ function escapeLabelValue(str) {
 	if (typeof str !== 'string') {
 		return str;
 	}
-	return escapeString(str).replace(/"/g, '\\"');
+
+	let result = '';
+	for (let i = 0; i < str.length; i++) {
+		const char = str[i];
+		switch (char) {
+			case '\\':
+				result += '\\\\';
+				break;
+			case '\n':
+				result += '\\n';
+				break;
+			case '"':
+				result += '\\"';
+				break;
+			default:
+				result += char;
+		}
+	}
+
+	return result;
 }
+
 function escapeString(str) {
-	return str.replace(/\\/g, '\\\\').replace(/\n/g, '\\n');
+	let result = '';
+	for (let i = 0; i < str.length; i++) {
+		const char = str[i];
+		switch (char) {
+			case '\\':
+				result += '\\\\';
+				break;
+			case '\n':
+				result += '\\n';
+				break;
+			default:
+				result += char;
+		}
+	}
+	return result;
 }
+
 function standardizeCounterName(name) {
 	return name.replace(/_total$/, '');
 }


jdmarshall:
Additional notes:

TDigest.push() is consistently testing slower but all of the other functions are faster. I don't know if that means it should be in, or out. 

Eliminating the getMetricsAsArray() call doesn't have a measurable improvement in performance, but it does reduce allocations so I retained it. 

Empty Set checks also tangle with another PR I already got merged, so that gets watered down too.

Update: After further consideration, push() is the hot path and making it slower is a bad idea. Dropped.

jdmarshall:
A silver lining of this effort is that while trying to conserve the bintree and tdigest benchmarks @mcollina wrote, I ported them to faceoff so that we could get deltas between branches. That effort surfaced a rather substantial ergonomics problem with subsuites and I've opened a ticket to make that better.

https://github.com/cobblers-children/faceoff/issues/32

mcollina:
> - the tdigest changes have a net negative effect on the most commonly called function: `push()` - false economy

I do not recall exactly, but I would need to analyze again.


> - the findBounds() changes do not satisfy the unit tests on trunk, and are setting the wrong fields (fencepost error?)

This is actual one of the most significant optimizations.

> - the keyFrom changes are generating '|undefined|' values 

Yes. This is critical to improve performance.

>   - this undermines the memory improvements of the previous code in favor of microbenchmarks

Most of those changes were measured in real apps as well.

> - osMemoryHeapLinux.js is ignoring the inline comment claiming this is a non-blocking operation, and depends on the order in which the stats are evaluated and thus may be brittle

It was showing up in the flamegraph as a hot operation. Moving to async it made it disappear. Easy choice.

> - The LabelMap key sorting change is breaking a test.

This was needed to fix the keyFrom changes.

>   - Removing list comprehensions is not a substantial performance improvement and a substantial reduction in code legibility. Dropped or chopped several commits which simply changed to for loops.

We are not in agreement.

>  - The async optimization includes code that assumes that collect() calls are made in a certain order. The rest of this could also be a separate PR but I didn't want to tangle with it.

This is a critical optimization.

jdmarshall:
I would encourage you to file findBpunds() as its own PR, but with pinning tests and fixing the issues with the existing tests. Which you didn’t see to be running as you went along?  61 files is too much for a single PR. And red builds cannot be merged regardless.

I agree that it’s an important change. I wrote #671.

> It was showing up in the flamegraph as a hot operation. Moving to async it made it disappear. Easy choice.

I will have to test this and get back to you. You also have to be careful you aren’t triggering dead code elimination in benchmarks but that’s probably not what is happening here. 

> Yes. This is critical to improve performance [of keyFrom]

`keyFrom()` as you see it was written by me. The changes I made were to answer complaints of high memory usage by prom-client. Yes it reduces string interpolation operations and so is a net performance improvement. Expanding the key names again undoes *my* most important contribution to this code base and really makes the whole change about as bad if not worse than the original. You are missing that I had already optimized the empty check in another PR that has now landed2 and that erodes most of the gains made in the rest of the function. 

> for loop

I’ve tested for loops versus list comprehensions extensively in NodeJS. Particularly under benchmark.js, in production code, in production batch processing, and with tinybench. Yes they used to be quite slow especially in hot paths. Part of that has been fixed in later v8 versions. Part of it is fixed by using arrow functions. Part of it is fixed by not touching `arguments`. It ends up being swamped by the complexity of the operation inside the loop, especially if it ends up touching a lot of data. Particularly if you limit your discussion to Node 20 and later. It’s true that under 16 and 14 this gap was wider. But V8 is pretty good at this stuff. 

If the loop code were having a large effect you would see it show up under the aggregate tests like summary and registry and I’m just not seeing it under bench-node. If you use trunk as the baseline instead of latest, it’s easier to see what’s new versus whats already landed. 

My MO is to make all of the performance optimizations that also improve the code quality, regardless of how well or poorly they benchmark. 2% gains done a dozen times is still a 25% improvement and I will take that with few caveats.  Nobody can really argue with better code even if the 2% seems like a waste of my time. What matters though is how much QA is necessary to validate the changes and for that you restrict your changes to one functional area at a time instead of the entire codebase. 

As a general rule I avoid `reduce()` and `forEach()`. Without the transformed array in the intermediate step, the performance gap starts to matter more, and the legibility advantage becomes suspect. `reduce` is for legibility reasons. The noun of the reduce verb is lost at the end of the action and I’ve seen it lead to reading comprehensions problems. Except in Elixir, where the argument order is reversed.

In batch processing unrolling a map() to a `for of` can improve throughput and throttle fanout in async functions. But that’s much more about network latency and congestion. It’s fundamentally changing the order of operations. 

https://benchmarklab.azurewebsites.net/Benchmarks/Show/34274/1/for-in-loop-vs-map#latest_results_block be sure to look at only the Chrome results. Since that’s what’s relevant to nodeJS performance. But I’m seeing similar results in Safari which would cover Bun. 

Stats should ideally be 5% of your application workload and no more. So 2% of 5% is lost in the noise floor. I had a large code base where the cross cutting concerns, including stats, was closer to 20%, and I did enough work to find 90 ms in response time. I didn’t have to murder any `map()` calls to get there. Well, one, but that was merging two maps into one. I did have to upgrade nodeJS for that to be true though.

jdmarshall:
As far as I’m aware prom-client makes no guarantees about what order stats will be evaluated in and in fact #692 alters the order they are collected and achieves a 2.4x improvement in aggregation time as part of that. The intermittent even queue stalls are perhaps more troubling to users.

jdmarshall:
On the topic of osMemoryHeap:

The problem with readFile is that it blocks the event queue. There's a big comment in the code about why that's okay. 

The alternative is to call `fs/promises.readFile`, which doesn't block but is also orders of magnitude slower. When you're trying to juggle a bunch of operations in a busy NodeJS container that's generally a good thing, but the penalty for doing so is extremely high latency.

Running the benchmark in a docker container:

> Summary (vs. baseline):
>   default metrics ⇒ osMemoryHeap ⇒ collect ⇒ prom-client@trunk    (baseline)
>   default metrics ⇒ osMemoryHeap ⇒ collect ⇒ prom-client@current  **(17.78x slower)**

So the wall clock time goes from 12 µs to 227 µs, which is not terrible if it reduces blocking for other operations. The user and system time are both about the same that feels like the amount of effort is about even, it's just that benchmarks don't work well for this sort of async code. 

What I'm going to do is include the benchmarks I wrote in this branch, but we can discuss the osMemoryHeapLinux changes as its own PR. Might be worth doing that at the same time the async collect changes. To see if it helps or exacerbates the situation.

PR #740, opened 2026-01-09
Fixed OpenMetrics Content-Type support for `cluster` module

gslandtreter:
There's a call to `Registry.aggregate` that does not pass in any value to the register content-type parameter. This causes the cluster/worker aggregator to always return `PROMETHEUS_CONTENT_TYPE` formatted metrics, instead of `OPENMETRICS_CONTENT_TYPE`, even when the latter is used when constructing the original `AggregatorRegistry`.

This commit fixes support for `Registry.OPENMETRICS_CONTENT_TYPE` by passing the AggregatorRegistry's content-type into the call to `Registry.Aggregate` within the cluster metrics response listener.

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -28,7 +28,7 @@ const requests = new Map(); // Pending requests for workers' local metrics.
 class AggregatorRegistry extends Registry {
 	constructor(regContentType = Registry.PROMETHEUS_CONTENT_TYPE) {
 		super(regContentType);
-		addListeners();
+		addListeners(regContentType);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


jdmarshall:
Can we get a test or three for this?

jdmarshall (changes requested):
Some tests would be good

jdmarshall:
@gslandtreter I'm open to landing this in the next release but there's only one other PR I want to get into it and then we're ready to go.  You can refer to the existing TS tests for examples.

jdmarshall:
Unfortunately my changes for shutdown() now conflict quite a bit with this PR.

Instead the `#gather()` function needs the change.

PR #742, opened 2026-03-15
perf: use the faster api to collect rss

chidam333:
https://nodejs.org/api/process.html#processmemoryusagerss

i think we can use the faster api to collect rss

### DIFF
--- a/lib/metrics/helpers/safeRss.js
+++ b/lib/metrics/helpers/safeRss.js
@@ -1,0 +1,12 @@
+'use strict';
+
+// process.memoryUsage.rss() can throw on some platforms, see #67
+function safeRss() {
+	try {
+		return process.memoryUsage.rss();
+	} catch {
+		return;
+	}
+}
+
+module.exports = safeRss;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - perf: Improve performance of registry defaultLabels during metric processing
 - perf: New, more space-efficient storage engine, 20-45% faster stats recording
 - perf: Further improvement to key generation cost
+- perf: Use faster `process.memoryUsage.rss()` API for resident memory collection
 - fix: Browser compatibility for Gauge.startTimer()
 - ci: Run benchmarks for pull requests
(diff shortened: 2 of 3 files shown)


lib/metrics/helpers/safeRss.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+// process.memoryUsage.rss() can throw on some platforms, see #67
+function safeRss() {
+	try {
+		return process.memoryUsage.rss();
+	} catch {
+		return;
```
jdmarshall, line 8: I'd like to use debug for this in the future but for now we only use it for the benchmarks, so I'm not going to force the issue here, despite the silent return being kind of smelly.

jdmarshall (approved):

jdmarshall:
@chidam333 Something is glitching with github actions and I'm unsure why I can't make this build. You might need to push a do-nothing change.

chidam333:
hello @jdmarshall i added a dummy commit, is the issue resolved ?

chidam333:
DCO check happened now? was that the issue ?

jdmarshall:
Not entirely sure yet. There’s a handful that are in each state. I submitted one without DCO and I’m fairly sure the builds ran anyway.

jdmarshall:
Can you update the change list?

also, do we have a benchmark for this? Would be good to identify or write one for “perf” changes.

chidam333:
image

image

image


I am seeing 30-40% improvement. (added changelog)

https://github.com/chidam333/prom-client/pull/1 <- code i used for benchmarking.

chidam333:
@jdmarshall i had to rebase for signing, but i can see that it shows a bunch of unrelated
 diffs in my pr in github, I verified that those changes are already in main, do you still want me to cherry pick those changes ?

image

[rss.patch](https://github.com/user-attachments/files/29442975/rss.patch)

added patch file for confirmation

jdmarshall:
That's... weird for a rebase.  I'm going to say no specifically because of the workflow changes. 

I'd try resetting your branch to HEAD and cherry-picking your changes back on top, please.

chidam333:
done, thank you

jdmarshall (approved):

chidam333:
thank you :) 

also faceoff looks great ❤️

Issue #743, opened 2026-03-16
`url.parse()` deprecation warning on Node v24


Wirone:
Our CLI tool reports this deprecation warning when used on Node 24.

```
(node:76468) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
    at urlParse (node:url:136:13)
    at Object.urlResolve [as resolve] (node:url:725:10)
    at Pushgateway.useGateway (<redacted>/node_modules/prom-client/lib/pushgateway.js:62:21)
    at Pushgateway.pushAdd (<redacted>/node_modules/prom-client/lib/pushgateway.js:29:21)
    ...
```

Thanks in advance for fixing this 🙂.

tomasoliveirz:
Following up on this as the deprecation warnings are still quite persistent in Node v22. I’ve identified a usage in `@open-rpc/client-js/src/__mocks__/requestData.ts` that seems to be contributing to the logs.

Any updates on a potential fix for this?

Katomoto:
One thing worth trying: setting the timeout at the transport layer rather than the request level — avoids the race condition.

jdmarshall:
I'll be merging a PR for this later today from @yvon-morice-wmx

PR #744, opened 2026-04-01
Replace usage of old node:url api with WHATWG one in pushgateway.js

yvon-morice-wmx:
Related to https://github.com/siimon/prom-client/issues/743

Replaces usage of `url.urlParse` and `url.resolve` from `node:url` by `new url.Url` from same package as these functions have been deprecated and trigger a warning on usage.

Changes based on code and documentation available at https://nodejs.org/api/url.html#urlresolvefrom-to and https://github.com/nodejs/userland-migrations/tree/main/recipes/node-url-to-whatwg-url

### DIFF
--- a/lib/pushgateway.js
+++ b/lib/pushgateway.js
@@ -46,16 +46,6 @@ class Pushgateway {
 	}
 }
 
-function legacyUrlResolve(from, to) {
-	const resolvedUrl = new URL(to, new URL(from, 'resolve://'));
-	if (resolvedUrl.protocol === 'resolve:') {
-		// `from` is a relative URL.
-		const { pathname, search, hash } = resolvedUrl;
-		return pathname + search + hash;
-	}
-	return resolvedUrl.toString();
-}
-
 function legacyUrlFromWhatwgUrl(whatwgUrl) {
 	const legacyUrl = {
 		hash: whatwgUrl.hash,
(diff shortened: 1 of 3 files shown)


lib/pushgateway.js
```diff
@@ -45,10 +45,46 @@ class Pushgateway {
...
+		// `from` is a relative URL.
+		const { pathname, search, hash } = resolvedUrl;
+		return pathname + search + hash;
+	}
+	return resolvedUrl.toString();
+}
+
+function legacyUrlFromWhatwgUrl(whatwgUrl) {
+	const legacyUrl = {
+		hash: whatwgUrl.hash,
```
jdmarshall: If you're going to go through the trouble of unwinding URL->url attribute copying it'd be better off to work with URL internally and just do the jiggery pokery for forwarding the old style requestOptions over the top, so we don't round-trip to the legacyUrlResolve function twice.

jdmarshall (changes requested):

jdmarshall (changes requested):
Still hasn't removed the old url, just uses it less.

Until the http request is being made with URL I don't believe this PR achieves the goal of replacing the legacy url parser. 

Also this PR will need pinning tests. 

So what I'd want to see is:

 - no urls, full stop
 - No manual interpolation. I'm a stickler about this an every time I haven't been has turned into a security hole. And you shouldn't need it at any rate if we stop going back and forth between. 
 - splitting the URL and the request Options before calling `http.request(url, options)` removes the old format and any translation issues lurking there, and also solves the no interpolation problem handily.


This is a Kernighan's Law situation. It's a "there are no obvious bugs" vs "there are obviously no bugs" (Hoare) scenario. The new version should be obviously correct or we haven't really fixed the hole, just punched new ones.

jdmarshall (commented):
Great work. Thanks for the contribution.

I'll take a bit longer perusal of this later today but this looks good on first scan.

Edit: Can you hit the changelog? I know it's kind of a pain. We're thinking about how else to handle it but that's the setting for now.

jdmarshall (approved):

Issue #632, opened 2024-05-21
v8 heap_size_limit


sakari:
Would it be possible to add [heap_size_limit](https://nodejs.org/api/v8.html#v8getheapspacestatistics) gauge to the default nodejs metrics that get collected?

It would be useful for displaying an upper limit for heap usage of node processes

riosje:
take a look to [N|solid](https://github.com/nodesource/nsolid) it is  a nodejs alternative 100% compatible which provides telemetry out of the box, no instrumentation needed.
Even [the metrics](https://docs.nodesource.com/nsolid/5.0/docs#nsolidmetrics) you're looking for and much more.

I wrote a blog post some time ago about how to integrate it with prometheus+grafana, but currently you can connect anything that supports OTLP

PR #745, opened 2026-04-04
feat: expose V8 heap size limit as default metric (nodejs_heap_size_limit_bytes)

szymonskirgajllo:
Motivation
Node’s heap grows over time up to a configured maximum. Metrics such as heapUsed or heapTotal from process.memoryUsage() reflect current allocation, not how close the process is to its hard ceiling. Alerts based only on “heap is large” or on heapUsed / heapTotal often track elastic growth rather than risk of hitting the V8 limit and OOM.

When operators set an explicit heap cap (e.g. V8’s --max-heap-size, or related flags), they need a stable denominator in Prometheus: the maximum JavaScript heap size the runtime will allow. Comparing used heap to that limit supports alerts such as “≥90% of the configured heap is in use,” which align with capacity planning and OOM risk, not with normal heap expansion.

What this PR does
* Adds default gauge nodejs_heap_size_limit_bytes, sourced from v8.getHeapStatistics().heap_size_limit, alongside the existing heap metrics in heapSizeAndUsed (same labels as nodejs_heap_size_* / external memory).
* Omits this metric when getHeapStatistics is unavailable or throws ERR_NOT_IMPLEMENTED (e.g. some non-Node runtimes); metricNames stays consistent with what is actually registered.\
* Treats a missing getHeapStatistics implementation like unsupported (safe with partial v8 mocks in tests).
* Extends tests (including isolated v8 behaviour) and updates the default-metrics example / changelog.

Example alerting (Prometheus)
`nodejs_heap_size_used_bytes / nodejs_heap_size_limit_bytes > 0.9`

Tune thresholds and for: to your SLOs. Ensure the process is started with a known heap cap so the limit metric matches your intent.

Notes
* heap_size_limit is read from V8 on each collection (same pattern as before for clarity).
* Limit is process-level; it is not per heap space (those remain in the existing space metrics).

### DIFF
--- a/lib/metrics/heapSizeAndUsed.js
+++ b/lib/metrics/heapSizeAndUsed.js
@@ -1,21 +1,42 @@
 'use strict';
 
 const Gauge = require('../gauge');
+const v8 = require('v8');
 const safeMemoryUsage = require('./helpers/safeMemoryUsage');
 
 const NODEJS_HEAP_SIZE_TOTAL = 'nodejs_heap_size_total_bytes';
 const NODEJS_HEAP_SIZE_USED = 'nodejs_heap_size_used_bytes';
+const NODEJS_HEAP_SIZE_LIMIT = 'nodejs_heap_size_limit_bytes';
 const NODEJS_EXTERNAL_MEMORY = 'nodejs_external_memory_bytes';
+
+function isHeapStatisticsSupported() {
+	if (typeof v8.getHeapStatistics !== 'function') {
+		return false;
+	}
+	try {
+		v8.getHeapStatistics();
+		return true;
+	} catch (e) {
+		if (e.code === 'ERR_NOT_IMPLEMENTED') {
+			return false;
+		}
+		throw e;
+	}
+}
+
+const heapStatisticsSupported = isHeapStatisticsSupported();
 
 module.exports = (registry, config = {}) => {
 	if (typeof process.memoryUsage !== 'function') {
 		return;
 	}
+
 	const labels = config.labels ? config.labels : {};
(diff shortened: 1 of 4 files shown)


lib/metrics/heapSizeAndUsed.js
```diff
@@ -1,21 +1,42 @@
 'use strict';
 
 const Gauge = require('../gauge');
+const v8 = require('v8');
```
jdmarshall, line 4: This isn't going to work with Bun. Probably better to move this into `isHeapStatisticsSupported()`

lib/metrics/heapSizeAndUsed.js
```diff
@@ -25,6 +46,13 @@ module.exports = (registry, config = {}) => {
 				externalMemUsed.set(labels, memUsage.external);
 			}
 		}
+		if (heapSizeLimit) {
```
jdmarshall, line 49: Not running the check here?

jdmarshall (changes requested):

jdmarshall:
This should fix #632

lib/metrics/heapSizeAndUsed.js
jdmarshall: For Bun, I believe this is what would be needed:

https://bun.com/reference/bun/jsc/heapStats

jdmarshall (changes requested):

jdmarshall:
@szymonskirgajllo If you're available to look at this, we are actively monitoring the PRs now and we can get this published sooner rather than later.

Issue #458, opened 2021-08-19
Add example of using Summary


toymachiner62:
The /examples folder has examples for the other metrics but an example of using Summary is missing

toymachiner62:
The /examples folder has examples for the other metrics but an example of using Summary is missing

vishwa357:
Hi @zbjornson , I've raised a PR.
I'm requesting you to review as you've worked on examples before.
How does the change look?

jdmarshall:
Closed fixed.

PR #746, opened 2026-05-06
Adding example of Summary metric 

vishwa357:
## Description

Adding example for Summary metric.

Issue: https://github.com/siimon/prom-client/issues/458


jdmarshall (approved):
Not a huge fan of console.log but this will do.

PR #751, opened 2026-06-25
chore: rename module to @prometheus/client_js

cjihrig:
This commit updates the project name from prom-client to @prometheus/client_js to reflect the recent move.

This is my first time contributing to this project under any package name, so please let me know if anything here is incorrect.

### DIFF
--- a/benchmarks/index.js
+++ b/benchmarks/index.js
@@ -13,9 +13,9 @@ const debug = require('debug')('benchmark');
  */
 
 const benchmarks = new Benchmark({
-	'@prometheus-io/client@latest': '@prometheus-io/client@latest',
-	'@prometheus-io/client@trunk': 'git@github.com:prometheus/client_js',
-	'@prometheus-io/client@current': { location: process.cwd() },
+	'@prometheus/client@latest': '@prometheus/client@latest',
+	'@prometheus/client@trunk': 'git@github.com:prometheus/client_js',
+	'@prometheus/client@current': { location: process.cwd() },
 });
 
 benchmarks.suite('counter', require('./counter'));
--- a/benchmarks/index.js
+++ b/benchmarks/index.js
@@ -13,9 +13,9 @@ const debug = require('debug')('benchmark');
  */
 
 const benchmarks = new Benchmark({
-	'prom-client@latest': 'prom-client@latest',
-	'prom-client@trunk': 'git@github.com:siimon/prom-client',
-	'prom-client@current': { location: process.cwd() },
+	'@prometheus/client_js@latest': '@prometheus/client_js@latest',
+	'@prometheus/client_js@trunk': 'git@github.com:prometheus/client_js',
+	'@prometheus/client_js@current': { location: process.cwd() },
 });
 
 benchmarks.suite('counter', require('./counter'));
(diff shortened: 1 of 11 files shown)


slagiewka:
I know that @krajorama mentioned `client_js` as the desired state in https://github.com/prometheus/client_js/issues/750#issuecomment-4791013670. But this might've been intended to point to the repository name, not package name.

Reality check:
Python: `pip install prometheus-client`
Ruby: `gem install prometheus-client`
Rust: `cargo add prometheus-client`

Go is of course an outlier since it imports directly from domain's namespace. And in Go packages with `_` in name are more commonplace. 

`prometheus-client` is taken on npm. `@prometheus/client` would be more aligned with the npm ecosystem.

krajorama:
I don't have a strong opinion, I feel that @prometheus-io/client_js  would be nice to reflect the name of the repo it comes from, but I'm ok with @prometheus-io/client as well.  (Note : currently Prometheus publishes packages under [prometheus-io](https://www.npmjs.com/search?q=%40prometheus-io%2F)).

@jdmarshall any preference?

UPDATE: as far as I see I can freely add @prometheus organization, so we can do @prometheus prefix as well.

cjihrig:
Personally, I think `@prometheus/client` would be the nicest and clearest, since that is what people will be typing when they install. However, if `@prometheus-io` is already being used, then I think `@prometheus-io/client` would be the best option to avoid confusion and managing multiple package scopes. The `_js` probably isn't needed since the package is coming from npm.

benchmarks/index.js
```diff
@@ -13,9 +13,9 @@ const debug = require('debug')('benchmark');
  */
 
 const benchmarks = new Benchmark({
-	'prom-client@latest': 'prom-client@latest',
-	'prom-client@trunk': 'git@github.com:siimon/prom-client',
-	'prom-client@current': { location: process.cwd() },
+	'@prometheus-io/client@latest': '@prometheus-io/client@latest',
```
jdmarshall: These are going to conflict with another PR that simplifies the names for the benchmarks. I'm landing that momentarily.

cjihrig: Thanks for the heads up. I can rebase after the other PR lands.

jdmarshall:
With no other knowledge I'd prefer @prometheus but I'd like to know what's already published under prometheus-io

Hmm.  The only thing that suggests prometheus-io as a better option is https://www.prometheus.io which I suppose is perhaps enough.

cjihrig:
Is there anything I can do to help move this forward?

benchmarks/index.js
jdmarshall: Also the line mentioned above is currently failing the benchmarks. You'll need to revert this line for now.

jdmarshall:
@cjihrig Yeah let's go with @prometheus instead of @promethius-io since Github seems like a more common destination from npmjs.com.

And you need to revert the @latest change since there's a chicken and egg problem with updating the docs and _then_ publishing the code. 

I was hoping @SimenB would approve the other change and you could build yours on top but at this point it's more expedient for me to rebase on top of yours instead, since mine is about 12 lines of code.

cjihrig:
I've made the requested changes.

jdmarshall (approved):

PR #756, opened 2026-07-01
drop node 20 support

cjihrig:
Node 20.x went End-of-Life on 2026-04-30.

The package rename seems like a good time for such a change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


jdmarshall:
I'm usually loathe to drop backward compatibility if there's no gain to be had for doing so, but there are at least a couple we could do.

Small performance gain to be had on aggregation for 

https://github.com/prometheus/client_js/commit/b3916cdff6097745c1182cacf623876e702da964

Which I had to modify for reasons of node 20 compatibility. Wonder if there are any more.

jdmarshall (approved):

PR #759, opened 2026-07-02
Add node 26 to the test matrix.

jdmarshall:


cjihrig (approved):
I think the CI wants a changelog entry, but happy to see this.

jdmarshall:
> I think the CI wants a changelog entry, but happy to see this.

I don't think we need a changelog for every single chore, but this one might rate a mention...

cjihrig (approved):
LGTM. The only thing I'm unsure of is whether or not you want to update the engines field in package.json. Right now it has `>=24`, which covers this. I'm just not sure what the project's stance is. That range would also cover Node 25, for example, which may or may not be supported by the project, but is technically EOL.

cjihrig (approved):

jdmarshall:
@zbjornson welcome back and can you take a look at this one?

zbjornson (approved):
I'm good either way on `engines`.

PR #769, opened 2026-07-08
Support registry as Pushgateway second argument

### LINKED ISSUE #652: Problem in .d.ts definition,  -> logical errors when using typescript code
First, thx for your great work.

The last 3 hours i was looking for a problem. My pushgateway instance seems to have no connection to the registry.

Since the TS Compiler does not show an error, i've not checked to code ...

```ts
let push = new Pushgateway('http://127.0.0.1:9091', reg);
```

This code is TS correct, but wrong.

Reason: 
The Typedefintion

https://github.com/siimon/prom-client/blob/c1d76c5d497ef803f6bd90c56c713c3fa811c3e0/index.d.ts#L700

``options`` is typed  as any and ``registry`` as optional.

So it is correct to forget the ``option`` parameter, so the registry parameter ist used as "option"

Which does not make any sence. I created a Pushgateways without a connected registry ...

Solution:

If i look into the source code the type is ``Url`` from packege "url" extended with a single property ``requireJobName: true,`` ?

So a better type for option is `{requireJobName: boolean}`.

bugprone:
## Purpose

`Pushgateway` accepts an optional `options` argument before the `registry` argument. In TypeScript, `new Pushgateway(url, registry)` was accepted because `options` was typed as `any`, but at runtime that registry object was treated as request options and the Pushgateway fell back to the global registry.

This makes the two-argument registry form work at runtime and declares it explicitly in the TypeScript definitions.

Fixes #652.

## Changes

- Treat a `Registry` passed as the second `Pushgateway` constructor argument as the registry
- Add constructor overloads for `new Pushgateway(url, registry)` and `new Pushgateway(url, options, registry)`
- Add a small `Pushgateway.Options` type with `requireJobName` plus request option passthrough
- Add runtime coverage for pushing metrics from a second-argument registry
- Add TypeScript compile coverage for the supported constructor forms
- Update the unreleased changelog

## Testing

- `npm test -- --runInBand --forceExit`
- Manual smoke check with a local HTTP server verifying `new Pushgateway(url, registry).push()` sends metrics from the provided registry

### DIFF
--- a/lib/pushgateway.js
+++ b/lib/pushgateway.js
@@ -7,13 +7,10 @@ const Registry = require('./registry');
 const { globalRegistry } = Registry;
 
 class Pushgateway {
-	constructor(gatewayUrl, options, registry) {
+	constructor(gatewayUrl, options, registry = globalRegistry) {
 		if (options instanceof Registry) {
 			registry = options;
-			options = undefined;
-		}
-		if (!registry) {
-			registry = globalRegistry;
+			options = {};
 		}
 		this.registry = registry;
 		this.gatewayUrl = gatewayUrl;
--- a/lib/pushgateway.js
+++ b/lib/pushgateway.js
@@ -3,10 +3,15 @@
 const http = require('http');
 const https = require('https');
 const { gzipSync } = require('zlib');
-const { globalRegistry } = require('./registry');
+const Registry = require('./registry');
+const { globalRegistry } = Registry;
 
 class Pushgateway {
 	constructor(gatewayUrl, options, registry) {
+		if (options instanceof Registry) {
+			registry = options;
+			options = undefined;
+		}
 		if (!registry) {
 			registry = globalRegistry;
 		}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This release marks our first release under the Prometheus umbrella.
 
 ### Changed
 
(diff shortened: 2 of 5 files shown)


lib/pushgateway.js
```diff
@@ -3,10 +3,15 @@
 const http = require('http');
 const https = require('https');
 const { gzipSync } = require('zlib');
-const { globalRegistry } = require('./registry');
+const Registry = require('./registry');
+const { globalRegistry } = Registry;
 
 class Pushgateway {
 	constructor(gatewayUrl, options, registry) {
+		if (options instanceof Registry) {
```
jdmarshall, line 11: Strictly speaking, This block and the following block are entangled, since gatewayUrl, registry, registry would not be a legal combination.

Otherwise, I'm surprised this is missing. Someone got enthusiastic in the TS file.

Probably this should be:

	constructor(gatewayUrl, options, registry = globalRegistry) {
		if (options instanceof Registry) {
			registry = options;
			options = {};
		}

jdmarshall (changes requested):

jdmarshall:
Related to #652

lib/pushgateway.js
bugprone: Good point about the entangled blocks. Collapsed them into the default parameter form you suggested.

jdmarshall (commented):
LGTM

jdmarshall (approved):

PR #771, opened 2026-07-12
fix(cluster): aggregate metrics in a deterministic order

### LINKED ISSUE #539: Cluster should always aggregate metrics in a same order
The issue with the current metrics aggregation is that it occurs in the order in which individual workers respond to the GET_METRICS_REQ message. This may cause variations in the aggregation results when the metrics contain floats, such as a histogram. 

For example, if workers respond to the first request with values 0.5848208, 0.5479198, 0.3437699 (which sum to 1.4765105), and then respond to a second request with the same values but in a different order (0.3437699+0.5848208+0.5479198), the result of the aggregation will be1.4765104999999998 in JS, due to float errors. 

This can trigger a "reset" detection in Prometheus and severely impact the accuracy of the graphs.

AysajanE:
## Problem

Cluster metric aggregation (`lib/metricAggregators.js`) sums worker-reported values in the order workers respond to the `GET_METRICS_REQ` message. Because floating-point addition is not associative, the same set of values summed in different orders can produce slightly different results (e.g. `1.4765105` vs `1.4765104999999998`). As reported in #539, this non-determinism can trigger a false "reset" detection in Prometheus and corrupt graphs.

## Fix

- Sort values before summing (`sumValues`) so the `sum` and `average` aggregators produce a reproducible result regardless of the order in which workers respond.
- Add a unit test that aggregates the same values in two different orders and asserts the results are strictly equal.

### Checks (Node 22)
- `npm run test-unit` — 543 passing, including the new order-independence test.

Fixes #539

---
Disclosure: this change was prepared with AI assistance and reviewed by me before submitting. I ran the check above in a network-isolated container and published a signed, re-runnable record of that run, verifiable via GitHub artifact attestation: https://northset-oss.github.io/verification-pilot/. Contributor self-run, not a maintainer verification.

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -39,16 +39,6 @@ let requestCtr = 0; // Concurrency control
 let listenersAdded = false;
 const requests = new Map(); // Pending requests for workers' local metrics.
 
-function deferred() {
-	let resolve;
-	let reject;
-	const promise = new Promise((resolvePromise, rejectPromise) => {
-		resolve = resolvePromise;
-		reject = rejectPromise;
-	});
-	return { promise, resolve, reject };
-}
-
 class AggregatorRegistry extends Registry {
 	constructor(regContentType = Registry.PROMETHEUS_CONTENT_TYPE) {
 		super(regContentType);
@@ -83,14 +73,9 @@ class AggregatorRegistry extends Registry {
 				}
 			}
 
-			const responses = new Map();
-			const responsePromises = workers.map(worker => {
-				const response = deferred();
-				responses.set(worker.id, response);
-				return response.promise;
-			});
+			const responseHandlers = new Map();
 			const request = {
-				responses,
+				responseHandlers,
 				done,
 				errorTimeout: setTimeout(() => {
 					const err = new Error('Operation timed out.');
@@ -110,7 +95,16 @@ class AggregatorRegistry extends Registry {
 				return;
 			}
 
-			for (const worker of workers) worker.send(message);
+			const responsePromises = workers.map(
+				worker =>
+					new Promise((resolveResponse, rejectResponse) => {
+						responseHandlers.set(worker.id, {
+							resolve: resolveResponse,
+							reject: rejectResponse,
+						});
+						worker.send(message);
+					}),
+			);
 
 			Promise.all(responsePromises)
 				.then(metrics => Registry.aggregate(metrics.flat()).metrics())
@@ -177,11 +171,11 @@ function addListeners() {
 					return;
 				}
 
-				const response = request.responses.get(worker.id);
+				const response = request.responseHandlers.get(worker.id);
 				if (response === undefined) {
 					return;
 				}
-				request.responses.delete(worker.id);
+				request.responseHandlers.delete(worker.id);
 
 				if (message.error) {
 					response.reject(new Error(message.error));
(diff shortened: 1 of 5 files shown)


lib/metricAggregators.js
```diff
@@ -51,14 +51,21 @@ function AggregatorFactory(aggregatorFn) {
 // Export for users to define their own aggregation methods.
 exports.AggregatorFactory = AggregatorFactory;
 
+function sumValues(values) {
+	return values
+		.slice()
+		.sort((a, b) => a.value - b.value)
```
jdmarshall: This is not the place or the field that needs to be sorted. The results need to be sorted by threadId at the caller. 

And that could be achieved by creating a promise at the point of the message send, and doing a Promise.all() on the results. 

Something slightly different might need to be done in worker.js

jdmarshall (changes requested):

lib/metricAggregators.js
AysajanE: Thanks — understood. I put determinism in the wrong layer. I will remove the aggregator-level sort and instead collect responses in stable worker-ID order in `cluster.js` and stable thread-ID order in `worker.js`, with regression tests that reverse response arrival. I understand this addresses response-order nondeterminism for a stable worker set, not changes caused by workers entering or exiting. Is that the scope you want before I update the branch?

lib/cluster.js
```diff
@@ -85,20 +104,17 @@ class AggregatorRegistry extends Registry {
...
-
-			if (request.pending === 0) {
+			if (workers.length === 0) {
 				// No workers were up
-				clearTimeout(request.errorTimeout);
 				process.nextTick(() => done(undefined, ''));
+				return;
 			}
+
+			for (const worker of workers) worker.send(message);
```
jdmarshall: If this was rolled into the function starting on line 87 you wouldn't need the 'deferred()' function, which is a confusing name for what it does, since it doesn't do anything that's deferred.

lib/cluster.js
```diff
@@ -53,13 +63,17 @@ class AggregatorRegistry extends Registry {
...
+			.filter(worker => worker.isConnected())
+			.sort((left, right) => left.id - right.id);
 
 		return new Promise((resolve, reject) => {
 			let settled = false;
 			function done(err, result) {
 				if (settled) return;
 				settled = true;
 
+				clearTimeout(request.errorTimeout);
```
jdmarshall, line 66: 👍

lib/cluster.js
```diff
@@ -161,21 +177,16 @@ function addListeners() {
 					return;
 				}
 
-				if (message.error) {
-					request.done(new Error(message.error));
+				const response = request.responses.get(worker.id);
+				if (response === undefined) {
 					return;
 				}
+				request.responses.delete(worker.id);
```
jdmarshall: I think your AI is slipping. Two functions, same purpose, different APIs.

In one you're using has() to short circuit on results that have already been reported (which probably shouldn't happen anyway and could perhaps use a debug statement here) and in the other you're deleting keys from the Map. 

They should be built on the same strategies.

jdmarshall (changes requested):

lib/cluster.js
AysajanE: Addressed in b64dc65: removed the deferred helper and now create each response promise inline where the worker message is sent.

lib/cluster.js
AysajanE: Addressed in b64dc65: cluster.js and worker.js now use the same per-peer response-handler lifecycle. Each removes the pending handler when a response arrives, resolves or rejects through Promise.all(), and aggregates in stable ID order. The full npm test gate passes with 543 tests.

jdmarshall (approved):

PR #773, opened 2026-07-13
feat(counter): optionally use the counter value as the exemplar value

### LINKED ISSUE #621: Exemplar values in counter metrics are always 1
## Problem

### Example

image

---

## Proposal

### Example

Given code such as:

```ts
```
  
```
```

```
```

---

AysajanE:
## Problem

When a counter is incremented with `inc({ exemplarLabels })`, the exemplar's value is set to the amount passed to `inc` — `1` by default. As raised in #621, that makes exemplars unhelpful for the common case of calling `inc()` with no argument: every exemplar reads `1`, so in tools like Grafana the exemplar markers all sit at the bottom of the graph regardless of the counter's actual value.

## Change

- Add an opt-in Counter option `useCounterValueAsExemplar` (default `false`, so existing behavior is unchanged). When `true`, the exemplar value recorded on an increment is the counter's current total rather than the increment amount.
- `lib/counter.js` threads the option through and records `entry.value` vs. the passed value accordingly; `index.d.ts` types the new option.
- Add a unit test covering both modes (exemplar value `2` with the option after two `inc()` calls, `1` without it).

### Checks (Node 22)
- `npm run test-unit` — 543 passing, including the new exemplar-value test.

Fixes #621

---
Disclosure: this change was prepared with AI assistance and reviewed by me before submitting.
I ran the check(s) above in a network-isolated container and published a signed, re-runnable
record of that run, verifiable via GitHub artifact attestation:
https://northset-oss.github.io/verification-pilot/. Contributor self-run, not a maintainer verification.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -323,6 +323,7 @@ export interface CounterConfiguration<
 	T extends string,
 > extends MetricConfiguration<T> {
 	collect?: CollectFunction<Counter<T>>;
+	useCounterValueAsExemplar?: boolean;
 }
 
 export interface IncreaseDataWithExemplar<T extends string> {
--- a/lib/counter.js
+++ b/lib/counter.js
@@ -24,7 +24,7 @@ const Exemplar = require('./exemplar');
 
 class Counter extends Metric {
 	constructor(config) {
-		super(config);
+		super(config, { useCounterValueAsExemplar: false });
 		this.type = 'counter';
 		this.defaultLabels = {};
 		this.defaultValue = 1;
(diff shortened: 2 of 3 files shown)


jdmarshall (changes requested):
I've added some analysis to the linked issue. If you're willing to rework this PR to follow that advice I'll let you have the credit on this one. You'll save me having to write a decent set of tests for the fix.  

This does not require a flag I don't think. The next release will be effectively semver major, and this feature is so broken I don't see how anyone could be relying on it as written.  You'll need a changelog entry for this one for sure. Under Breaking.

jdmarshall:
Two different users, filing PRs for the same bug, with the exact same (incorrect) unit test in both versions. My bandwidth for dealing with AI generated PRs has reached its end for today.

PR #774, opened 2026-07-13
fix(types): allow string[] label values

### LINKED ISSUE #582: types: handler might be string or array of strings
Hello

handler as string[] is correctly displayed in metrics:

> http_request_duration_milliseconds_count{handler="/tiles/:segments(*),/tiles",code="200",method="HEAD"} 3

but TS can't agree to use it:

Screenshot 2023-08-10 at 10 45 17

Screenshot 2023-08-10 at 10 45 47

AysajanE:
## Problem

prom-client accepts an array of strings as a label value at runtime — e.g. `.inc({ handler: ['/a', '/b'] })`, which renders as `handler="/a,/b"` — but its TypeScript types reject it: the label-values type only allows `string | number`, so that call fails to type-check even though it works. Reported in #582.

## Change

- `index.d.ts`: widen the label-values type to `string | number | string[]` so array-valued labels type-check.
- `test/typescript.ts`: add an array-valued-label usage so `tsc` proves the type now accepts it.

### Checks (Node 22)
- `npm run compile-typescript` (`tsc --project .`) and `npm run test-unit` — both pass.

Fixes #582

---
Disclosure: this change was prepared with AI assistance and reviewed by me before submitting.
I ran the check(s) above in a network-isolated container and published a signed, re-runnable
record of that run, verifiable via GitHub artifact attestation:
https://northset-oss.github.io/verification-pilot/. Contributor self-run, not a maintainer verification.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -304,7 +304,7 @@ type MetricValueWithName<T extends string> = MetricValue<T> & {
 
 type LabelValues<T extends string> = T extends NoLabelNameType
 	? Partial<Record<string, never>>
-	: Partial<Record<T, string | number>>;
+	: Partial<Record<T, string | number | string[]>>;
 
 interface MetricConfiguration<T extends string> {
 	name: string;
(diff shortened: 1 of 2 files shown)


jdmarshall (commented):
This is a subset of #553 that doesn't really move things forward and there was resistance to landing that one that hasn't really abated. 

Just making it work for counter isn't really answering the ticket.

PR #775, opened 2026-07-13
fix(counter): use counter total as exemplar value instead of increment

### LINKED ISSUE #621: Exemplar values in counter metrics are always 1

c436zhan:
## Summary

When a `Counter` with exemplars enabled is incremented via `inc({ exemplarLabels })`, the exemplar's value was set to the increment amount (default `1`) instead of the counter's current total. This caused every exemplar to read `1` in common usage, making exemplar markers in tools like Grafana sit at the bottom of the graph regardless of the counter's actual value.

## Root cause

In `Counter.incWithExemplar`, after calling `incWithoutExemplar(labels, value)`, the code passed the raw increment `value` to `updateExemplar`. The exemplar should instead record the counter's total after the increment, which is what other Prometheus clients do.

## Changes

- `lib/counter.js`: pass `this.store.entry(labels).value` (the counter total) to `updateExemplar` instead of `value` (the increment amount).
- `test/exemplarsTest.js`: add regression test verifying that after `inc()` then `inc(3)`, the second exemplar records `4` (the total), not `3`.
- `CHANGELOG.md`: document the breaking behavior change under the Unreleased section.

## Testing

- `npm test` — 543 passing (28 test suites)
- `npx jest test/exemplarsTest.js` — 9 passing, 1 snapshot passing
- Verified with a standalone reproduction script that exemplar values now match the counter total on each increment.

## Compatibility

This changes the value written to counter exemplars from the increment amount to the counter total. Anyone relying on the previous (buggy) behavior of exemplars always being `1` will see a change. The fix aligns with the behavior described in the linked issue and with other Prometheus client libraries.

Fixes #621

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This release marks our first release under the Prometheus umbrella.
 - Drop support for Node.js versions 16, 18, 20, 21 and 23
 - Metric internal storage ('hashMap') changed to a separate object, LabelMap. If you have
   subclassed the built-in metric types you may need to adjust your code.
+- Counter exemplar values now record the counter's total value instead of the increment
+  amount, matching the behavior of other Prometheus clients (Fixes [#621](https://github.com/prometheus/client_js/issues/621))
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


jdmarshall:
Two different users, filing PRs for the same bug, with the exact same unit test in both versions. My bandwidth for dealing with AI generated PRs has reached its end for today.

Issue #708, opened 2025-09-08
Debug npm package was compromised this morning


jdmarshall:
For anyone working on a fork of this code, if you did an npm update in the last 24 hours you should pay attention to whether you got debug version 4.4.2. 

https://github.com/debug-js/debug/issues/1005#issuecomment-3266885191

Since I was responsible for the PR that introduced it as a dev dependency I felt I should ping the prom-client community. 

Of course if you’re just using prom-client, you won’t have the dev dependencies, but debug is a popular library.

PR #778, opened 2026-07-13
Add debug logging for metrics collection failures

### LINKED ISSUE #649: Add integration with rpm package debug
DEBUG=prom:* node app.js

rishh09:
Fixes #649

#### What this PR does
Adds debug logging for metrics collection failures that were previously silently ignored.

The debug namespaces are scoped by metric area, for example:

- `prom:metrics:gc`
- `prom:metrics:event-loop-lag`
- `prom:metrics:process-open-fds`
- `prom:metrics:process-max-fds`

This keeps the default behavior quiet, while allowing users to inspect these failure paths with `DEBUG=prom:*`.

#### Notes
`debug` is moved from `devDependencies` to `dependencies` because it is now required by runtime library files.

#### Tests
- `npm run lint`
- `npm run check-prettier`
- `npm run compile-typescript`
- `npm run test-unit -- --runInBand --forceExit`
- `npm test -- --runInBand --forceExit`

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,6 @@
 	},
 	"dependencies": {
 		"@opentelemetry/api": "^1.4.0",
-		"debug": "^4.4.1",
 		"tdigest": "^0.1.1"
 	},
 	"types": "./index.d.ts",
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
 	"homepage": "https://github.com/prometheus/client_js",
 	"devDependencies": {
 		"@eslint/js": "^9.29.0",
-		"debug": "^4.4.1",
 		"eslint": "^9.29.0",
 		"eslint-config-prettier": "^10.1.5",
 		"eslint-plugin-jsdoc": "^61.4.1",
@@ -54,6 +53,7 @@
 	},
 	"dependencies": {
 		"@opentelemetry/api": "^1.4.0",
+		"debug": "^4.4.1",
 		"tdigest": "^0.1.1"
 	},
 	"types": "./index.d.ts",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ This release marks our first release under the Prometheus umbrella.
 
 ### Added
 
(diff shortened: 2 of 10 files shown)


rishh09:
Hi @jdmarshall, I opened this PR for #649 based on your latest comment about adding debug output to empty catch blocks.

I wasn’t able to add you as a reviewer directly, so tagging you here for visibility. Happy to adjust the namespace or message style if you prefer a different convention.

package.json
```diff
@@ -54,6 +53,7 @@
 	},
 	"dependencies": {
 		"@opentelemetry/api": "^1.4.0",
+		"debug": "^4.4.1",
```
jdmarshall: Since this library is only used within other tools, should this be a dep or a peer dependency? @SimenB  @zbjornson

zbjornson: It'd be a dependency, but I'd actually use Node.js's built-in debug functionality and avoid a dependency entirely: https://nodejs.org/dist/latest-v22.x/docs/api/util.html#utildebuglogsection-callback. (Note there's a `debug` alias documented just below that section.)

Are there specific features of the debug package you prefer over the built-in one, @rishh09?

jdmarshall: Oh! Right. I completely forgot about that. Despite moving a project to the builtin not five months ago. It's been a year.

jdmarshall (changes requested):
As @zbjornson recommended, please migrate to the Node native debug logging found in node:util

jdmarshall:
@rishh09 If you would be so kind, would you mind fixing https://github.com/prometheus/client_js/blob/4339b07033c26a565edbcaab492249685c3d2ef9/benchmarks/index.js#L18 while you're in here so we can remove debug entirely?

I'm thinking back to #708 and the fact that the benchmarks run as a CI action and maybe we should just not...

rishh09:
sure, I'll proceed with these recommended changes and push a commit.

jdmarshall:
This one deserves a Changelog entry under additions: support for debug messages

jdmarshall (approved):

PR #779, opened 2026-07-13
fix: skip cluster responses after IPC disconnect

### LINKED ISSUE #563: [BUG] ERR_IPC_CHANNEL_CLOSED showing up on new deployments
### The Problem
Our app uses the prom-client for updating several counters, via worker threads as well as the main thread. Whenever we do a deployment of our app, and if a worker thread with one of the counters active is interrupted it will give the following error

Screenshot 2023-04-04 at 14 46 24

### The Likely Cause

It's [these](https://github.com/siimon/prom-client/blob/master/lib/cluster.js#L201-L212) lines that are likely causing this error (as can be seen in the stack trace). The worker issuing the process.send has disconnected, and so when the process.send call is made it triggers this error.

It's the same as an issue seen nearly 3 years ago - https://github.com/siimon/prom-client/pull/244.

A likely fix for this would be to wrap those process.send calls using:

```
if ( process.connected ) {
// do the process.send here
}
```

Hope that provides enough information for this issue.

Seekers2001:
## Summary

- route worker metric responses through a small `sendToPrimary` guard
- skip both successful and error responses after the worker IPC channel disconnects
- add a regression test that emulates a disconnected worker and verifies `process.send` is not called
- document the fix in the unreleased changelog

Fixes #563.

### Regression test

- before the fix: failed because `process.send` was called once after `process.connected` became false
- after the fix: passed

### Full suite

- `npm test`
  - ESLint passed
  - Prettier passed
  - TypeScript compilation passed
  - 29 test suites passed
  - 541 tests and 34 snapshots passed

The full suite emitted its existing Jest worker force-exit warning after completion, but exited successfully with every test passing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This release marks our first release under the Prometheus umbrella.
 
 ### Changed
 
+- Avoid sending cluster worker responses after the IPC channel disconnects (Fixes [#563](https://github.com/prometheus/client_js/issues/563))
 - Allow `Pushgateway` to accept a custom registry as the second constructor argument
 - Add `Registry#getMetricsAsString()` to the TypeScript definitions
 - Improve types for no labels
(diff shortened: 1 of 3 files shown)


jdmarshall:
In this scenario the primary thread exits? How does the app keep working in that situation?

PR #785, opened 2026-07-20
Add event loop utilization to default metrics

bugprone:
Replacement for #518 as discussed there, keeping Ivan Tymoshenko credited as commit co-author.

Adds `nodejs_eventloop_utilization_summary` and `nodejs_eventloop_utilization_histogram` to the default metrics, sampled on an unref'd interval so a blocked event loop is still accounted for (the interval, buckets, percentiles, and summary window are configurable via `collectDefaultMetrics` options).

Beyond the rebase of #518 onto current main:

- aligned the config key name `eventLoopUtilizationPercentiles` across code, types, and README (the code previously read `eventLoopUtilizationSummaryPercentiles` while docs and types said otherwise)
- dropped the `perf_hooks` availability guard and the Node 10 test skip, both dead code now that Node.js >= 22 is required
- fixed the stale `node/no-unsupported-features` eslint disable that no longer resolves under the current lint setup

Full suite passes locally: 544 tests across 29 suites, lint and TypeScript compile clean.

Closes #518

### DIFF
--- a/lib/metrics/eventLoopUtilization.js
+++ b/lib/metrics/eventLoopUtilization.js
@@ -1,0 +1,91 @@
+'use strict';
+
+const Summary = require('../summary');
+const Histogram = require('../histogram');
+
+const { performance } = require('perf_hooks');
+
+// Reported always.
+const NODEJS_EVENTLOOP_UTILIZATION_SUMMARY =
+	'nodejs_eventloop_utilization_summary';
+
+const NODEJS_EVENTLOOP_UTILIZATION_HISTOGRAM =
+	'nodejs_eventloop_utilization_histogram';
+
+const DEFAULT_ELU_HISTOGRAM_BUCKETS = [
+	0.01, 0.05, 0.1, 0.25, 0.5, 0.6, 0.7, 0.75, 0.8, 0.9, 0.95, 0.99, 1,
+];
+
+const DEFAULT_ELU_SUMMARY_PERCENTILES = [
+	0.01, 0.05, 0.5, 0.9, 0.95, 0.99, 0.999,
+];
+
+module.exports = (registry, config = {}) => {
+	const eventLoopUtilization = performance.eventLoopUtilization;
+
+	const namePrefix = config.prefix ? config.prefix : '';
+	const labels = config.labels ? config.labels : {};
+	const labelNames = Object.keys(labels);
+	const registers = registry ? [registry] : undefined;
+
+	const ageBuckets = config.eventLoopUtilizationAgeBuckets
+		? config.eventLoopUtilizationAgeBuckets
+		: 5;
+
+	const maxAgeSeconds = config.eventLoopUtilizationMaxAgeSeconds
+		? config.eventLoopUtilizationMaxAgeSeconds
+		: 60;
+
+	const percentiles = config.eventLoopUtilizationPercentiles
+		? config.eventLoopUtilizationPercentiles
+		: DEFAULT_ELU_SUMMARY_PERCENTILES;
+
+	const summary = new Summary({
+		name: namePrefix + NODEJS_EVENTLOOP_UTILIZATION_SUMMARY,
+		help: 'Ratio of time the event loop is not idling in the event provider to the total time the event loop is running.',
+		maxAgeSeconds,
+		ageBuckets,
+		percentiles,
+		registers,
+		labelNames,
+	});
+
+	const buckets = config.eventLoopUtilizationBuckets
+		? config.eventLoopUtilizationBuckets
+		: DEFAULT_ELU_HISTOGRAM_BUCKETS;
+
+	const histogram = new Histogram({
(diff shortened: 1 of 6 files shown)


lib/metrics/eventLoopUtilization.js
```diff
@@ -0,0 +1,91 @@
...
+];
+
+const DEFAULT_ELU_SUMMARY_PERCENTILES = [
+	0.01, 0.05, 0.5, 0.9, 0.95, 0.99, 0.999,
+];
+
+module.exports = (registry, config = {}) => {
+	const eventLoopUtilization = performance.eventLoopUtilization;
+
+	const namePrefix = config.prefix ? config.prefix : '';
```
jdmarshall: These should be nullish checks, which will avoid problems with things being initialized to falsey values. and avoid profligate use of ternaries.

```
	const namePrefix = config.prefix ?? '';
```

lib/metrics/eventLoopUtilization.js
```diff
@@ -0,0 +1,91 @@
...
+
+	const histogram = new Histogram({
+		name: namePrefix + NODEJS_EVENTLOOP_UTILIZATION_HISTOGRAM,
+		help: 'Ratio of time the event loop is not idling in the event provider to the total time the event loop is running.',
+		buckets,
+		registers,
+		labelNames,
+	});
+
+	const intervalTimeout = config.eventLoopUtilizationTimeout || 100;
```
jdmarshall: config.eventLoopUtilizationTimeout ?? 100;

test/metrics/eventLoopUtilizationTest.js
```diff
@@ -0,0 +1,106 @@
...
+
+			const le1Bucket = eluHistogramMetric.values[bucketsCount - 2];
+			expect(le1Bucket.labels.le).toEqual(1);
+			expect(le1Bucket.value).toEqual(count.value);
+		}
+	});
+});
+
+async function blockEventLoop(ratio, ms) {
+	const frameMs = 1000;
```
jdmarshall: This is a long time for a unit test. Can we make it 100ms slices and still get the sort of fidelity we are looking for? I don't really want to see an unit test go above 1 second, and ideally they are a half a millisecond apiece.

If you've ever done code flow using the watch feature you know that slow tests throw your groove off once you have more than a handful.

lib/metrics/eventLoopUtilization.js
bugprone: Done in e36e847 — switched all the config defaults in this file to `??` (the two you flagged plus ageBuckets/maxAgeSeconds/percentiles/buckets, which had the same falsey-override problem). One note: `??` now lets an explicit `0` timeout through, which would make `timeMs / intervalTimeout` infinite in the sampling loop — happy to clamp it if you'd like.

lib/metrics/eventLoopUtilization.js
bugprone: Done in e36e847.

test/metrics/eventLoopUtilizationTest.js
bugprone: Done in e36e847 — 100ms slices, 800ms total. Ran it 10 times in a row to check fidelity: all passed with the ELU delta well inside the 0.05 tolerance, and the test now finishes in ~0.8s.

jdmarshall:
Oh boy. This is already covered by lib/metrics/eventLoopLag.js but using different metric types.

And confusingly, this metric is much older than the associated ticket. Why do we want both? And if some people want both, perhaps the feature to eliminate unwanted defaults should also land with this feature.

bugprone:
As I understand it, ELU and lag complement each other. ELU shows how busy the loop is before lag starts to spike. That was also the motivation in #506. But you know this project much better than I do. If you think lag is enough, I'm fine closing this. If you want the exclude option first, I can work on that.

jdmarshall:
I have few opinions at present on the default metrics and am mostly going on feedback from others. 

Oh, you’re right. I talked myself into this one and then talked myself back out again. They both do an immediate call and measure the latency but they are reporting separate metrics otherwise.

test/metrics/eventLoopUtilizationTest.js
```diff
@@ -0,0 +1,112 @@
+'use strict';
+
+const { setTimeout: sleep } = require('timers/promises');
+const register = require('../../index').register;
+const elu = require('../../lib/metrics/eventLoopUtilization');
+const { eventLoopUtilization } = require('perf_hooks').performance;
+
+describe('eventLoopUtilization', () => {
+	beforeAll(() => {
```
jdmarshall: A beforeEach would suffice here.

jdmarshall (approved):

test/metrics/eventLoopUtilizationTest.js
bugprone: Done.

PR #786, opened 2026-07-21
fix(types): export MetricObjectWithValues, MetricValue and related types

milcho0604:
`Registry.getMetricsAsJSON()` returns `MetricObjectWithValues<MetricValue<string>>[]`, and the per-metric `get()` methods return `MetricObjectWithValues<...>`, but those types — along with the base `MetricObject` and `MetricValueWithName` — are not exported from `index.d.ts`.

As a result, consumers cannot name the return type of these public methods and have to fall back to workarounds like `Awaited<ReturnType<Registry['getMetricsAsJSON']>>`.

This exports the four types, consistent with the ~20 other public types already exported from the same declaration file. No behavior change, declaration-only.

- `npm run compile-typescript` passes.
- `prettier --check index.d.ts` passes.
- Verified a consumer can now `import { MetricObjectWithValues, MetricValue } from 'prom-client'` and annotate the `getMetricsAsJSON()` result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ This release marks our first release under the Prometheus umbrella.
 - perf: Stat aggregation uses similar strategy to collection. 60% faster aggregation
 - chore: Add copyright license headers and test
 - Make cluster and worker-thread metric aggregation order deterministic
+- Export `MetricObject`, `MetricObjectWithValues`, `MetricValue` and `MetricValueWithName` from the TypeScript definitions
 
 ### Added
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -279,7 +279,7 @@ export enum MetricType {
 
 type CollectFunction<T> = (this: T) => void | Promise<void>;
 
-interface MetricObject {
+export interface MetricObject {
 	name: string;
 	help: string;
 	type: MetricType;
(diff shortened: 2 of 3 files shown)


jdmarshall:
I think some TS unit tests would be great here.

milcho0604:
Thanks for taking a look, and good call on wanting test coverage.

One honest caveat I ran into while adding it: because TypeScript resolves top-level types in *type position* even without the `export` keyword, a compile-only test can't actually assert the `export` itself — removing `export` from these types still type-checks. So I couldn't write a test that fails purely on the export being dropped.

What I did add to `test/typescript.ts` is a usage example that names the return types of `getMetricsAsJSON()`, `getMetricsAsArray()`, and `Metric#get()` via the now-exported type names. It exercises the intended public surface and will catch the types being removed, renamed, or reshaped, and it's compiled by `tsc --project .` in CI.

For context, the real value of the explicit `export` here is discoverability rather than a compile fix: exported symbols are what IDE auto-import suggests and what API-doc tooling (typedoc/api-extractor) surfaces, so consumers can `import { MetricObjectWithValues }` intentionally instead of relying on TS's non-exported-type leniency. Happy to adjust the test if you'd prefer a different shape.

jdmarshall (approved):

PR #787, opened 2026-07-22
fix(cluster): skip responses after IPC disconnect

### LINKED ISSUE #563: [BUG] ERR_IPC_CHANNEL_CLOSED showing up on new deployments

alencristen:
Fixes #563.

Stops cluster workers from calling `process.send()` when metric collection finishes after their IPC channel has closed. Adds a regression test for a disconnect between request receipt and response.

Checks: `npm test`; focused cluster tests on Node 22, 24, and 26.

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -190,6 +190,7 @@ function addListeners() {
 			if (message.type === GET_METRICS_REQ) {
 				Promise.all(registries.map(r => r.getMetricsAsJSON()))
 					.then(metrics => {
+						if (!process.connected) return;
 						process.send({
 							type: GET_METRICS_RES,
 							requestId: message.requestId,
@@ -197,6 +198,7 @@ function addListeners() {
 						});
 					})
 					.catch(error => {
+						if (!process.connected) return;
 						process.send({
 							type: GET_METRICS_RES,
 							requestId: message.requestId,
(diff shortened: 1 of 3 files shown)


lib/cluster.js
```diff
@@ -190,13 +190,15 @@ function addListeners() {
 			if (message.type === GET_METRICS_REQ) {
 				Promise.all(registries.map(r => r.getMetricsAsJSON()))
 					.then(metrics => {
+						if (!process.connected) return;
```
jdmarshall, line 193: Given that #778 has landed, this would be another good spot for debug output

jdmarshall (changes requested):
As of a few PRs ago, completely silent catch blocks are no longer allowed. Please copy the debug behavior seen in other filev.

jdmarshall:
This has been cherry-picked into #789 to deal with some extensive changes to this code.

Issue #788, opened 2026-07-27
The 'memory leak' detection in cluster and worker aggregation prevents supporting prometheus and openmetrics in the same process


jdmarshall:
There are better ways to prevent memory leaks for event listeners in unit tests that don't clobber runtime functionality in the bargain.

jdmarshall:
The PR I have open for this breaks the per instance lookup tables into file-scoped, and relies on the fact that any parallel requests have distinct IDs which will end up routing the responses to the proper listener.

That could still have a problem with `watch` commands but is more of an issue when you're actually trying to debug telemetry instead of something else in your application. We could do something with a global variable to track the counter, so that reloads still result in monotonically increasing numbers.

Issue #501, opened 2022-04-22
AggregatorRegistry Metrics Error - Operation Timed out


austonpramodh:
I am trying to integrate prom client into a cluster-enabled project. Unfortunately, I don't get any response from the Aggregator endpoint. I do get responses from the "/metrics" from the worker nodes, but when I hit the "/cluster_metrics" I get an error saying "Operation Timed out".

**I would like to understand how the aggregator gets the metrics from the works, This might help me understand if I am doing and mistakes.**

## Master Server
<img width="773" alt="image" src="https://user-images.githubusercontent.com/5049050/164570771-8191833f-bb3a-43e2-999d-e11dbc850572.png">

## Worker Server - Already runs an HTTP server, therefore created a handler.
<img width="773" alt="image" src="https://user-images.githubusercontent.com/5049050/164570960-cb12fe6f-cf13-4a81-be35-d23549848730.png">

## Error
<img width="482" alt="Screen Shot 2022-04-21 at 8 13 22 PM" src="https://user-images.githubusercontent.com/5049050/164570831-3d7f6ae5-43cb-45ff-89a2-c2ec4072d294.png">

austonpramodh:
I found the solution for this, So looking at the [ClusterMtrics file](https://github.com/siimon/prom-client/blob/master/lib/cluster.js), It uses even emitter to collect the metrics from the workers. Therefore the AggregatorRegistry class needs to be instantiated on workers as well which I wasn't doing, I was just instantiating it in the master process.

![image](https://user-images.githubusercontent.com/5049050/165050271-4268d24a-6851-4f2d-b932-8b14c21945a8.png)

Thanks. Closing this issue.

zbjornson:
This was actually a mistake in a recent release, see #464. I haven't had any time to work on prom-client to address it, but you found the workaround.

SpComb:
We ran into the same issue as a regression, where cluster metrics stopped working after an upgrade from 13.1.0 -> 14.0.1.

Perhaps this issue should be re-opened, as #464 has not been merged, and there isn't yet any released version that fixes this issue?

zbjornson:
Since so much time has passed and since a few releases have been made since the regression, I'm going to try to find a non-breaking fix that allows either usage pattern.

morgaan:
I hoped that the latest major version (15) update would address that. It's not. Feel like I'm stuck with `13.1.0` for another while.
[...]

PR #789, opened 2026-07-28
Cluster fixes

### LINKED ISSUE #155: Cluster event listeners are attached even if not using clusters
Because of https://github.com/siimon/prom-client/blob/426941fbc22433fff932d8f9fb09ebba4818f514/index.js#L24, https://github.com/siimon/prom-client/blob/426941fbc22433fff932d8f9fb09ebba4818f514/lib/cluster.js#L131-L170 is always ran

This results in a memory leak if `require('prom-client');` is done multiples times and the `require.cache` is blown away, as might happen in tests.

I get the following after rerunning my tests, which blow away the require cache in order to mock a dep. `(node:80194) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 message listeners added. Use emitter.setMaxListeners() to increase limit`

jdmarshall:
Rework of worker and cluster lifecycle management. Also allows the primary cluster thread to report stats, adds debug information to common possible failure paths.

Fixes #155, #181, #183, #280, #501, #563, #788

### DIFF
--- a/example/cluster.js
+++ b/example/cluster.js
@@ -22,6 +22,10 @@ const metricsServer = express();
 const clusterRegistry = new ClusterRegistry();
 
 if (cluster.isPrimary) {
+	require('../').collectDefaultMetrics({
+		gcDurationBuckets: [0.001, 0.01, 0.1, 1, 2, 5], // These are the default buckets.
+	});
+
 	for (let i = 1; i <= 4; i++) {
 		cluster.fork({ ...process.env, PORT: 3000 + i });
 	}
--- a/example/cluster.js
+++ b/example/cluster.js
@@ -36,6 +36,7 @@ if (cluster.isPrimary) {
 			res.set('Content-Type', clusterRegistry.contentType);
 			res.send(metrics);
 		} catch (ex) {
+			console.error(ex);
 			res.statusCode = 500;
 			res.send(ex.message);
 		}
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -22,6 +22,7 @@
  * cluster master.
  */
 
+const { debuglog } = require('node:util');
 const Registry = require('./registry');
 // We need to lazy-load the 'cluster' module as some application servers -
 // namely Passenger - crash when it is imported.
@@ -31,20 +32,71 @@ let cluster = () => {
 	return data;
 };
 
+const debug = debuglog('prom:metrics:cluster');
+const ANNOUNCEMENT = '@prometheus-io/client:announcement';
 const GET_METRICS_REQ = '@prometheus-io/client:getMetricsReq';
 const GET_METRICS_RES = '@prometheus-io/client:getMetricsRes';
 
 let registries = [Registry.globalRegistry];
 let requestCtr = 0; // Concurrency control
-let listenersAdded = false;
 const requests = new Map(); // Pending requests for workers' local metrics.
 
 class AggregatorRegistry extends Registry {
+	/**
+	 * Create a Registry.
+	 * @param regContentType
+	 */
 	constructor(regContentType = Registry.PROMETHEUS_CONTENT_TYPE) {
 		super(regContentType);
-		addListeners();
+
+		if (cluster().isPrimary) {
+			this.workers = new Map();
+		}
+
+		addListeners(this);
 	}
 
+	addWorker(id) {
+		if (this.workers.has(id)) {
+			debug('duplicate worker announcement', id);
+			return;
+		}
+
+		const worker = cluster().workers[id];
+
+		worker.on('disconnect', () => {
+			debug('worker disconnected', id);
+			this.workers.delete(id);
+		});
+
+		worker.on('message', message => {
+			if (message.type === GET_METRICS_RES) {
+				const request = requests.get(message.requestId);
+
+				if (request === undefined) {
+					debug('unexpected results from worker', id);
+					return;
+				}
+
+				const response = request.responseHandlers.get(id);
+				if (response === undefined) {
+					return;
+				}
+				request.responseHandlers.delete(id);
+
+				if (message.error) {
+					response.reject(new Error(message.error));
+				} else {
+					response.resolve({
+						threadId: id,
+						metrics: message.metrics,
+					});
+				}
+			}
+		});
+
+		this.workers.set(id, worker);
+	}
 	/**
 	 * Gets aggregated metrics for all workers. The optional callback and
 	 * returned Promise resolve with the same value; either may be used.
@@ -53,9 +105,9 @@ class AggregatorRegistry extends Registry {
 	 */
 	clusterMetrics() {
 		const requestId = requestCtr++;
-		const workers = Object.values(cluster().workers)
-			.filter(worker => worker.isConnected())
-			.sort((left, right) => left.id - right.id);
+		const orderedWorkers = [...this.workers.values()].sort(
+			(left, right) => left.id - right.id,
+		);
 
 		return new Promise((resolve, reject) => {
 			let settled = false;
@@ -78,37 +130,37 @@ class AggregatorRegistry extends Registry {
 				responseHandlers,
 				done,
 				errorTimeout: setTimeout(() => {
-					const err = new Error('Operation timed out.');
+					const err = new Error(
+						`Operation timed out. ${request.responseHandlers.size} outstanding responses.`,
+					);
 					request.done(err);
-				}, 5000),
+				}, 5_000),
 			};
 			requests.set(requestId, request);
-
-			const message = {
-				type: GET_METRICS_REQ,
-				requestId,
-			};
-
-			if (workers.length === 0) {
-				// No workers were up
-				process.nextTick(() => done(undefined, ''));
-				return;
-			}
-
-			const responsePromises = workers.map(
+			const responsePromises = orderedWorkers.map(
 				worker =>
 					new Promise((resolveResponse, rejectResponse) => {
 						responseHandlers.set(worker.id, {
 							resolve: resolveResponse,
 							reject: rejectResponse,
 						});
-						worker.send(message);
+
+						worker.send({
+							type: GET_METRICS_REQ,
+							requestId,
+						});
 					}),
 			);
 
-			Promise.all(responsePromises)
-				.then(metrics => Registry.aggregate(metrics.flat()).metrics())
-				.then(result => done(undefined, result), done);
+			if (responsePromises.length === 0) {
+				debug('No workers found for requestId', requestId);
+				process.nextTick(() => done(undefined, ''));
+			} else {
+				Promise.all(responsePromises)
+					.then(responses => responses.flatMap(response => response.metrics))
+					.then(metrics => Registry.aggregate(metrics).metrics())
+					.then(result => done(undefined, result), done);
+			}
 		});
 	}
 
@@ -157,54 +209,51 @@ class AggregatorRegistry extends Registry {
  * than once).
  * @returns {void}
  */
-function addListeners() {
-	if (listenersAdded) return;
-	listenersAdded = true;
-
+function addListeners(registry) {
 	if (cluster().isPrimary) {
 		// Listen for worker responses to requests for local metrics
 		cluster().on('message', (worker, message) => {
-			if (message.type === GET_METRICS_RES) {
-				const request = requests.get(message.requestId);
-
-				if (request === undefined) {
-					return;
-				}
-
-				const response = request.responseHandlers.get(worker.id);
-				if (response === undefined) {
-					return;
-				}
-				request.responseHandlers.delete(worker.id);
-
-				if (message.error) {
-					response.reject(new Error(message.error));
-				} else {
-					response.resolve(message.metrics);
-				}
+			if (message.type === ANNOUNCEMENT) {
+				registry.addWorker(worker.id);
 			}
 		});
+
+		announce();
 	} else {
 		// Respond to master's requests for worker's local metrics.
-		process.on('message', message => {
-			if (message.type === GET_METRICS_REQ) {
-				Promise.all(registries.map(r => r.getMetricsAsJSON()))
-					.then(metrics => {
-						process.send({
-							type: GET_METRICS_RES,
-							requestId: message.requestId,
-							metrics,
-						});
-					})
-					.catch(error => {
-						process.send({
-							type: GET_METRICS_RES,
-							requestId: message.requestId,
-							error: error.message,
-						});
+		process.on('message', async message => {
+			if (message.type === ANNOUNCEMENT) {
+				process.send({ type: ANNOUNCEMENT });
+			} else if (message.type === GET_METRICS_REQ) {
+				const metrics = await Promise.all(
+					registries.map(r => r.getMetricsAsJSON()),
+				);
+
+				try {
+					process.send({
+						type: GET_METRICS_RES,
+						requestId: message.requestId,
+						metrics,
 					});
+				} catch (error) {
+					process.send({
+						type: GET_METRICS_RES,
+						requestId: message.requestId,
+						error: error.message,
+					});
+				}
 			}
 		});
+
+		process.send({ type: ANNOUNCEMENT });
+	}
+}
+
+function announce() {
+	for (const worker of Object.values(cluster().workers)) {
+		if (worker.isConnected()) {
+			worker.send({ type: ANNOUNCEMENT });
+		}
 	}
 }
 
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -39,7 +39,9 @@ const GET_METRICS_RES = '@prometheus-io/client:getMetricsRes';
 
 let registries = [Registry.globalRegistry];
 let requestCtr = 0; // Concurrency control
+let listenersAdded = false;
 const requests = new Map(); // Pending requests for workers' local metrics.
+const workers = new Map();
 
 class AggregatorRegistry extends Registry {
 	/**
@@ -49,54 +51,9 @@ class AggregatorRegistry extends Registry {
 	constructor(regContentType = Registry.PROMETHEUS_CONTENT_TYPE) {
 		super(regContentType);
 
-		if (cluster().isPrimary) {
-			this.workers = new Map();
-		}
-
-		addListeners(this);
+		addListeners();
 	}
 
-	addWorker(id) {
-		if (this.workers.has(id)) {
-			debug('duplicate worker announcement', id);
-			return;
-		}
-
-		const worker = cluster().workers[id];
-
-		worker.on('disconnect', () => {
-			debug('worker disconnected', id);
-			this.workers.delete(id);
-		});
-
-		worker.on('message', message => {
-			if (message.type === GET_METRICS_RES) {
-				const request = requests.get(message.requestId);
-
-				if (request === undefined) {
-					debug('unexpected results from worker', id);
-					return;
-				}
-
-				const response = request.responseHandlers.get(id);
-				if (response === undefined) {
-					return;
-				}
-				request.responseHandlers.delete(id);
-
-				if (message.error) {
-					response.reject(new Error(message.error));
-				} else {
-					response.resolve({
-						threadId: id,
-						metrics: message.metrics,
-					});
-				}
-			}
-		});
-
-		this.workers.set(id, worker);
-	}
 	/**
 	 * Gets aggregated metrics for all workers. The optional callback and
 	 * returned Promise resolve with the same value; either may be used.
@@ -105,9 +62,9 @@ class AggregatorRegistry extends Registry {
 	 */
 	clusterMetrics() {
 		const requestId = requestCtr++;
-		const orderedWorkers = [...this.workers.values()].sort(
-			(left, right) => left.id - right.id,
-		);
+		const orderedWorkers = [...workers.values()]
+			.filter(worker => worker.isConnected())
+			.sort((left, right) => left.id - right.id);
 
 		return new Promise((resolve, reject) => {
 			let settled = false;
@@ -216,44 +173,106 @@ class AggregatorRegistry extends Registry {
  * than once).
  * @returns {void}
  */
-function addListeners(registry) {
+function addListeners() {
+	if (listenersAdded) {
+		return;
+	}
+
+	listenersAdded = true;
+
 	if (cluster().isPrimary) {
-		// Listen for worker responses to requests for local metrics
-		cluster().on('message', (worker, message) => {
-			if (message.type === ANNOUNCEMENT) {
-				registry.addWorker(worker.id);
-			}
-		});
+		replaceListener('message', cluster(), primaryListener);
+		replaceListener('disconnect', cluster(), disconnect);
 
 		announce();
 	} else {
-		// Respond to master's requests for worker's local metrics.
-		process.on('message', async message => {
-			if (message.type === ANNOUNCEMENT) {
-				process.send({ type: ANNOUNCEMENT });
-			} else if (message.type === GET_METRICS_REQ) {
-				const metrics = await Promise.all(
-					registries.map(r => r.getMetricsAsJSON()),
-				);
+		replaceListener('message', process, workerListener);
 
-				try {
-					process.send({
-						type: GET_METRICS_RES,
-						requestId: message.requestId,
-						metrics,
-					});
-				} catch (error) {
-					process.send({
-						type: GET_METRICS_RES,
-						requestId: message.requestId,
-						error: error.message,
-					});
-				}
-			}
-		});
-
-		process.send({ type: ANNOUNCEMENT });
+		if (typeof process.send !== 'function') {
+			debug('worker has no process.send()');
+		} else if (!process.connected) {
+			debug('worker is not connected to parent process');
+		} else {
+			process.send({ type: ANNOUNCEMENT });
+		}
 	}
+}
+
+/**
+ * Watch for metrics events and aggregator announcements
+ *
+ * Whereas clusters are a top-level activity, multiple modules may start their
+ * own workers and require telemetry collection.
+ * @param	message {MessageEvent}
+ */
+async function workerListener(message) {
+	if (message.type === ANNOUNCEMENT) {
+		process.send({ type: ANNOUNCEMENT });
+	} else if (message.type === GET_METRICS_REQ) {
+		try {
+			const metrics = await Promise.all(
+				registries.map(r => r.getMetricsAsJSON()),
+			);
+
+			process.send({
+				type: GET_METRICS_RES,
+				requestId: message.requestId,
+				metrics,
+			});
+		} catch (error) {
+			process.send({
+				type: GET_METRICS_RES,
+				requestId: message.requestId,
+				error: error.message,
+			});
+		}
+	}
+}
+
+/**
+ * Add workers to the aggregation list when they are announced.
+ *
+ * Whereas clusters are a top-level activity, multiple modules may start their
+ * own workers and require telemetry collection.
+ * @param	event {MessageEvent}
+ */
+
+async function primaryListener(worker, event) {
+	if (event.type === ANNOUNCEMENT) {
+		if (workers.has(worker.id)) {
+			debug('duplicate worker announcement', worker.id);
+			return;
+		}
+
+		workers.set(worker.id, worker);
+	} else if (event.type === GET_METRICS_RES) {
+		const request = requests.get(event.requestId);
+
+		if (request === undefined) {
+			debug('unexpected results from worker', worker.id);
+			return;
+		}
+
+		const response = request.responseHandlers.get(worker.id);
+		if (response === undefined) {
+			return;
+		}
+		request.responseHandlers.delete(worker.id);
+
+		if (event.error) {
+			response.reject(new Error(event.error));
+		} else {
+			response.resolve({
+				threadId: worker.id,
+				metrics: event.metrics,
+			});
+		}
+	}
+}
+
+function disconnect(event) {
+	debug('worker disconnected', event.id);
+	workers.delete(event.id);
 }
 
 function announce() {
@@ -262,6 +281,29 @@ function announce() {
 			worker.send({ type: ANNOUNCEMENT });
 		}
 	}
+}
+
+/**
+ * Replace any listeners with new ones.
+ *
+ * @param messageType
+ * @param emitter {EventEmitter}
+ * @param fn
+ */
+function replaceListener(messageType, emitter, fn) {
+	// Reloading a module creates a unique instance of each function, so the
+	// identity checks is cluster.off() will fail.
+	const functionString = fn.toString();
+
+	for (const listener of emitter.listeners(messageType)) {
+		// eslint-disable-next-line eqeqeq
+		if (functionString == listener) {
+			debug('removing duplicate listener', messageType);
+			emitter.off(messageType, listener);
+		}
+	}
+
+	emitter.on(messageType, fn);
 }
 
 module.exports = AggregatorRegistry;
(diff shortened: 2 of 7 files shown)


jdmarshall:
This changes unreleased API modifications for AggregatorRegistry, so this will need to be part of v0.16.

Also fixes a memory leak in trunk for broadcastchannels, and improves lifecycle management for dead worker threads (both kinds)

lib/cluster.js
```diff
@@ -158,54 +174,102 @@ class AggregatorRegistry extends Registry {
...
+				threadId: worker.id,
+				metrics: event.metrics,
+			});
+		}
+	}
+}
+
+function announce() {
+	Object.values(cluster().workers)
+		.filter(worker => worker.isConnected())
```
cjihrig: You could skip the `filter()` loop and check `worker.isConnected()` as an `if` statement in the `forEach()`. It would be just a tiny bit more efficient.

lib/registry.js
```diff
@@ -253,8 +253,11 @@ class Registry {
 		aggregatedRegistry.setContentType(registryType);
 
 		// Gather by name
-		metricsArr.forEach(metrics => {
-			metrics.forEach(metric => {
+		metricsArr.forEach((metrics, i) => {
+			metrics.forEach((metric, j) => {
+				if (metric.name === undefined) {
+					console.warn('wtf', i, j);
```
cjihrig: Is this left over?

test/workerTest.js
```diff
@@ -53,16 +54,22 @@ describe.each([
...
+						name,
 						threadId,
-						channel: new BroadcastChannel(name),
-					};
+					});
+
+					return { threadId, channel };
 				});
 
+				await delay(5); // Let announcements arrive
```
cjihrig, line 70: Just speaking from past experience, this seems like it could become a flaky test in the future (on slow CI machines).

lib/cluster.js
```diff
@@ -78,37 +87,44 @@ class AggregatorRegistry extends Registry {
...
+							requestId,
+						});
 					}),
 			);
 
-			Promise.all(responsePromises)
-				.then(metrics => Registry.aggregate(metrics.flat()).metrics())
-				.then(result => done(undefined, result), done);
+			const myMetrics = Promise.all(
+				registries.map(async r => r.getMetricsAsJSON()),
```
cjihrig: It looks like you do something similar in a couple other places in this PR, and those places don't add the extra `async` to the map callback.

lib/cluster.js
jdmarshall: This shouldn't be called very often but sure.

lib/registry.js
jdmarshall: There's a reason this code now contains debug messages. The tests were excessively mocked and every change resulted in timeouts.

test/workerTest.js
jdmarshall: I'll check if there's  callback.

jdmarshall: This test already includes delay(), btw.

jdmarshall: No callback. Might be stuck with this one.

cjihrig (approved):
LGTM, but there are a couple linter issues in lib/worker.js.

lib/worker.js
```diff
@@ -222,50 +168,114 @@ class WorkerRegistry extends Registry {
...
+/**
+ * Watch for metrics events and aggregator announcements
+ *
+ * Whereas clusters are a top-level activity, multiple modules may start their
+ * own workers and require telemetry collection.
+ * @param	event {MessageEvent}
+ */
+async function workerListener(event) {
+	const name = `@prometheus/client:worker:${threadId}`;
+	const channel = new BroadcastChannel(name).unref();
```
krajorama, line 188: Wouldn't this  create and leak a BroadcastChannel on every event?

lib/cluster.js
```diff
@@ -158,53 +174,108 @@ class AggregatorRegistry extends Registry {
...
+ *
+ * Whereas clusters are a top-level activity, multiple modules may start their
+ * own workers and require telemetry collection.
+ * @param	message {MessageEvent}
+ */
+async function workerListener(message) {
+	if (message.type === ANNOUNCEMENT) {
+		process.send({ type: ANNOUNCEMENT });
+	} else if (message.type === GET_METRICS_REQ) {
+		const metrics = await Promise.all(
```
krajorama: What happens if await throws an exception, this is outside the `try` so possibly unhandled error?

lib/cluster.js
```diff
@@ -158,53 +174,108 @@ class AggregatorRegistry extends Registry {
...
+			workers.delete(deadWorker.id);
+		});
 
-				if (request === undefined) {
-					return;
-				}
+		announce();
+	} else {
+		process.on('message', workerListener);
+		process.send({ type: ANNOUNCEMENT });
```
krajorama: The non-primary branch calls process.send unconditionally. With cluster.isPrimary === false but no IPC channel (NODE_UNIQUE_ID leaked into a grandchild — the case some process managers hit), this throws where main constructs fine:


PR:   THREW: TypeError: process.send is not a function
main: OK: constructed without throwing
main never called send at construction time, so this is a new failure mode. Guard with typeof process.send === 'function' && process.connected.

lib/worker.js
```diff
@@ -146,7 +97,7 @@ class WorkerRegistry extends Registry {
 				}, 5_000),
 			};
 			requests.set(requestId, request);
-			const responsePromises = [...this.channels.keys()].map(
+			const responsePromises = [...channels.keys()].sort().map(
```
krajorama: `[...channels.keys()].sort()` sorts channel names, so with ≥10 threads you get `1, 10, 11, 2, 20, 3`. lib/cluster.js sorts *numerically* on worker.id; this undoes the numeric intent of 0162ab5 ("aggregate metrics in a deterministic order") on the worker-threads side. Still deterministic, but no longer matching cluster.js. The threadId now carried in the response payload is also unused for ordering — keep a name→threadId map and sort on that.

krajorama (commented):
I'm not a frontend developer , but trying to help with review using LLM.

I think in general smaller PRs are easier to review and argue about.

Findings:

### index.d.ts 
index.d.ts still declares the removed addWorker — index.d.ts:209

addWorker(worker: Worker): void; no longer exists at runtime. TypeScript users compile clean and get a TypeError. Delete the declaration.

### Primary metric inclusion is breaking

But filed under ### Changed — CHANGELOG.md:44

clusterMetrics() now always aggregates the primary's registries — confirmed: with zero workers it returns the primary's process_cpu_seconds_total etc. Calling collectDefaultMetrics() unconditionally in both primary and workers is a very common pattern, and those users will silently start seeing the primary's process metrics summed into every series. This is released API (v15.x), so it belongs under ### Breaking with a migration note.

It also deserves an opt-out. WorkerRegistry takes a primary flag; ClusterRegistry has no equivalent knob. Two smaller consequences: setRegistries's doc comment (lib/cluster.js:154 still says "Call from workers to…", and the empty-cluster return value changed from '' to '\n' — which is what forced the .trim() at test/clusterTest.js:76 and is worth its own changelog line.

### Unreachable branch

lib/cluster.js:119

allMetrics = [myMetrics, ...workerMetrics] always has ≥1 element, so allMetrics.length === 0, the debug('No workers found…') call, and the '' fast path are dead. Test workerMetrics.length or drop the branch.

### The worker-threads half of "lifecycle for dead worker threads (both kinds)" isn't there

lib/cluster.js gained a disconnect handler; lib/worker.js gained nothing. The 'close' listener at lib/worker.js:247only fires when the local channel object is closed, never when the remote thread dies. Verified: after worker.terminate(), every subsequent workerMetrics() rejects after 5 s, permanently. This is pre-existing (identical on main) so not a regression — but given the PR's stated scope, it's the obvious gap.

### The two new "listeners don't accumulate" tests cannot fail

test/clusterTest.js:79, test/workerTest.js:140

The require sits outside the loop, so jest.resetModules() inside it never produces a fresh module instance; ar is assigned and never read (also a lint smell); and there are no assertions. Measured:


test-style loop        -> cluster message listeners: 1
fresh module each time -> cluster message listeners: 6
So the property under test isn't guaranteed by listenersAdded at all — it's guaranteed by the require cache. Re-require inside the loop and assert cluster.listenerCount('message').

### Test hygiene in test/clusterTest.js line 101

 the originalWorkers restore was dropped, so the global cluster.workers stays mocked for the rest of the file (later announce() calls then send to jest mocks).
line 89: jest.resetModules() followed by new Registry(regType) from the top-of-file require is a no-op.
line 147: gauge.remove() isn't in a finally — a failed assertion leaves primary_gauge_test in the global registry, polluting later tests.
Nothing covers disconnect pruning workers, duplicate-announcement dedup, or collector-error propagation.

lib/cluster.js
jdmarshall: Hmm. Good catch. I think I might have to resurrect some mocks from another contributor to get code coverage on this change.  Let me see if I can find it..

lib/cluster.js
jdmarshall: This is a bug in the original code, but I'm surprised I missed it after the refactor to remove the promise chaining.

I staged a long campaign to migrate chained code to async/await on my last project and it was exactly little nuggets like this and incorrect `.catch()` chaining that were my arguments against tolerating them in any code paths getting new features or with proven concurrency bugs. I must be getting soft.

jdmarshall:
Something in this branch is now fighting with the PRs that have been landed in main.  Trying to work out what it is.

Incorrect replace string with constant, and a weird thing trying to reduce the number of mocks in someone else's tests.

lib/worker.js
jdmarshall: That's a regression. I split the listener functions in an attempt to do `removeListener()` calls but the module reload logic also changes the identity of the functions so that was moot. I'll see what should be done here. Either 

- put it back
- nested functions with the BroadcastChannel declared in the closure
- pass in channel as an argument

jdmarshall:
> The worker-threads half of "lifecycle for dead worker threads (both kinds)" isn't there

According to the Nodejs docs:

> The 'close' event is emitted once either side of the channel has been disconnected.

Furthermore, we are screwed here because NodeJS elected NOT to implement BroadcastChannel the way it was specified in HTML5 - the worker field on the events is unpopulated (set to null). I filed this as https://github.com/nodejs/node/issues/59053 and someone is currently bargaining to implement a version that still does not include a Worker object and so we would still have no access beyond the close event. 

I will see if I can confirm that the AI is hallucinating. I think it might be.

It is not. But as this was an existing error, I think I'll be doing a separate PR for that. It's a sticky problem given https://github.com/nodejs/node/issues/59053

jdmarshall:
Ooof. Cleaning up excess event listeners on cluster left behind after `jest.resetModules()` is paradoxically causing `cluster.emit()` to fire even later rather than sooner. A bunch of the tests treat it as if it's a synchronous call. It appears that it is not.

lib/worker.js
jdmarshall: Still working on this one FYI.

jdmarshall: Alright, that should be it. I also fixed some of the linter suggestions about duplicate variable names.

example/cluster.js
```diff
@@ -32,6 +36,7 @@ if (cluster.isPrimary) {
 			res.set('Content-Type', clusterRegistry.contentType);
 			res.send(metrics);
 		} catch (ex) {
+			console.log(ex);
```
krajorama: nit: 
```suggestion
			console.error(ex);
```

lib/cluster.js
```diff
@@ -158,54 +174,145 @@ class AggregatorRegistry extends Registry {
...
-				const request = requests.get(message.requestId);
+		replaceListener('message', cluster(), primaryListener);
+		replaceListener('disconnect', cluster(), disconnect);
 
-				if (request === undefined) {
-					return;
-				}
+		announce();
+	} else {
+		replaceListener('message', process, workerListener);
```
krajorama, line 189: I've been trying to understand an issue flagged by LLM. "replaceListener" was removing listeners from the "old" registry, but that means you can lose metrics as they won't be called anymore. So reverting to simply doing `on()` without `replaceListener` fixes that, see diff and regression test. Is this what you also flagged [here](https://github.com/prometheus/client_js/issues/155#issuecomment-5097906398) ? 

It does reintroduce #155 warning in "listeners don't accumulate" in test/clusterTest.js, but there is a way to fix it apparently (in a separate PR):
> Removing replaceListener does re-expose #155's original symptom — 11+ reloads of the module in one process will emit MaxListenersExceededWarning again. That's the honest trade, and it's the right one: a warning in a test harness that clears require.cache is strictly better than silent metric loss in production. If the author wants both, the fix is a shared registration marker (a symbol on the emitter, or a globalThis key) that skips installing a second listener rather than removing the first — skipping leaves the earlier instance's workers map intact, so no instance goes blind.

```suggestion
		cluster().on('message', primaryListener);
		cluster().on('disconnect', disconnect);

		announce();
	} else {
		process.on('message', workerListener);
```


Regression test for this:
```
		it('keeps polling workers after a duplicate module instance is loaded', async () => {
			const originalWorkers = cluster.workers;

			jest.resetModules();
			const FirstInstance = require('../lib/cluster');
			const registry = new FirstInstance(regType);

			// A duplicated copy of the package - a nested dependency, say - loads
			// its own module instance in the same primary process.
			jest.resetModules();
			const SecondInstance = require('../lib/cluster');
			new SecondInstance(regType);

			const worker = { id: 1, isConnected: () => true, send: jest.fn() };
			cluster.workers = { 1: worker };
			cluster.emit('message', worker, { type: ANNOUNCEMENT });

			let result;
			try {
				result = registry.clusterMetrics();

				const requests = worker.send.mock.calls
					.map(([message]) => message)
					.filter(message => message.type === GET_METRICS_REQ);

				// The first instance must still know about the worker. Otherwise it
				// silently reports primary-only metrics, with no timeout or error.
				expect(requests).toHaveLength(1);
			} finally {
				cluster.emit('message', worker, {
					type: GET_METRICS_RES,
					requestId: 0,
					metrics: [[metric(1)]],
				});
				await result?.catch(() => {});
				cluster.emit('disconnect', worker);
				cluster.workers = originalWorkers;
			}
		});
```

lib/worker.js
```diff
@@ -146,15 +97,17 @@ class WorkerRegistry extends Registry {
...
+				.sort((left, right) => +right - +left)
+				.map(
+					name =>
+						new Promise((resolveResponse, rejectResponse) => {
+							responseHandlers.set(name, {
+								resolve: resolveResponse,
+								reject: rejectResponse,
+							});
+						}),
+				);
```
krajorama: Here `left` and `right` are channel names, so their numeric value is `NaN` which means no sort.
Let's do the same as for workers in cluster.js and store the threadId.

```suggestion
			const orderedWorkers = [...channels.values()].sort(
				(left, right) => left.threadId - right.threadId,
			);
			const responsePromises = orderedWorkers.map(
				({ name }) =>
					new Promise((resolveResponse, rejectResponse) => {
						responseHandlers.set(name, {
							resolve: resolveResponse,
							reject: rejectResponse,
						});
					}),
			);
```

lib/worker.js
```diff
@@ -265,7 +213,63 @@ function addListeners(registry) {
...
+	if (message.type === ANNOUNCEMENT) {
+		const workerName = message.name;
+
+		if (channels.has(workerName)) {
+			debug('duplicate worker announcement', workerName);
+			return;
+		}
+
+		const workerChannel = new BroadcastChannel(workerName, {}).unref();
+		channels.set(workerName, workerChannel);
```
krajorama: For the sort above.

```suggestion
		const workerChannel = new BroadcastChannel(workerName).unref();
		channels.set(workerName, {
			name: workerName,
			threadId: message.threadId,
			channel: workerChannel,
		});
```

test/workerTest.js
```diff
@@ -93,21 +99,35 @@ describe.each([
...
 		it('does not error out on unexpected (or late) responses', () => {
 			jest.resetModules();
 
 			const WorkerRegistry = require('../lib/worker');
-
 			const registry = new WorkerRegistry(regType);
-			const emitter = new EventEmitter();
-
-			registry.addWorker(emitter);
+			const announcementChannel = new BroadcastChannel(ANNOUNCEMENT).unref();
```
krajorama: Isn't `ANNOUNCEMENT` a message type , but not a channel name? Should be instead:

```suggestion
			const announcementChannel = new BroadcastChannel(
				'@prometheus-io/client:announce',
			).unref();
```

lib/cluster.js
```diff
@@ -158,54 +174,145 @@ class AggregatorRegistry extends Registry {
...
+	for (const listener of emitter.listeners(messageType)) {
+		// eslint-disable-next-line eqeqeq
+		if (functionString == listener) {
+			debug('removing duplicate listener', messageType);
+			emitter.off(messageType, listener);
+		}
 	}
+
+	emitter.on(messageType, fn);
 }
```
krajorama, line 317: For above.

```suggestion
```

krajorama (commented):
Getting close now. I'm still using LLM, but also taking the time to understand what's going on, so the comments are my own, except where quoted.

lib/cluster.js
jdmarshall: Major edit:

Because the responses are being aggregated through a promise, only the first result was ever being seen anyway. In fact what you were probably always seeing before was the oldest or second oldest metrics per process, based on when the event was delivered and processing time to gather the metrics.  Which is exactly the wrong data for functional and integration tests. 

To the best of my knowledge prom-client has never worked with hot reload.  Let alone well. And anyone would see that it doesn't within a few minutes of trying, especially if they used older versions that were especially crabby about this.

We have a bigger problem with what to do about dead workers. Because their metrics disappear when they do, and since we are gathering them, we are getting the wrong answers for counts and gauges. #803 which is a problem since the general wisdom is 'let the process crash' when unhandledException or unhandledRejection fires.

What I think that suggests is an update to the README, suggesting you let a Prometheus sidecar handle the aggregation in Serious Projects rather than using cluster.js or worker.js

lib/worker.js
jdmarshall: #$@&!  I haven’t looked at this PR for two weeks and have been pushing changes and looking at bugs on three other repos the entire time. Half of this has slipped out of my head now.

test/workerTest.js
jdmarshall: I thought I caught this one. Maybe it happened twice.  It’s odd the test still passed, but that’s always the problem with editing negative tests.  Particularly when they’re convoluted enough that red-green isn’t a no-brainer.

test/workerTest.js
```diff
@@ -93,21 +99,37 @@ describe.each([
...
 
 			const WorkerRegistry = require('../lib/worker');
-
 			const registry = new WorkerRegistry(regType);
-			const emitter = new EventEmitter();
-
-			registry.addWorker(emitter);
+			const announcementChannel = new BroadcastChannel(
+				'@prometheus-io/client:announce',
+			);
```
krajorama: woops, I missed this:

```suggestion
			).unref();;
```

lib/cluster.js
krajorama: I'm sorry I'm having a hard time understanding this :( I'll let this go after this.

>Because the responses are being aggregated through a promise, only the first result was ever being seen anyway.

In the proposed test, there's no first/second result, it's one result per instance. Also using `on()` doesn't change that we have a promise which still keeps only one result if there were multiple. Shouldn't some test break if using `on()` changed something about handling multiple messages? E.g. `does not error out on unexpected (or late) responses` or `aggregates worker responses in worker id order` test?

The proposed test is about multiple instance at the same time sending just one message and in that case `replaceListener` makes us completely lose one set of results.


I've asked the LLM to show what's the worse that can happen if we keep replaceListener and under what circumstances, I'm ok with approving this PR and filing an issue instead based on this:

# PR #789 — risk of keeping `replaceListener`

Assessment of the worst-case outcome if [prometheus/client_js#789](https://github.com/prometheus/client_js/pull/789)
is approved with `replaceListener` retained in `lib/cluster.js`.

Tested against commit `1adc9d5` ("Simplify process.send sanity checks."), Node v20.9.0,
compared against the same tree with `replaceListener` reverted to plain `on()`.

---

The worst case is worse than "missing metrics" — it's a **crash of the cluster primary**,
and it's deterministic.

## The failure chain

`replaceListener` strips *both* listeners from the earlier instance — `message` **and**
`disconnect`. That second one is what escalates this. Reproduced 3/3, app registry
constructed first, a nested copy loading later:

```
before 2nd copy loads: wc_counter=2                      ← healthy
--- second module instance constructed ---
after  2nd copy loads: REJECTED: Operation timed out. 2 outstanding responses.
after worker death:    Error [ERR_IPC_CHANNEL_CLOSED]: Channel closed
                       Emitted 'error' event on Worker instance
                       Node.js v20.9.0                   ← process exits
```

Three stages:

1. **Every scrape times out at 5 s.** The earlier instance still has both workers in its
   `workers` map, but responses no longer route to it. Worse than the empty-map case —
   that returned fast and wrong; this hangs the metrics endpoint for 5 s and then 500s.
2. **Dead workers are never pruned**, because the `disconnect` listener went with the
   `message` one.
3. **`worker.send()` on a dead worker** emits `ERR_IPC_CHANNEL_CLOSED` as an `'error'`
   event on the Worker object. Nothing listens for it, so Node throws and the primary
   exits — taking the whole cluster down, since the primary is the process manager.

Same scenario with plain `on()`: `wc_counter=1` in 2 ms, correctly reflecting the
surviving worker.

There's an uncomfortable detail here — the PR body says **"Fixes #563"**, and
[#563](https://github.com/prometheus/client_js/issues/563) is *"[BUG] ERR_IPC_CHANNEL_CLOSED
showing up on new deployments"*. `replaceListener` can produce precisely that error, fatally.

## How likely is it

All of these must hold:

- Two copies of the package as separate module instances, with **byte-identical**
  `primaryListener` source — different versions won't dedup, so this needs same-version
  duplication.
- **Both construct a `ClusterRegistry` in the primary.** Requiring twice is harmless;
  `addListeners()` only runs from the constructor. This is the narrow link — most apps
  construct once. It opens up when a framework plugin or APM wrapper instruments
  alongside app code.
- The app scrapes via the earlier-constructed instance (the later one wins the listener),
  so it's roughly a coin flip on init order.
- For the crash specifically, a worker then has to die — which for long-running clusters
  is a matter of time, and is the normal response to `unhandledRejection`.

So: unlikely to hit most users, near-certain for anyone who does hit it, and it recurs on
every boot with that dependency tree. Detectability is poor — the only signal is a
`debug()` line behind `NODE_DEBUG=prom:metrics:cluster`, and the symptoms point nowhere
near prom-client.

## If you approve anyway

Reasonable position — the PR is a large net improvement, and it targets unreleased v0.16,
so there's runway. Two things worth asking for as a condition, both cheap:

**Restore the `isConnected()` filter** in `clusterMetrics()` — the
`.filter(worker => worker.isConnected())` this PR dropped. Tested: it doesn't fix the
metric loss, but it does downgrade the crash to a timeout:

```
after worker death, scrape#1: 5002ms  REJECTED: Operation timed out.
after worker death, scrape#2: 5006ms  REJECTED: Operation timed out.
```

A broken metrics endpoint is survivable; a dead primary isn't. This is worth having
regardless of how the `replaceListener` argument lands, since the same
`send()`-to-a-dead-worker race exists in a narrower window even with plain `on()`.

**File a follow-up issue** with the reproduction, so the decision is recorded rather than
lost in a review thread.

For what it's worth, the actual fix stays small — `−26/+3`, no interaction with anything
else in the PR — so "approve with `on()`" costs the author very little compared to
shipping a known primary-crash path.

---

## Appendix: reproduction

Not part of the original assessment — included because the scratch copies used above are
session-temporary. Requires two trees: one as the PR stands, one with
`replaceListener('message', cluster(), primaryListener)` /
`replaceListener('disconnect', cluster(), disconnect)` /
`replaceListener('message', process, workerListener)` reverted to the equivalent
`.on(...)` calls.

```js
// worstcase.js — run as: VARIANT=/path/to/tree node worstcase.js
// App constructs its registry first; a library loads its own copy LATER.
// By then the app's instance already knows about the workers.
const cluster = require('cluster');
const P = process.env.VARIANT;
const path = P + '/lib/cluster.js';

if (cluster.isPrimary) {
	const A = require(path);
	const regA = new A(); // app's registry
	const workers = [cluster.fork(), cluster.fork()];

	(async () => {
		await new Promise(r => setTimeout(r, 700)); // workers announce to A
		let m = await regA.clusterMetrics().catch(e => `REJECTED: ${e.message}`);
		console.log(
			`  before 2nd copy loads: wc_counter=${(String(m).match(/^wc_counter (\S+)/m) || [])[1] ?? 'ABSENT'}`,
		);

		// A nested dependency initialises its own copy now.
		delete require.cache[require.resolve(path)];
		const B = require(path);
		new B();
		console.log('  --- second module instance constructed ---');

		m = await regA.clusterMetrics().catch(e => `REJECTED: ${e.message}`);
		console.log(
			`  after  2nd copy loads: wc_counter=${(String(m).match(/^wc_counter (\S+)/m) || [])[1] ?? String(m).slice(0, 60)}`,
		);

		// Now one worker dies. A never pruned it, because A lost its disconnect listener.
		process.kill(workers[0].process.pid, 'SIGKILL');
		await new Promise(r => setTimeout(r, 600));
		for (const scrape of [1, 2]) {
			const t = Date.now();
			m = await regA.clusterMetrics().catch(e => `REJECTED: ${e.message}`);
			const v = (String(m).match(/^wc_counter (\S+)/m) || [])[1];
			console.log(
				`  after worker death, scrape#${scrape}: ${String(Date.now() - t).padStart(4)}ms  ${v ? `wc_counter=${v}` : String(m).slice(0, 55)}`,
			);
		}
		for (const w of workers)
			try {
				w.kill();
			} catch {}
		process.exit(0);
	})();
} else {
	const { Counter } = require(P + '/index.js');
	new Counter({ name: 'wc_counter', help: 'h' }).inc(1);
	const AR = require(path);
	new AR();
	setInterval(() => {}, 1000);
}
```

krajorama (commented):
A small nit and one more round about the replaceListener

test/workerTest.js
jdmarshall: No I cut and pasted that from one spot to another which means it's probably in two spots...

Update: yep, it is.

jdmarshall:
Can you and I have this conversation instead of me and the LLM through you? 

My response got longer than yours and ends up repeating comments I already made, so here's the chopped
version which is still unfortunately long. As you said, this PR is already long, so any bugs not already in scope
are going to need to be a separate PR.

```
Three stages:

    Every scrape times out at 5 s. The earlier instance still has both workers in its
    workers map, but responses no longer route to it. Worse than the empty-map case —
    that returned fast and wrong; this hangs the metrics endpoint for 5 s and then 500s.
```
That bug already exists on trunk. Twice. That's why I filed #803. There are unreleased changes on
main that reduce the surface area of this on cluster.js (the new connected checks) but there is no
analog for Workers. So #803.

We are not going to be able to support hot reload in production with this library any time soon. It has
never worked and people have acknowledged they know that across multiple Issues in the backlog. It is
not a feature. **It will not be a feature**, because making it so would in fact be exactly the opposite of
what the current functionality is attempting to do: nuke accumulated telemetry data and globally 
unique objects between reloads.

That fact should just go right in the docs and will for #803. I spent a couple years working with the
Worker API in batch processing systems and failed repeatedly to translate those good experiences
to production, online systems. IPC overhead was the largest blocker but the second most important
issue was this:

The Worker APIs in Node do not currently compose.

You can use them to do exactly one thing. Once you start trying to layer cross-cutting concerns like
telemetry, reloadable config, or security/audit on top of those tasks, you're gonna have a bad time with
geometric complexity of your code. So we are having a bad time. Supporting hot reload in production 
would be a second cross-cutting concern, and would quadruple the amount of work. Not just for me, 
but for the users, because they would have to rearchitect their apps and their bootstrapping to be able
to constantly feed us the information we need to keep their lifecycles and ours in sync. 

With piscina, you have a tool that at least can handle auto-scaling, and some process lifecycle bookkeeping,
so I recommend it to anyone who can't or won't invest in a first rate education in concurrent or distributed
programming. I actually tried to interest the piscina author in trying to fix the composition problem more
generally, but I kind of asked it obliquely and now that I'm typing this I may try a second time. 

```
    Dead workers are never pruned, because the disconnect listener went with the
    message one.
    worker.send() on a dead worker emits ERR_IPC_CHANNEL_CLOSED as an 'error'
    event on the Worker object. Nothing listens for it, so Node throws and the primary
    exits — taking the whole cluster down, since the primary is the process manager.
```

This is hallucination. There is no `worker.send()`, and we couldn't call it even if we wanted to (
which I very much do). See above about composition.  Your LLM is also assuming that the worker
features have already shipped. That's a reasonable enough assumption to make, but we have in fact
shipped nothing for 2 years.

And also 'disconnect' never actually fires, except in the tests. We will need to document teardown procedures
and probably provide a function to facilitate making that happen.

And we are in fact going to have to keep data around about dead workers. Because otherwise all aggregate
statistics will behave incorrectly. I'll have to do something about that in #803 as well. Which is going to suck, 
especially for transient communication failures since we have no actual way to detect a Worker with a high
event loop utilization. Which is basically what you use Workers for. 

And keeping that data around either 1) precludes hot reloading of prom-client at all, or 2) requires storing this data and the counter in global memory so that it can't be wiped by a reload. The whole point of the module-local storage being used in this code is to facilitate test and debug-time wiping of telemetry data, NOT for supporting applications using hot-reloading of prom-client specifically (They can, I believe, hot reload their own stuff since it will just require or import the original copy of prom-client)

jdmarshall:
@krajorama  After some initial panic about scope creep in #803, I think I have a much smaller scope that will require some cooperation from our users in order to use prom-client. We will have to ask them to wire in some calls to us in their process lifecycle management code. In most applications, the stats bootstrapping and the process signals are zero to two degrees of separation from each other, but most often one, so the bookkeeping shouldn't be onerous.

If Node ever fixes BroadcastChannel then we can step in to facilitate listening to some of the events automatically. But that still won't help us with cluster.js (although some day I hope to deprecate that feature).

I have marked #803 as the v0.16 milestone and as urgent. I should be able to get to it by tomorrow.

jdmarshall:
> Restore the isConnected() filter in clusterMetrics() — the
> .filter(worker => worker.isConnected()) this PR dropped. Tested: it doesn't fix the
> metric loss, but it does downgrade the crash to a timeout:

I have restored the `isConnected()` filter. I was sure there was a reason why it wasn't needed, to do with close events, but I've lost the thread there and also there might be some concurrency concerns. That still leaves us with issues with dead cluster metrics, but those will need to be addressed in #803 anyway and that takes one more thing off of your concerns list. 

I disagree with the AI here as well

1) We're a library. We aren't even the reason the cluster instances got created. (unlike workers, where one worker *is* our fault). If the developers aren't listening for error events on cluster send, then they are already going to suffer crashes. If anything, us adding a listener on error **changes the existing behavior of their application**, because the error events are such a coarse-grained control mechanism that we can't see our errors without also seeing theirs. E

Either they have a crash, or they do not. We don't run in isolation. Except in tests and examples.

2) There is no timeout. We only send GET_METRICS_REQ to the workers who are registered in 'workers', and now only those that are still connected. So there should be no timeout.

krajorama (approved):
LGTM, sorry about the noise, I don't have the knowledge to truly judge AI on this :( It can be very convincing ...

jdmarshall:
There's definitely some concurrency issues in this code but some of them are duplicated from the existing cluster.js feature. If you compare the code with 15 we are already doing better. But I was more optimistic about the compatibility of workers with cross-cutting concerns like telemetry and I'm feeling like I'm having to trick workers into being useful, which is not so much a new experience as it is revisiting an old one. Hence the walls of text in the node issue I filed.

We probably would have shipped a few more bugs without your insistence.

PR #793, opened 2026-08-01
Fix escaping for non-string label values

### LINKED ISSUE #791: Non-string label values bypass escaping and can produce malformed exposition

happykawayigt:
Fixes #791.

Non-string label values are now converted to strings during exposition formatting before the existing backslash, newline, and quote escaping runs. This keeps the metric-recording path unchanged and preserves the existing fast path for string values.

A regression test covers an array label whose string representation contains a quote, newline, and backslash. The test fails against the previous implementation and passes with this change.

Validation:
- `npm test`
- 29 test suites passed
- 546 tests passed
- ESLint, Prettier, and TypeScript checks passed

### DIFF
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -305,7 +305,7 @@ function flattenSharedLabels(labels) {
 }
 function escapeLabelValue(str) {
 	if (typeof str !== 'string') {
-		return str;
+		str = String(str);
 	}
 	return escapeString(str).replace(/"/g, '\\"');
 }
--- a/test/registerTest.js
+++ b/test/registerTest.js
@@ -340,6 +340,29 @@ describe('Register', () => {
 			expect(escapedResult).toMatch(/\\"/);
 		});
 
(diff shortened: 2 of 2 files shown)


jdmarshall:
This solution has the same performance regression as #792 and I now have two competing implementations neither of which I'm willing to accept. I made a counterproposal in the ticket #791, but that suggestion is a bit more technically challenging. It does on the other hand offer a chance at a perf tag in the changelog.

PR #795, opened 2026-08-01
Deprecate functions that were replaced by LabelMap

jdmarshall:
These functions are exposed by the public API and are likely to be used by third party custom Metrics, so cannot simply be deleted.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -32,6 +32,7 @@ exports.getValueAsString = function getValueString(value) {
 
 /**
  * @function removeLabels
+ * @deprecated please use LabelMap
  * @param {Map} hashMap
  * @param {string[]} labels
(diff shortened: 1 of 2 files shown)


cjihrig (approved):

lib/util.js
```diff
@@ -142,6 +145,11 @@ function hashObject(labels, labelNames) {
 
 	return fastHashObject(keys, labels);
 }
+
+/**
+ * @deprecated please use LabelMap
+ * @type {(function(*, *): (string))|*}
```
zbjornson: This `@type` annotation isn't valid. Could delete and rely on the inferred type since it's deprecated anyway.

krajorama (approved):
stamp to unblock

PR #798, opened 2026-08-02
fix: normalize label values once, when a label combination is first stored

### LINKED ISSUE #791: Non-string label values bypass escaping and can produce malformed exposition

milcho0604:
Fixes #791, along the direction proposed there: pay the coercion cost once, when a new label combination is first stored, instead of on every `metrics()` call (#792/#793 were declined for adding cost to rendering, which is linear in total cardinality).

## What

- `LabelMap` gains a single insertion point (`#insert`) used by `set`, `setDelta`, `getOrAdd` and `merge`. It coerces label values with template interpolation — the same `ToString` the exposition applies — via `normalizeLabels()`, which always returns a copy the store owns.
- **Nullish values stay untouched.** `keyFrom()` treats them as absent, so coercing them to `"null"`/`"undefined"` would make stored labels compute a different key than the one they are stored under — breaking `remove(entry.labels)` round-trips (Summary's pruning does exactly that) and collapsing `{a: null}` with `{a: 'null'}` after worker serialization. Their rendered form contains nothing that needs escaping.
- `merge()` keeps the stored (normalized) labels on update instead of overwriting them with the caller's raw object — the update path that could previously replace stored labels without going through insertion.
- `getOrAdd()` passes the normalized labels to `init()`, so values that keep their own copy of the labels store the same normalized object. `Summary` needs this: its exported labels come from the stored value, not the map entry.
- `LabelGrouper` deliberately does **not** normalize, per the discussion in #791: aggregation input comes from `registry.getMetricsAsJSON()`, whose store-backed labels were already normalized on first insertion, so re-checking every value would tax `aggregate()` for work the stores already did — `aggregate()` keeps its current cost, byte-for-byte. I verified the assumption: worker payloads are exactly `getMetricsAsJSON()` output, where metric values carry stored (normalized) labels. The inputs that reach aggregation without normalization are unchanged from `main`: custom collector results, registry default labels (merged in raw by `getMetricsAsJSON()`), and the synthesized `le`/`quantile` labels, which are attached numerically at export time — all reserved or user-controlled values that pass through exactly as today (`promtool` accepts the aggregated output, see below). On #789: its diff touches `lib/cluster.js`/`lib/worker.js` lifecycle only — payloads are still `getMetricsAsJSON()` output fed to `Registry.aggregate()`, so the two changes stay orthogonal.

With the stores normalized, the existing render-time escaping (which already handles strings correctly) produces well-formed exposition for the #791 reproduction:

```
g{x="say \"hi\""} 1
```

## Cost

The recording path for existing combinations is unchanged — lookup only, no new code. First insertion of a combination pays one scan of its labels, plus a copy and coercion when something is non-string (`getOrAdd` scans once more inside `#insert`; normalization is idempotent). Interleaved 5-round medians (Node 25, arm64, lower is better; ranges in parentheses):

| path | main | this PR |
| --- | --- | --- |
| hot: `inc()` on existing combination × 5M | 135.2 ms (135.0–137.9) | 132.4 ms (131.9–137.4) — within noise |
| `metrics()` with 10k series × 50 | 199.8 ms (195.7–210.6) | 199.1 ms (184.2–202.0) — within noise |
| 300k first insertions | 48.3 ms (47.7–49.2) | 51.0 ms (50.2–53.1) — +5.6%, ~9 ns per new combination |

`aggregate()` is untouched — no code change on that path.

<details>
<summary>Benchmark script (run against two checkouts, interleaved)</summary>

```js
'use strict';
// usage: node bench.js <prom-client dir> <hot|insert|render>
const path = process.argv[2];
const which = process.argv[3];
const client = require(require('node:path').resolve(path, 'index.js'));

function hrms(fn) {
	const t0 = process.hrtime.bigint();
	fn();
	return Number(process.hrtime.bigint() - t0) / 1e6;
}

if (which === 'hot') {
	const r = new client.Registry();
	const c = new client.Counter({ name: 'c_total', help: 'h', labelNames: ['x'], registers: [r] });
	c.inc({ x: 'warm' }, 1);
	const N = 5_000_000;
	console.log(hrms(() => {
		for (let i = 0; i < N; i++) c.inc({ x: 'warm' }, 1);
	}).toFixed(1));
} else if (which === 'insert') {
	const { LabelMap } = require(require('node:path').resolve(path, 'lib/util.js'));
	const N = 300_000;
	const maps = [];
	for (let i = 0; i < 10; i++) maps.push(new LabelMap(['a', 'b']));
	console.log(hrms(() => {
		for (const m of maps) {
			for (let i = 0; i < N / 10; i++) m.set({ a: `v${i}`, b: 'k' }, i);
		}
	}).toFixed(1));
} else if (which === 'render') {
	const r = new client.Registry();
	const c = new client.Counter({ name: 'c_total', help: 'h', labelNames: ['x'], registers: [r] });
	for (let i = 0; i < 10_000; i++) c.inc({ x: `v${i}` }, 1);
	(async () => {
		await r.metrics();
		const t0 = process.hrtime.bigint();
		for (let i = 0; i < 50; i++) await r.metrics();
		console.log((Number(process.hrtime.bigint() - t0) / 1e6).toFixed(1));
	})();
}
```

</details>

## Observable changes

- Label values that pass through the built-in stores are reported as strings (`3` becomes `"3"`) wherever stored labels surface: `getMetricsAsJSON()`, each metric's public `get()`, worker payloads, and — transitively — aggregation output. `Registry.aggregate()` itself passes labels through untouched. Noted in the changelog.
- No `index.d.ts` change: label output types are already `string | number`, and `le`/`quantile`/default/custom-collector labels can still be numeric.
- `keyFrom()` already coerces ordinary non-nullish values while building keys, so `get()`/`set()` lookups accept either representation — pinned by new tests, including `null` vs `'null'` staying distinct and `remove(entry.labels)` round-tripping.

## Scope

Covered: everything that goes through the built-in metric stores (`Counter`, `Gauge`, `Histogram`, `Summary`); cluster aggregation benefits transitively because worker payloads carry stored labels. Not covered (unchanged, documented): custom collector results rendered directly from `get()`, registry default labels, and exemplar labels — none of these pass through the stores. Relocating the quote/backslash/newline escaping itself into the stores is deliberately left out: rendering currently escapes backslashes, so it would have to move atomically across every label source or double-escape; I can scope and benchmark that separately.

## Test

- New `label normalization` unit block for `LabelMap`: coercion at insertion, caller's object never mutated, all-string sets stored without copying, lookups by either representation, nullish round-trips (`remove(entry.labels)` works; `null` vs `'null'` stay distinct), `getOrAdd` init receiving normalized labels, `merge` keeping normalized labels across updates. `LabelGrouper` tests pin the pass-through behaviour (raw labels preserved).
- End-to-end regressions through real metrics (both content types): `Gauge.set` with array/number labels renders escaped; `Summary.observe` covers the stored-value labels path.
- Existing expectations that pinned numeric stored labels updated to the normalized form (`versionTest`, `defaultMetricsTest`, `utilTest`); aggregation expectations stay raw, pinning the pass-through.
- Applying the full test patch to `main` without the lib changes yields 23 failures; the full suite passes with them: 556/556, lint + prettier + tsc clean.
- `promtool check metrics` accepts the rendered output (rc=0) for a registry mixing quote/newline/backslash/boolean/number labels across all four metric types, and for a simulated cluster round-trip (two workers' `getMetricsAsJSON()` through JSON serialization into `AggregatorRegistry.aggregate()`). The same direct-render check on `main` fails with the error from #791 (`unexpected end of label value`, rc=1).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,11 @@ This release marks our first release under the Prometheus umbrella.
 - chore: Add copyright license headers and test
 - Make cluster and worker-thread metric aggregation order deterministic
 - Export `MetricObject`, `MetricObjectWithValues`, `MetricValue` and `MetricValueWithName` from the TypeScript definitions
+- fix: Non-string label values (except `null`/`undefined`) are coerced to strings when a
+  combination is first stored, so exposition escapes them and `getMetricsAsJSON()` reports
+  them as strings. The store now keeps its own copy of the labels: mutating the caller's
+  object after recording no longer changes the stored series
+- fix: Label-less summaries report `labels: {}` in `getMetricsAsJSON()`, like other metrics
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - perf: Remove truthy conditionals from default metric collectors
 - fix: Preserve zero-valued Counter exemplars
 - perf: Remove object and array fallback truthiness from core metric paths
+- fix: Non-nullish, non-string label values are coerced to strings when a combination is first stored, so exposition escapes them; the store also keeps its own copy, so mutating the caller's object after recording no longer changes the stored series
+- fix: Label-less summaries report `labels: {}` in `getMetricsAsJSON()`, like other metrics (Fixes [#838](https://github.com/prometheus/client_js/issues/838))
 
 ### Added
 
--- a/lib/summary.js
+++ b/lib/summary.js
@@ -39,14 +39,14 @@ class Summary extends Metric {
 		this.store = new LabelMap(this.labelNames);
 
 		if (this.labelNames.length === 0) {
-			this.store.set(
-				{},
-				{
+			this.store.getOrAdd({}, storedLabels => {
+				return {
+					labels: storedLabels,
 					td: new timeWindowQuantiles(this.maxAgeSeconds, this.ageBuckets),
 					count: 0,
 					sum: 0,
-				},
-			);
+				};
+			});
 		}
 	}
 
@@ -177,14 +177,17 @@ function observe(labels) {
 			);
 		}
 
-		const summaryOfLabel = this.store.getOrAdd(labelValuePair.labels, () => {
-			return {
-				labels: labelValuePair.labels,
-				td: new timeWindowQuantiles(this.maxAgeSeconds, this.ageBuckets),
-				count: 0,
-				sum: 0,
-			};
-		});
+		const summaryOfLabel = this.store.getOrAdd(
+			labelValuePair.labels,
+			storedLabels => {
+				return {
+					labels: storedLabels,
+					td: new timeWindowQuantiles(this.maxAgeSeconds, this.ageBuckets),
+					count: 0,
+					sum: 0,
+				};
+			},
+		);
 
 		summaryOfLabel.td.push(labelValuePair.value);
 		summaryOfLabel.count++;
--- a/lib/summary.js
+++ b/lib/summary.js
@@ -72,9 +72,10 @@ class Summary extends Metric {
 			if (this.pruneAgedBuckets && s.td.size() === 0) {
 				this.store.remove(entry.labels);
 			} else {
-				values.push(...extractSummariesForExport(s, this.percentiles));
-				values.push(getSumForExport(s, this));
-				values.push(getCountForExport(s, this));
+				const labels = entry.labels;
+				values.push(...extractSummariesForExport(s, labels, this.percentiles));
+				values.push(getSumForExport(s, labels, this));
+				values.push(getCountForExport(s, labels, this));
 			}
 		}
 
@@ -126,30 +127,30 @@ class Summary extends Metric {
 	}
 }
 
-function extractSummariesForExport(summaryOfLabels, percentiles) {
+function extractSummariesForExport(summaryOfLabels, labels, percentiles) {
 	summaryOfLabels.td.compress();
 
 	return percentiles.map(percentile => {
 		const percentileValue = summaryOfLabels.td.percentile(percentile);
 		return {
-			labels: Object.assign({ quantile: percentile }, summaryOfLabels.labels),
+			labels: Object.assign({ quantile: percentile }, labels),
 			value: percentileValue ? percentileValue : 0,
 		};
 	});
 }
 
-function getCountForExport(value, summary) {
+function getCountForExport(value, labels, summary) {
 	return {
 		metricName: `${summary.name}_count`,
-		labels: value.labels,
+		labels,
 		value: value.count,
 	};
 }
 
-function getSumForExport(value, summary) {
+function getSumForExport(value, labels, summary) {
 	return {
 		metricName: `${summary.name}_sum`,
-		labels: value.labels,
+		labels,
 		value: value.sum,
 	};
 }
@@ -179,7 +180,6 @@ function observe(labels) {
 
 		const summaryOfLabel = this.store.getOrAdd(labelValuePair.labels, () => {
 			return {
-				labels: labelValuePair.labels,
 				td: new timeWindowQuantiles(this.maxAgeSeconds, this.ageBuckets),
 				count: 0,
 				sum: 0,
--- a/lib/util.js
+++ b/lib/util.js
@@ -169,6 +169,72 @@ exports.nowTimestamp = function nowTimestamp() {
  */
 
 /**
+ * Copy the labels the store is about to take ownership of, coercing
+ * non-nullish values to the string Prometheus expects. Coercion uses
+ * template interpolation — exactly what rendering would have done later —
+ * so the rendered form is unchanged while escaping no longer gets
+ * skipped (#791).
+ *
+ * The copy is unconditional: the entry keeps these labels for its lifetime,
+ * so holding on to an object the caller can still mutate would let a later
+ * mutation change what a stored series reports.
+ *
+ * Nullish values are copied as-is: `keyFrom()` treats them as absent, so
+ * coercing them to `"null"`/`"undefined"` would make the stored labels
+ * compute a different key than the one they are stored under — breaking
+ * `remove(entry.labels)` round-trips (Summary's pruning does exactly that)
+ * and collapsing `{a: null}` with `{a: 'null'}` after serialization. Their
+ * rendered form (`"null"`) contains nothing that needs escaping anyway.
+ *
+ * Two corner cases are deliberately unsupported (each needs another
+ * `for...in`-visible property present — alone, `isEmpty()` keeps original
+ * and copy agreed on `''`):
+ * - `Object.create(null)` labels missing a declared name that collides with
+ *   `Object.prototype` (`constructor`, …): the plain copy reads the
+ *   inherited member where the original read `undefined`.
+ * - Non-enumerable label properties: no enumeration sees them, so the copy
+ *   drops what `keyFrom()` read by property access.
+ * Symbols are not label data: `keyFrom()` iterates declared names,
+ * exposition uses `Object.entries()`; neither reads them.
+ * @param {object} labels
+ * @returns {object} a copy owned by the store
+ */
+function normalizeLabels(labels) {
+	// Spread first, for the packed object shape V8 gives a clone: building the
+	// copy up key by key instead costs about 24 bytes per stored series.
+	const copy = { ...labels };
+
+	// Then walk the source with `for...in`, which picks up inherited enumerable
+	// labels too. `keyFrom()` reads labels by name, so it sees those as well, and
+	// a copy that dropped them could not reproduce the key its entry is filed
+	// under — `remove(entry.labels)`, which Summary's pruning uses, would quietly
+	// miss. Non-enumerable labels stay out of reach of any enumeration; they only
+	// survived before because the store kept the caller's object.
+	for (const name in labels) {
+		const value = labels[name];
+		const stored =
+			typeof value === 'string' || value === null || value === undefined
+				? value
+				: `${value}`;
+
+		if (name === '__proto__') {
+			// A legal label name, but assigning it would invoke the prototype
+			// setter instead of defining a property, dropping the label.
+			Object.defineProperty(copy, name, {
+				value: stored,
+				writable: true,
+				enumerable: true,
+				configurable: true,
+			});
+		} else {
+			copy[name] = stored;
+		}
+	}
+
+	return copy;
+}
+
+/**
  * Lookup table for stats by labels.
  */
 class LabelMap {
@@ -180,6 +246,22 @@ class LabelMap {
 
 	constructor(labelNames = []) {
 		this.#labelNames = new Set(labelNames.slice().sort());
+	}
+
+	/**
+	 * The single insertion point — every new label combination enters the map
+	 * here, and takes its own copy of the labels on the way in. Only a
+	 * combination's first record reaches this method, so the recording fast
+	 * path for existing combinations copies nothing.
+	 * @param {string} key precomputed `keyFrom(entry.labels)`
+	 * @param {StatsEntry} entry
+	 * @returns {StatsEntry}
+	 */
+	#insert(key, entry) {
+		entry.labels = normalizeLabels(entry.labels);
+		this.#map.set(key, entry);
+
+		return entry;
 	}
 
 	/**
@@ -195,7 +277,7 @@ class LabelMap {
 		if (entry !== undefined) {
 			entry.value = value;
 		} else {
-			this.#map.set(key, { value, labels });
+			this.#insert(key, { value, labels });
 		}
 
 		return this;
@@ -214,7 +296,7 @@ class LabelMap {
 		if (entry !== undefined) {
 			entry.value += value;
 		} else {
-			this.#map.set(key, { value, labels });
+			this.#insert(key, { value, labels });
 		}
 
 		return this;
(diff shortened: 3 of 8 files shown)


lib/util.js
```diff
@@ -400,6 +455,13 @@ class LabelGrouper {
 
 	/**
 	 * Adds the `value` to the `key`'s array of values.
+	 *
+	 * NB: no label normalization here, by design. Aggregation input comes from
+	 * `registry.getMetricsAsJSON()`, whose store-backed labels were already
+	 * normalized by LabelMap on first insertion — re-checking every value on
+	 * this path would tax `aggregate()` for work the stores already did.
+	 * Labels that never pass through the stores (custom collector results,
+	 * registry default labels) arrive here as-is, unchanged from before.
```
jdmarshall: 👍

lib/util.js
```diff
@@ -168,6 +168,40 @@ exports.nowTimestamp = function nowTimestamp() {
...
+			const copy = { ...labels };
+			for (const key in copy) {
+				const v = copy[key];
+				if (typeof v !== 'string' && v !== null && v !== undefined) {
+					copy[key] = `${v}`;
+				}
+			}
+			return copy;
+		}
+	}
```
jdmarshall, line 240: I don't like the nested loops here, but there are several ways we could go here.

Do you believe the guard loop is necessary to reduce the need to copy the labels, or would it be simpler to defensively copy them on every `#insert()` call? This is a spot where we accept a value from the user as if we now own it and perhaps that is not the best idea. 

If you think so, then we could still set up these loops as two sequential loops, which means the eye won't stick on this code in perpetuity every time someone goes hunting around for a bug. For instance by setting a 

```
    needsNormalization = true;
    break;
```

in the first loop and then returning if the boolean is false.

jdmarshall (changes requested):
This is pretty close to ready to go, except that the nested double loop is giving me some pause.  

I can't land this yet though because we are working on a dot release to get a huge pile of existing changes out. But this should be able to make the 1.0 cutoff for sure.

lib/util.js
milcho0604: Good question — I measured it, and the guard loop isn't worth keeping, so I took the defensive-copy route you suggested. `normalizeLabels()` is flat now — one loop, no nesting — and always copies.

The structural half of the answer first: `#insert()` runs once per stored combination, not once per record. All four call sites look the key up and fall through only on a miss, so a new combination's first record always pays for the copy, and every record after it pays nothing.

Numbers are on Node 24.11. (I first measured on Node 25, which is outside `engines`, and the deltas there were both larger and less consistent — please disregard that if you saw it.)

**The repo's own suite**, with this branch's previous revision as baseline: *no significant regressions* across all 46 cases, from 1.03x slower to 1.08x faster.

**Focused A/B on the storage path** — one implementation per process, since loading both into one makes the shared call sites polymorphic and taxes both; median of 9 samples, order alternated between runs, ns/op:

|  | guard loop | defensive copy |  |
| --- | ---: | ---: | --- |
| insert, 1 label (per new combination) | 55.0 | 59.7 | +8.6% |
| insert, 3 labels | 125.5 | 148.7 | +18.5% |
| insert, 6 labels | 200.0 | 244.8 | +22.4% |
| record on an existing combination | 51.5 | 51.9 | +0.7% |
| create + 10 records, per record | 125.0 | 131.6 | +5.3% |
| create + 100 records, per record | 106.4 | 105.9 | −0.5% |

So the cost is real but confined to a once-per-series path: still visible at ten records of a series (+5.3%), down in the noise by a hundred, which is why none of the suite's workloads can see it. Retained memory at 250k unique series is unchanged (64.6 MiB against 64.7 MiB) — the store simply retains an object of its own instead of retaining the caller's.

Taking ownership did mean pinning down what a "copy" means, and two things in that loop are load-bearing rather than decorative:

- It iterates the **source** with `for...in`, so inherited enumerable labels are materialized. `keyFrom()` reads labels by name and sees those too, so an own-properties-only copy could not reproduce the key its entry is filed under, and `remove(entry.labels)` — which `Summary`'s pruning depends on — would quietly miss.
- `__proto__` gets `Object.defineProperty`. It passes the label-name regexp, and plain assignment invokes the prototype setter instead of defining a property, so the label would vanish from the stored copy with the same consequence.

Both are covered: removing the `defineProperty` branch fails one test, and switching the loop to iterate the copy fails two. The spread that seeds the copy is there for the packed object shape — building the copy key by key instead costs about 24 bytes per stored series — so nothing asserts it; it does also carry over symbol-keyed properties, which the loop ignores.

Two divergences I did not chase, in case you'd rather I did. Both come from the copy being an ordinary object built by enumeration, and both were previously masked by the store keeping the caller's object:

- A labels object created with `Object.create(null)` becomes an ordinary one. That is observable only if a declared label name is also an `Object.prototype` member (`constructor`, `toString`, …) *and* absent from the object — `keyFrom()` then reads the inherited function for the copy but `undefined` for the original.
- A non-enumerable label property is invisible to any enumeration, so it is dropped, while `keyFrom()` still reads it.

Preserving the source prototype gives up the packed-shape win above, and `keyFrom()`'s unguarded property access is the underlying sharp edge in both cases, so I left them rather than widen this PR.

One knock-on that I think is a net simplification: once the store owns its labels, `Summary`'s stored value no longer needs its own copy of them. `getOrAdd()` is back to calling `init()` with no arguments, as on main, and the summary export helpers take `entry.labels` — `summary.js` ends up back at main's length. The one externally visible effect is that a label-less summary now returns `labels: {}` rather than `labels: undefined` from `getMetricsAsJSON()` (serialized JSON previously omitted the property), matching what counters, gauges and histograms already do. Text exposition for ordinary label objects is unchanged apart from the intended difference: a caller that mutates its labels object after recording no longer changes what the stored series reports. Happy to split that part out if you'd rather keep this PR to `util.js`.

I also shortened the changelog entries, per your note on #790.

jdmarshall: On the topic of label iteration, I would say that consistency with metrics() is the goal. And documenting corner cases we mean not to support. 

As it is, this PR is expanding the supported inputs. And while supporting Symbol is more than I would have done myself, if it’s cheap enough then I’m not going to block adding it, because I don’t see the case where continuing to support it later blocks a feature, even performance. 

However, supporting Enums better is a goal I support, because label values should be statically defined for OPEX and DOS reasons, and what better way to do so than to fix this bug.

milcho0604: Documented both unsupported corner cases next to the relevant code: (1) `Object.create(null)` label objects that omit a declared label name matching an `Object.prototype` member, and (2) non-enumerable label properties. The notes describe the narrow trigger conditions rather than treating either object shape as unsupported in general.

Agreed that consistency with `metrics()` is the goal, and you're right that this PR expands the supported input surface. I rechecked the two cases from my last comment. On symbols, to pin the scope: symbol-keyed properties have no support path of their own — the spread that builds the copy incidentally carries own enumerable ones, and neither `keyFrom()` nor exposition reads them. Symbol *values* throw in `keyFrom()` on main and on this branch alike. My statement about inherited enumerable labels was incorrect: on this branch the copy makes them own properties, so keying and exposition see the same label set — I added a registry-level test for that case.

On numeric-enum-backed label values: agreed, and that's the case this fix serves — the member is coerced once when its label combination is first stored, and the resulting string then follows the existing escaping path at exposition.

Enforcing own-only labels consistently would require the same rule across insert, lookup, and the public `remove(labels)`, which puts a check on the labeled-series hot path — I'd treat that as a separate, benchmarked change. The updated docs and test are pushed; the diff is ready for another look.

jdmarshall (approved):

lib/util.js
```diff
@@ -202,6 +202,72 @@ exports.waitFor = async function waitFor(promise, limit = 5_000) {
...
+ * exposition uses `Object.entries()`; neither reads them.
+ * @param {object} labels
+ * @returns {object} a copy owned by the store
+ */
+function normalizeLabels(labels) {
+	// Spread first, for the packed object shape V8 gives a clone: building the
+	// copy up key by key instead costs about 24 bytes per stored series.
+	const copy = { ...labels };
+
+	// Then walk the source with `for...in`, which picks up inherited enumerable
```
jdmarshall: This is too much exposition.  And restates the jsdoc unless I'm misreading.

lib/util.js
```diff
@@ -202,6 +202,72 @@ exports.waitFor = async function waitFor(promise, limit = 5_000) {
...
+/**
+ * Copy the labels the store is about to take ownership of, coercing
+ * non-nullish values to the string Prometheus expects. Coercion uses
+ * template interpolation — exactly what rendering would have done later —
+ * so the rendered form is unchanged while escaping no longer gets
+ * skipped (#791).
+ *
+ * The copy is unconditional: the entry keeps these labels for its lifetime,
+ * so holding on to an object the caller can still mutate would let a later
+ * mutation change what a stored series reports.
```
jdmarshall: This restates paragraph 1.

lib/util.js
```diff
@@ -202,6 +202,72 @@ exports.waitFor = async function waitFor(promise, limit = 5_000) {
...
+ * Two corner cases are deliberately unsupported (each needs another
+ * `for...in`-visible property present — alone, `isEmpty()` keeps original
+ * and copy agreed on `''`):
+ * - `Object.create(null)` labels missing a declared name that collides with
+ *   `Object.prototype` (`constructor`, …): the plain copy reads the
+ *   inherited member where the original read `undefined`.
+ * - Non-enumerable label properties: no enumeration sees them, so the copy
+ *   drops what `keyFrom()` read by property access.
+ * Symbols are not label data: `keyFrom()` iterates declared names,
+ * exposition uses `Object.entries()`; neither reads them.
```
jdmarshall: Some of this would do better as the commit message, where I can still find it when I'm trying to figure out how to add additional functionality.  

In fact maybe this whole section should be in the commit message.

CHANGELOG.md
```diff
@@ -57,6 +57,11 @@ This release marks our first release under the Prometheus umbrella.
 - Export `MetricObject`, `MetricObjectWithValues`, `MetricValue` and `MetricValueWithName` from the TypeScript definitions
 - chore: Old label processing code marked as deprecated
 - perf: Stop rebuilding the label array for every rendered series, and skip shared label handling entirely for values that do not set it; 9-22% faster `metrics()` in the registry benchmarks
+- fix: Non-string label values (except `null`/`undefined`) are coerced to strings when a
```
jdmarshall: Shorter, and you'll need to rebase and move these two the new section.

jdmarshall (changes requested):

lib/summary.js
```diff
@@ -126,30 +127,30 @@ class Summary extends Metric {
...
 		};
 	});
 }
 
-function getCountForExport(value, summary) {
+function getCountForExport(value, labels, summary) {
 	return {
 		metricName: `${summary.name}_count`,
-		labels: value.labels,
+		labels,
```
jdmarshall: If we are storing the normalized keys in the store in `#insert()` then these extra parameters should not be necessary, right?

jdmarshall (changes requested):

lib/summary.js
```diff
@@ -72,9 +72,10 @@ class Summary extends Metric {
 			if (this.pruneAgedBuckets && s.td.size() === 0) {
 				this.store.remove(entry.labels);
 			} else {
-				values.push(...extractSummariesForExport(s, this.percentiles));
-				values.push(getSumForExport(s, this));
-				values.push(getCountForExport(s, this));
+				const labels = entry.labels;
```
jdmarshall: So are we not encoding the label values in the individual store entries? I would have figured we did not need to pass these down from above.

milcho0604:
You were right, the extra parameters were not needed. `getOrAdd()` now hands the stored labels to `init()`, so `get()` and the export helpers are byte-identical to main again. Also rebased and moved the changelog entries.

lib/summary.js
```diff
@@ -39,14 +39,14 @@ class Summary extends Metric {
 		this.store = new LabelMap(this.labelNames);
 
 		if (this.labelNames.length === 0) {
-			this.store.set(
-				{},
-				{
+			this.store.getOrAdd({}, storedLabels => {
```
jdmarshall, line 42: This is a curious change. Are you fixing a bug? Can you file it for posterity?

lib/summary.js
```diff
@@ -39,14 +39,14 @@ class Summary extends Metric {
...
+				return {
```
jdmarshall, line 43: Nit: if you want to return a struct from an arrow function, you can wrap parentheses around it to avoid the 'return'.

lib/util.js
```diff
@@ -270,16 +320,15 @@ class LabelMap {
 	 * called to create an object to put there. This allows for nested structures.
 	 *
 	 * @param {object} labels labels for the new entry
-	 * @param {[Function]} init function to generate an empty record
+	 * @param {[Function]} init receives the stored labels, returns an empty record
```
jdmarshall: We've already been going back and forth on this PR a lot so it's fine to say 'no', but the init parameter really deserves a type declaration, especially after I defined StatsEntry, and now that we are introducing a second call site.

I don't use this much but jsdoc has: 

And there's no typescript entry either.
```
/**
 * @callback myCallback
 * @param {number} x - ...
 */
```

jdmarshall (approved):
Two nits but otherwise good to go.

milcho0604:
Filed #838 for the summary labels change and linked it from the changelog entry. Added the @callback declaration for init.

I kept the braced return in the two summary callbacks. arrow-body-style in eslint.config.mjs is set to as-needed with requireReturnForObjectLiteral true, so the parenthesised form fails npm run lint at both sites. Say the word if you would rather relax the rule and take the parens.

Rebased onto main.

PR #799, opened 2026-08-03
Rename to prometheus-io to avoid naming collision on npmjs.org

jdmarshall:


cjihrig (approved):

krajorama (approved):
LGTM, I've opened the small #801 to update the changelog and readme as well if you don't mind

PR #801, opened 2026-08-10
Add CHANGELOG entry and README install section for the prometheus-io rename

krajorama:
Stacked on top of #799 — this branch is based on `jdmarshall:prometheus-io`, so until #799 merges the diff here also shows that PR's rename commit. Only the second commit (`CHANGELOG.md`, `README.md`) is new.

#799 renames the package to `@prometheus-io/client` but leaves the change unrecorded, which is why its Changelog Reminder check is red.

Also update the readme with some install instructions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ This release marks our first release under the Prometheus umbrella.
 
 ### Breaking
 
+- The package is renamed from `prom-client` to `@prometheus-io/client`. Update your
+  dependencies and any `require()`/`import` statements accordingly.
+- The cluster and worker thread IPC message types are renamed from `prom-client:*`
+  to `@prometheus-io/client:*`. A cluster primary and its workers must therefore run
+  the same major version.
 - Drop support for Node.js versions 16, 18, 20, 21 and 23
 - Metric internal storage ('hashMap') changed to a separate object, LabelMap. If you have
   subclassed the built-in metric types you may need to adjust your code.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -41,7 +46,7 @@ This release marks our first release under the Prometheus umbrella.
 - ci: Run benchmarks for pull requests
 - ci: switch out deprecated benchmark-regression library for replacement
 - AggregatorRegistry renamed to ClusterRegistry, old name deprecated
-- chore: update faceoff to 1.1
+- chore: update faceoff to 1.3
```
jdmarshall, line 49: Faceoff didn't even exist the last time this package was released. Strictly speaking, if faceoff warrants an entry at all it should be, "replaced benchmark-regression library with faceoff"

`benchmark-regression` is built on `benchmark` and both of them had serious issues with async functions and neither are being maintained. There's been a rash of async-hostile libraries simply going dark rather than wrestling with the design assumptions that make it a challenge to modify them to support async code. The other module I maintain also had that issue, and if it weren't for an 11th hour epiphany while preparing the next major version of that library, it would still be awkwardly bolted on to the side. It's going to take a second major release and relying on top-level async to make it fully integrated.

I created faceoff to be a drop-in, async replacement for benchmark-regression while exposing the more legible output of bench-node, which was created by someone on the NodeJS steering committee. And now for some reason I'm a maintainer on that project as well.

jdmarshall (approved):

Issue #800, opened 2026-08-04
Speed up label escaping


jdmarshall:
On the back of #798, there is now an opportunity to escape label values once at insert time rather than on every metrics call.

The metrics call is a high priority target for both time and space improvements. When I first came to the project, it was the most common complaint in the issue backlog. We have managed to speed up everything but this part. Collection is ~20% faster, aggregation 2.4x faster, but metrics are only a couple % faster. 

With the introduction of #insert, we should be able to move the `escapeString()` call to insert. This could be done two ways:

1. change the labels to only be escaped labels 
2. store the escaped form of the labels beside the original form

Number 2 may be more compatible, and since most strings do not need to be escaped in the first place, represent only a small increase in memory usage. 

This work also may reduce or remove the need for the label cache in the Registry. Despite the use of a WeakMap for storage, this does represent a memory cost, and if the label concatenation is cheaper, then it's worth also considering whether the added overhead of the lookups is really worth the extra space and code, or whether the whole thing can be deleted.

milcho0604:
Author of #798 here. Thanks for updating that branch — and I would be glad to take this one on if it is still free.

Before starting, could I check the direction with you? Reading the render path on the current head:

- `formatLabels()` escapes **label values** on every line (`escapeLabelValue(value)`, `lib/registry.js:288`); label *names* are interpolated raw there.
- Metric `name` and `help` are escaped once per metric (`lib/registry.js:58-59`), not per series.

So the per-call cost that `#insert` could absorb looks like the label-value escaping. Options 1 and 2 read that way to me as well, but the opening line says "label names", and I would rather ask than solve the wrong half.

Assuming label values, I would go with your option 2. Beyond the compatibility point you make, the stored labels are round-tripped by `remove()` and by the Summary pruning path, so replacing them in place changes observable behaviour — #798 ran into exactly that.

One gap I would want to cover either way: `formatLabels()` is also called for exemplar label sets (`lib/registry.js:100`), and those do not go through `LabelMap#insert`. They would still need escaping at render time unless handled separately.

If you would rather keep this one yourself, no problem at all — and if option 1 is what you actually want, I am happy to do it that way instead.

jdmarshall:
I was hoping to look at this one but I am six hours into #803 and all I've managed to do is refactor the async and promise code to something more amenable to adding new features.

[...]

PR #804, opened 2026-08-14
perf(registry): Avoid label array manipulation in the happy path.

milcho0604:
Towards #800. You recommended landing the `formatLabels()` cleanup first, to see what it was worth on its own. I did not measure hoisting `exclude` to the caller separately, so treat that part as unmeasured. What I did measure is one level out.

`getMetricsAsString()` built the formatted labels, spread them into a second array along with the flattened shared labels, then filtered that into a third. The filter never filtered labels, because `formatLabels()` cannot return a falsy element, and the empty string from an empty flatten was the only thing it removed. A guard on the push does the same work with one array. That is where the time was.

Separately, `sharedLabels` was destructured with a `{}` default. Only Histogram sets the field, so every series of every other metric allocated that object, ran `Object.hasOwn` against it once per label, and handed `flattenSharedLabels()` a key its `WeakMap` would never see again. Branching on the field being absent skips all of it. A histogram declared without label names still shares an empty object, so it keeps the slow path; making Histogram omit it is a separate change.

Your sketch had `if (sharedLabels)` and I went with `!== undefined`. A collector passing `sharedLabels: null` throws on main, and the truthy test turns that into a silent render with the shared labels missing. Happy to switch if you would rather `null` render as no shared labels, it is a behaviour change either way and I kept main's.

The numbers are three reps of `npm run benchmarks` per Node major on 22, 24 and 26, median per case, against `trunk` at 1908ddb. On the registry `metrics()` cases this is 16%, 42 of 42 positive, +9% to +22%. Averaging the whole registry suite gives about 10%, since `getMetricsAsJSON()` is a third of those cases and cannot reach the change. Everything else in the harness sits at +0.07% median, 51 of 93 positive. Measured the same way, 623a74f is 4.9% and taking the helper out on its own is 4.2%, so removing the filter after that is the 12 points and the helper removal cost a little. faceoff's own verdicts on the run read `registry ⇒ metrics() no labels ⇒ current (1.24x faster)` and similar down the list.

The test is a histogram with two shared labels and a value that needs escaping. On main you can delete the dedup exclusion, or the escaping on shared label values, and all 551 tests still pass. Emitting only the first shared label is caught by one OpenMetrics snapshot in another file. The second test pins the `null` case, which nothing covered either.

I rendered 234 cases against main on both content types and compared hashes, covering 17 collector shapes including `null`, primitives and an object with a shadowed prototype. Identical byte for byte, down to which error is thrown and from where. Bypassing the deduplication moves the digest, so the comparison is not asleep.

### DIFF
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -69,24 +69,24 @@ class Registry {
 			this.contentType === Registry.OPENMETRICS_CONTENT_TYPE;
 
 		for (const val of metric.values || []) {
-			let { metricName = name, labels = {} } = val;
-			const { sharedLabels } = val;
-			if (isOpenMetrics && metric.type === 'counter') {
-				metricName = `${metricName}_total`;
-			}
+			const { metricName = name, labels = {}, sharedLabels } = val;
+			const seriesName =
+				isOpenMetrics && metric.type === 'counter'
+					? `${metricName}_total`
+					: metricName;
 
+			// Make a copy before mutating
+			const seriesLabels =
+				defaultLabelNames === undefined ? labels : { ...labels };
 			if (defaultLabelNames !== undefined) {
-				// Make a copy before mutating
-				labels = { ...labels };
-
 				for (const labelName of defaultLabelNames) {
-					labels[labelName] ??= this._defaultLabels[labelName];
+					seriesLabels[labelName] ??= this._defaultLabels[labelName];
 				}
 			}
 
 			// We have to flatten these separately to avoid duplicate labels appearing
 			// between the base labels and the shared labels
-			const labelParts = formatLabels(labels, sharedLabels);
+			const labelParts = formatLabels(seriesLabels, sharedLabels);
 			if (sharedLabels !== undefined) {
 				// A histogram declared without label names still shares an empty
 				// object, and appending its flattened form would emit `{le="1",}`
@@ -96,7 +96,7 @@ class Registry {
 				}
 			}
 			const labelsString = labelParts.length ? `{${labelParts.join(',')}}` : '';
-			let fullMetricLine = `${metricName}${labelsString} ${getValueAsString(
+			let fullMetricLine = `${seriesName}${labelsString} ${getValueAsString(
 				val.value,
 			)}`;
 
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -70,7 +70,7 @@ class Registry {
 
 		for (const val of metric.values || []) {
 			let { metricName = name, labels = {} } = val;
-			const { sharedLabels = {} } = val;
+			const { sharedLabels } = val;
 			if (isOpenMetrics && metric.type === 'counter') {
 				metricName = `${metricName}_total`;
 			}
@@ -86,10 +86,15 @@ class Registry {
 
 			// We have to flatten these separately to avoid duplicate labels appearing
 			// between the base labels and the shared labels
-			const formattedLabels = formatLabels(labels, sharedLabels);
-
-			const flattenedShared = flattenSharedLabels(sharedLabels);
-			const labelParts = [...formattedLabels, flattenedShared].filter(Boolean);
+			const labelParts = formatLabels(labels, sharedLabels);
+			if (sharedLabels !== undefined) {
+				// A histogram declared without label names still shares an empty
+				// object, and appending its flattened form would emit `{le="1",}`
+				const flattenedShared = flattenSharedLabels(sharedLabels);
+				if (flattenedShared) {
+					labelParts.push(flattenedShared);
+				}
+			}
 			const labelsString = labelParts.length ? `{${labelParts.join(',')}}` : '';
 			let fullMetricLine = `${metricName}${labelsString} ${getValueAsString(
 				val.value,
--- a/test/registerTest.js
+++ b/test/registerTest.js
@@ -461,6 +461,51 @@ describe('Register', () => {
 		describe('Registry with default labels', () => {
 			const Registry = require('../lib/registry');
 
+			it('should not emit a label twice when a default label name is also a shared label', async () => {
+				// Histogram buckets carry the series labels as shared labels; a default
+				// label of the same name must not be emitted alongside them. `env` does
+				// not collide, so it has to survive the merge.
+				const r = new Registry(regType);
+				r.setDefaultLabels({ route: 'default-route', env: 'production' });
+
+				const histogram = new Histogram({
+					name: 'my_histogram',
+					help: 'my histogram',
+					registers: [r],
+					labelNames: ['method', 'route'],
+					buckets: [1],
+				});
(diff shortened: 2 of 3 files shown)


lib/registry.js
```diff
@@ -84,12 +84,21 @@ class Registry {
...
+				// Of the built-in metrics only Histogram sets shared labels, so for the
+				// rest none of the work below applies. Skipping it here also avoids
+				// handing flattenSharedLabels a per-series object its cache can never
+				// hit. A collector may still supply them, hence the branch.
+				labelParts = formatLabels(labels);
+			} else {
+				// We have to flatten these separately to avoid duplicate labels
+				// appearing between the base labels and the shared labels.
+				labelParts = [
+					...formatLabelsExcludingShared(labels, sharedLabels),
```
jdmarshall: If you need a large comment, you probably need a different implentation.

This function doesn't need to exist at all. All we need is 

```
if (sharedLabels) {
    const distinctLabels = labels.filter(label => {
        ...
    };

    labelParts = [...formatLabels(distinctLabels),
        ...
    ].filter(Boolean); // TODO: should this be run on the inputs instead of the outputs?
}
```

There's not enough code here to support the indirection. You're better reading the funky code in the funky conditional directly.

jdmarshall (changes requested):

lib/registry.js
milcho0604: Done in 6d8bbf5. The helper is gone and formatLabels() is back to main's, unchanged.

jdmarshall: Thanks! Rule of Three felt like a good mantra here. The change I was asking for actually broke up the existing Ro3

test/registerTest.js
```diff
@@ -461,6 +461,51 @@ describe('Register', () => {
 		describe('Registry with default labels', () => {
 			const Registry = require('../lib/registry');
 
+			it('should not emit a label twice when a default label name is also a shared label', async () => {
```
jdmarshall, line 493: Did this increase code coverage? I'm surprised we don't already have these tests.

jdmarshall (approved):
If you want to add the `!== undefined` we can still squeeze that in. The last PR for v0.16 is still awaiting review.

lib/registry.js
```diff
@@ -70,7 +70,7 @@ class Registry {
 
 		for (const val of metric.values || []) {
 			let { metricName = name, labels = {} } = val;
-			const { sharedLabels = {} } = val;
+			const { sharedLabels } = val;
```
jdmarshall: Why not inline this into the previous line?

lib/registry.js
```diff
@@ -86,10 +86,15 @@ class Registry {
 
 			// We have to flatten these separately to avoid duplicate labels appearing
 			// between the base labels and the shared labels
-			const formattedLabels = formatLabels(labels, sharedLabels);
-
-			const flattenedShared = flattenSharedLabels(sharedLabels);
-			const labelParts = [...formattedLabels, flattenedShared].filter(Boolean);
+			const labelParts = formatLabels(labels, sharedLabels);
```
jdmarshall: wait, why is this back?

jdmarshall (changes requested):
Looking for author commentary about how we ended up back at 2 argument formatLabels

lib/registry.js
milcho0604: That line deserved commentary and I skipped it, sorry. It happened in two steps and I only explained the second one.

I did implement the caller side filter, two ways. The entries shape and the shipped shape ran head to head in one process against trunk, and the literal object-copy translation ran separately under the same protocol.

The filter itself is a real win. Kept as a filter on the entries array, with `formatLabels()` staying one-argument, it beats main by 10% on all 42 `metrics()` cases. Translating your sketch literally, copying the non-shared labels into a fresh object, measured about 10% slower than main, and it drops a label named `__proto__` when a custom collector supplies one, because plain assignment on `{}` hits the prototype setter. So I threw that shape out. Head to head the entries version trails the shipped one on all 42 cases, median about 4%, the shallowest 1% and the deepest 8%, control centered on zero within about 3%. The difference is one extra filtered array per shared-label series, and histogram series are 93% of what the benchmark renders.

That's the whole reason the second argument came back, it won the measurement. If the one-argument API matters more than the 4%, the entries version is already written and passes the suite. Your call.

test/registerTest.js
milcho0604: Statements and lines stay at 98.4%, branches go 96.72% to 98.48% on registry.js. The paths were already being executed, what was missing was any assertion on what they produce.

On main you can delete the dedup exclusion, or the escaping inside `flattenSharedLabels()`, and the suite is still 551 green. Dropping every shared label after the first is noticed by exactly one thing, an OpenMetrics snapshot in the exemplar tests.

lib/registry.js
milcho0604: Folded. It needed a bit more than moving the line, so flag it if you would rather I put it back.

With `sharedLabels` in that destructure `prefer-const` fires on it, since the other two are reassigned below and it is not. Getting a single `const` destructure means the reassignments become derived values, so the `_total` suffix is now `seriesName` and the defensive copy is `seriesLabels`. Same conditions as before on both. The 234-case differential against main is still byte identical, including absent and null labels, default label collisions and the OpenMetrics counter naming.

jdmarshall (approved):

lib/registry.js
jdmarshall: So would you say, "on average 10% faster" would be a reasonable blurb for the release notes?

milcho0604: Yes, I think that works, with one caveat you might want in the wording.

I tested five registry shapes on Node 22, 24, 25 and 26, alternating before and after. The mean across the 16 results that contained a histogram was 9.8%, so your 10% matches what I see. An express shaped registry came out about 11 to 22% depending on the major, and default metrics on their own about 6 to 8%, since collection dominates that scrape more than formatting does.

The one shape that sits well outside that is a registry of only counters and gauges, about 68 to 98% faster. The suite would not surface it, because every case pairs a histogram with a counter and about 93% of the rendered series come from the histogram.

The cause seems to be the shared label cache. Once a histogram label set is warm all 14 of its lookups hit, and a newly observed label set costs one miss. Counter, Gauge and Summary series carry no shared labels at all, so the old destructuring default handed `flattenSharedLabels` a fresh empty object for every series of every scrape, which missed and then stored. I patched the two halves apart on a counter and gauge registry, and removing only the array churn is 1.09x, skipping only that cache traffic is 1.67x, with the merged change at 1.93x.

Your call on the wording of course. If it were mine I would keep the 10% and add a clause, something like `registry.metrics()` renders about 10% faster, and closer to twice as fast for registries made mostly of counters and gauges. These were noisy local runs so I would treat the endpoints as approximate, though the median was faster on every shape and every major.

Happy to pass along the scripts if you want to check any of it.

Issue #803, opened 2026-08-11
Worker lifecycle issues with lib/worker.js and lib/cluster.js


jdmarshall:
AI caught that we don't see a close event if a worker is terminated. Thinking about this code has revealed some more issues.

### We cannot tell if a worker has gone away

Problem is that https://github.com/nodejs/node/issues/59053 means that we never get the Worker object and therefore we cannot determine if the worker is dead, except by controlling the top level application workflows that are responsible for creating Workers in the first place, and that seems a bit mad especially given how many libraries there are that already do that.

And if the broadcast channel hasn't closed, then we don't know when one of our siblings has exited.  If a worker terminates after it has been announced, then all subsequent GET_METRICS_REQ calls get an unfulfilled promise and you get no more metrics at all.

I believe we will have to do some sort of supervisor functionality in the main thread. But I'm not aware of a way to enumerate all of the workers that are currently running.

### We don't remember historical values from workers and simply return no answer if any are missing

Every time a worker dies, counts get screwed up.

There may be some argument for keeping the last value recorded from a worker. Probably not gauges, but maybe everything else.

Because the Worker feature is going to be new to all users, we can ask the users to do a bit of additional bookkeeping to use this
functionality as table stakes for using this feature at all. We still have the luxury of adding that ask without listing it as a breaking change in the process. And it's a non-breaking change for Cluster if we ship it in v0.16 because dead workers has always been a fatal error there.

##Tasks

[ ] Document that reloading prom-client has undefined behavior in production mode.
[ ] Add debugging output (or console.warn?) to warn about existing listeners being found on initial call
[x] Simplify the promise logic in `clusterMetrics()` and `workerMetrics()` and store a promise for in-flight calls
[x] Implement dead workers support in the aggregator - aggregate all 'sum' type metrics to keep stats from backtracking
[x] Add a shutdown function for syncing data
[x] Add a shutdown event to worker.js, which flushes a 'swan song' last metric update to the aggregator to be retained in perpetuity
[x] Add a shutdown event to cluster.js, which flushes a 'swan song' last metric update to the aggregator to be retained in perpetuity
[x] Implement an ACK for the flush messages, because BroadcastChannel is complete Fire and Forget and `process.exit()` is likely to cause it to abort.
[x] Update the examples to match the new logic

jdmarshall:
[...]

PR #806, opened 2026-08-17
Support worker termination by collecting sum stats on exit.

jdmarshall:
This is the last PR required for the next semver-major release of the Prometheus Client. It addresses #803, via code changes, exposing additional lifecycle events for managing worker lifetimes, and documentation changes to warn of corner cases.

This change creates a retained mode for defunct workers where all of their final telemetry for 'sum' metrics is recorded in the aggregator process and aggregated down with each subsequent terminated worker. It also uses a call and response 
system so that the worker can ensure that the flushed data was acknowledged by the collector before exiting.

As part of this work, most of the 'thenables' have been removed from the `clusterMetrics()` code path, in favor of async/await which should also reduce the surface area of incompatibility with third party Promise libraries. And in general the code complexity of `clusterMetrics` has been reduced close to the essential complexity of this code, to satisfy Kernighan's Law with respect to spotting bugs in the implementation. 

Pains are taken to make sure that the `workers` and `historicalMetrics` fields are only ever read or written on the same tick, so that any shutdown operations that overlap with scraping don't smash into each other and cause data artifacts. You either get a scrape because the workers all replied before shutting down, or you get a transient timeout error that corrects itself on the subsequent collection.

The collector thread only collects 'sum' metric types - a dead process cannot participate in average, max, or min calculations any longer and those outliers shouldn't be retained once the process exits. However counts and histograms must be kept in aggregate to prevent sawtooth patterns in the collected data due to elision of previously-reported data.

This PR also reworks a number of tests that relied on `delay()` to instead watch the appropriate EventEmitters to watch for events that are being emitted asynchronously. There are a couple of cases where tests have around a 5-10% failure rate due to timing issues and if you run the suite over and over again, you will see false positives intermittently. Most of these should now be corrected. 

Because this is trying to keep two implementations in lockstep, it is unfortunately again a rather large PR, but hopefully not difficult to comprehend.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,16 +45,17 @@ This release marks our first release under the Prometheus umbrella.
 - perf: Use faster `process.memoryUsage.rss()` API for resident memory collection (30-40% more ops/sec)
 - fix: Browser compatibility for Gauge.startTimer()
 - ci: Run benchmarks for pull requests
-- ci: switch out deprecated benchmark-regression library for replacement
 - AggregatorRegistry renamed to ClusterRegistry, old name deprecated
 - chore: replace benchmark-regression dependency with faceoff
 - perf: Stat aggregation uses similar strategy to collection. 60% faster aggregation
 - chore: Add copyright license headers and test
-- Make cluster and worker-thread metric aggregation order deterministic
+- Support for worker threads added
+- Cluster and Worker improvements:
+  - Workers now opt in to collection
+  - Make cluster and worker-thread metric aggregation order deterministic
+  - Graceful shutdown support, and data retention past worker termination
 - Export `MetricObject`, `MetricObjectWithValues`, `MetricValue` and `MetricValueWithName` from the TypeScript definitions
 - chore: Old label processing code marked as deprecated
-- Improve cluster support to allow workers to opt out
-- Abort cluster metric responses during process termination
 
 ### Added
 
--- a/README.md
+++ b/README.md
@@ -649,3 +649,33 @@ To avoid native dependencies in this module, GC statistics for bytes reclaimed
 in each GC sweep are kept in a separate module:
 https://github.com/SimenB/node-prometheus-gc-stats. (Note that that metric may
 no longer be accurate now that v8 uses parallel garbage collection.)
+
+## Notes
+
+### Hot Reloading
+
+While it is unusual for NodeJS applications to reload modules at runtime, some more esoteric
+codebases may still do so. You should be aware that there is quite a bit of undefined behavior
+potential when combining this strategy with @prometheus-io/client_js, and one should proceed with
+caution if they do so.
+
+@prometheus-io/client_js has some primitive support for resetting itself between calls, but these
+features were intended almost exclusively for compatibility with test harnesses in order to
+support unit and integration tests. It has not been hardened to support workloads where a live
+application forcibly reloads modules to for instance pick up changes from the filesystem without
+restarting the server.
+
+Using this in production will result in data loss. Lost telemetry can trigger anything from alert to
+incorrect triage of production issues when 'sum' metrics show glitches or sawtooth patterns in their
+charts.
+
+While reloading your own code will often not also reload your dependencies, some extra care may be
+needed to avoid the common error message:
+
+> "A metric with the name #### has already been registered."
+
+Because while your code may have forgotten about that Gauge you already initialized, Prometheus
+remembers. One option is to utilize `Registry.getSingleMetric(name)` to do collision checks on
+initialization. Note that you will not be able to handle live code changes where the labels
+associated with a Metric without first removing the metric, which of course will also result in
+artifacts in your prometheus mackend due to existing counts and histograms being zeroed out.
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ See example folder for a sample usage. The library does not bundle any web
 framework. To expose the metrics, respond to Prometheus's scrape requests with
 the result of `await registry.metrics()`.
 
-### Usage with Node.js's `cluster` module
+### Usage with Node.js's `cluster` or Worker module
 
 Note: the Prometheus client now also supports worker threads, with much the same
 constraints as cluster workers. See `example/worker.js`.
@@ -51,6 +51,12 @@ Metrics are aggregated from the global registry by default. To use a different
 registry, call
 `client.AggregatorRegistry.setRegistries(registryOrArrayOfRegistries)` from the
 worker processes.
+
+#### Process Lifecycle
+
+The `shutdown()` method is provided to help cleanly shut down the application while
+metrics calls are pending. In the future this will also help with workers that are
+short-lived or need to be restarted.
 
 Please also see [The Workers Readme](Workers.md) for special notes on handling workers
 that do not survive for the entire run time of the application.
--- a/README.md
+++ b/README.md
@@ -20,6 +20,9 @@ the result of `await registry.metrics()`.
 
 ### Usage with Node.js's `cluster` module
 
+Note: the Prometheus client now also supports worker threads, with much the same
+constraints as cluster workers. See `example/worker.js`.
+
 Node.js's `cluster` module spawns multiple processes and hands off socket
 connections to those workers. Returning metrics from a worker's local registry
 will only reveal that individual worker's metrics, which is generally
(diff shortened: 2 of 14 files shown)


CHANGELOG.md
```diff
@@ -45,16 +45,16 @@ This release marks our first release under the Prometheus umbrella.
 - perf: Use faster `process.memoryUsage.rss()` API for resident memory collection (30-40% more ops/sec)
 - fix: Browser compatibility for Gauge.startTimer()
 - ci: Run benchmarks for pull requests
-- ci: switch out deprecated benchmark-regression library for replacement
```
krajorama, line 48: should this be removed?

CHANGELOG.md
```diff
@@ -45,16 +45,16 @@ This release marks our first release under the Prometheus umbrella.
...
 - chore: Add copyright license headers and test
-- Make cluster and worker-thread metric aggregation order deterministic
+- Support for worker threads added
+- Cluster and Worker improvements:
+  - Make cluster and worker-thread metric aggregation order deterministic
+  - Graceful shutdown support, and data retention past worker termination
 - Export `MetricObject`, `MetricObjectWithValues`, `MetricValue` and `MetricValueWithName` from the TypeScript definitions
 - chore: Old label processing code marked as deprecated
-- Improve cluster support to allow workers to opt out
-- Abort cluster metric responses during process termination
```
krajorama, line 57: should these be removed?

Workers.md
```diff
@@ -0,0 +1,81 @@
+# Notes on collection and short-lived processes
+
+## Introduction
+
+Worker threads present a large surface area for potential loss of telemetry data. In order to avoid
+strange artifacts in your statistics, it's recommendeng that you hook the process lifecycle events
```
krajorama:
```suggestion
strange artifacts in your statistics, it's recommended that you hook the process lifecycle events
```

lib/cluster.js
```diff
@@ -60,76 +65,136 @@ class AggregatorRegistry extends Registry {
...
 			.filter(worker => worker.isConnected())
 			.sort((left, right) => left.id - right.id);
 
-		return new Promise((resolve, reject) => {
-			let settled = false;
-			function done(err, result) {
-				if (settled) return;
-				settled = true;
+		if (orderedWorkers.length === 0) {
+			debug('No workers found for requestId', requestId);
```
krajorama, line 75: I wonder if this is necessary. Could we have  a situation when the application is idling and this keeps getting called?

test/clusterTest.js
```diff
@@ -150,6 +171,168 @@ describe.each([
...
+	});
+
+	describe('shutdown()', () => {
+		let AggregatorRegistry;
+		let listener;
+		let discovery;
+
+		beforeEach(() => {
+			jest.resetModules();
+			AggregatorRegistry = require('../lib/worker');
```
krajorama: Seems like copy paste bug?

```suggestion
			AggregatorRegistry = require('../lib/cluster');
```

test/registerTest.js
```diff
@@ -340,7 +340,36 @@ describe('Register', () => {
...
+
+				expect(output.length).toEqual(1);
+				expect(output[0].name).toEqual('test_metric');
+				expect(output[0].get).toBeInstanceOf(Function);
+			});
+
+			describe('with aggregator argument', () => {
+				it('should filter out other aggregators', async () => {
+					const max = getMetric('max_metric');
+					max.aggregater = 'max';
```
krajorama: typo:

```suggestion
					max.aggregator = 'max';
```

Workers.md
```diff
@@ -0,0 +1,81 @@
...
+
+['SIGHUP', 'SIGINT', 'SIGQUIT', 'SIGTERM'].forEach(sig => {
+  process.on(sig, async () => {
+    await registry.shutdown();
+    process.exit(0);
+  });
+});
+```
+
+See [the example](examples/workerTest.js) for a complete rundown, including the parent process
```
krajorama, line 70: typo

```suggestion
See [the example](examples/worker.js) for a complete rundown, including the parent process
```

README.md
```diff
@@ -41,13 +44,23 @@ for an example.)
...
+cardinality situation, which the Aggregator is generally meant to avoid.
 
 Metrics are aggregated from the global registry by default. To use a different
 registry, call
 `client.AggregatorRegistry.setRegistries(registryOrArrayOfRegistries)` from the
 worker processes.
 
+#### Process Lifecycle
+
+The `shutdown()' method is provided to help cleanly shut down the application while
```
krajorama: typo:

```suggestion
The `shutdown()` method is provided to help cleanly shut down the application while
```

lib/cluster.js
```diff
@@ -293,26 +388,50 @@ function announce() {
 }
 
 /**
- * Replace any listeners with new ones.
+ * Look for and complain about duplicate listeners.
+ *
+ * In test and delevopment mode, this function will remove duplicate listeners.
```
krajorama: typo:

```suggestion
 * In test and development mode, this function will remove duplicate listeners.
```

README.md
```diff
@@ -636,3 +649,33 @@ To avoid native dependencies in this module, GC statistics for bytes reclaimed
...
+features were intended almost exclusively for compatibility with test harnesses in order to
+support unit and integration tests. It has not been hardened to support workloads where a live
+application forcibly reloads modules to for instance pick up changes from the filesystem without
+restarting the server.
+
+Using this in production will result in data loss. Lost telemetry can trigger anything from alert to
+incorrect triage of production issues when 'sum' metrics show glitches or sawtooth patterns in their
+charts.
+
+While reloadng your own code will often not also reload your dependencies, some extra care may be
```
krajorama: typo:


```suggestion
While reloading your own code will often not also reload your dependencies, some extra care may be
```

README.md
```diff
@@ -636,3 +649,33 @@ To avoid native dependencies in this module, GC statistics for bytes reclaimed
 in each GC sweep are kept in a separate module:
 https://github.com/SimenB/node-prometheus-gc-stats. (Note that that metric may
 no longer be accurate now that v8 uses parallel garbage collection.)
+
+## Notes
+
+### Hot Reloading
+
+While it is unusual for NodeJS applications to reload modules at runtime, some more esoteric
+codebases may still due so. You should be aware that there is quite a bit of undefined behavior
```
krajorama: typo:

```suggestion
codebases may still do so. You should be aware that there is quite a bit of undefined behavior
```

lib/cluster.js
```diff
@@ -276,6 +352,25 @@ async function primaryListener(worker, event) {
...
+
+			worker.send({
+				type: ACK,
+				requestId: event.requestId,
+			});
+
+			try {
+				const metrics = [historicMetrics, ...event.metrics];
+
+				historicMetrics = await Registry.aggregate(metrics).getMetricsAsArray();
```
krajorama: Seems to me these are just function calls, not promises, so no `await` needed?

```suggestion
				historicMetrics = Registry.aggregate(metrics).getMetricsAsArray();
```

krajorama (commented):
First pass.

The general approach makes sense to me, but will need another pass to go into the details.

lib/cluster.js
jdmarshall: In debug mode you’re not idling you’re debugging.

CHANGELOG.md
jdmarshall: There’s two years of updates in chronological order in one release we don’t need to announce we added a library and upgraded it twice. 

> chore: replace benchmark-regression dependency with faceoff 

Is still in there

CHANGELOG.md
jdmarshall:
```
 - Cluster and Worker improvements:
  - Make cluster and worker-thread metric aggregation order deterministic
  - Graceful shutdown support, and data retention past worker termination
```

I suppose “Improve cluster support to allow workers to opt out” got dropped

lib/cluster.js
jdmarshall: Not while you’re debugging. Knowing that there are no workers is information. Could be a bug in announcements I missed. 

Mostly this helped tremendously getting the mocks right in the tests. They were a mess, particularly in a couple of the contributed tests. I literally spent the majority of my time on this fixing tests I broke with the added events.

lib/cluster.js
jdmarshall: Hmm. That’s odd. One is sync and the rest are async. However did we manage that?

But I should double check that collect() is being called in the worker because I bet I missed one. 

And this needs to be fixed in both.

Also there's an unhandledRejectionError surface area here that I missed.  This code evolved during testing and I think I ran out of steam before it was finished.

test/clusterTest.js
jdmarshall: Looks like I missed one, yep.  Trying to make sure they test the same corner cases in the same way, so if someone adds a new feature...

krajorama (commented):
lgtm, llm claims some flakiness/cross pollination in test only - I'll open a PR for that afterwards, I need to understand it myself.

Please resolve the remaining small typo/fixe and I can approve.

krajorama (approved):
LGTM, I've also done a simple smoke test locally. should be good for RC

Issue #812, opened 2026-08-20
Reimplement 'labels' as a derivative of the Metric instead of a closure


jdmarshall:
With the existence of LabelMap, there are simpler mechanisms available to give multiple callers a set of labels to apply to all metrics calls. 

Problem areas:

 - copy constructor
 - reset - I already have a TODO in the code about how reset is poorly implemented, and this will need to be fixed.
 - metrics gathering - a derived metric has already had its stats reported and should return empty set

In the bargain however, this one feature addresses #777, #407, and #598 all without introducing performance regressions.

Issue #622, opened 2024-03-27
Gauge without label should not have a default value of 0 (zero)


thomaschaaf:
A gauge that is registered but does not have a value set is currently set to 0 (zero). I believe this behavior to be incorrect because a gauge without a label does not work like it.

```
const client = require('prom-client');
const gauge = new client.Gauge({ name: 'metric_name', help: 'metric_help', labelNames: ['example'] });
const gauge2 = new client.Gauge({ name: 'metric_name2', help: 'metric_help2' });

console.log(await client.register.metrics())

gauge.set({ example: 'true' }, 1)

console.log(await client.register.metrics())

gauge.reset();

console.log(await client.register.metrics())
```

Output 1 after initialization:
```
# HELP metric_name metric_help
# TYPE metric_name gauge

# HELP metric_name2 metric_help2
# TYPE metric_name2 gauge
metric_name2 0
```

Output 2 after setting gauge:
```
# HELP metric_name metric_help
# TYPE metric_name gauge
metric_name{example="true"} 1

# HELP metric_name2 metric_help2
# TYPE metric_name2 gauge
metric_name2 0
```

Output 3 after reset:
```
# HELP metric_name metric_help
# TYPE metric_name gauge

# HELP metric_name2 metric_help2
# TYPE metric_name2 gauge
metric_name2 0
```

I would expect the output 1 of both gauges to be the same and instead of a default of 0 having a default of undefined.

jdmarshall:
I am very confused by this feature. I cannot imagine a situation where I'd want metrics with absolutely no labels in the first place. I'm going to have a peek at the Prometheus docs to see what if anything they have to say about it. You always want at least one tag if you have more than one machine, and probably at least 2 if you're doing CI/CD so you know if all the errors are coming from the deployment/canary.

I've already said that the benchmarks with 0 labels are providing misinformation that has encouraged several contributors to submit PRs that move us in the wrong direction. 

The v0.16 release will show a performance regression for them because it makes labeled metrics much faster to do more work during the constructor, at a small but noticeable cost for metrics with no labels.

zbjornson:
It's not unusual. Sometimes the [automatically generated labels](https://prometheus.io/docs/concepts/jobs_instances/) (`job`, `instance`) and/or ones created using relabeling rules during collection are the only ones used.

jdmarshall:
[...]

PR #813, opened 2026-08-21
Don't zero out Gauges. 

jdmarshall:
This PR started as exploratory work for a different issue - there's been a TODO that I added to metric.js that fights the work needed for #812. I had already decided that doing #812 all in a single PR would probably get a little bit on the long side, so I was looking for some stopping point in the middle to split it up.

I noticed that if #622 were fixed, then a default `reset()` implementation would be used by 2 of the 4 builtin metric types.

This PR initializes the store in metrics.js, and then retains the LabelMap object across `reset()` calls, and removes the zeroing logic from `Gauge.reset()`, thereby fixing #622.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Breaking
 
+- Gauges no longer initialize to zero, even if they have no labels
+- `reset()` changes that may require modifications to custom Metrics:
+  - it is no longer called in the middle of the constructor
+  - it now defaults to clearing the labelMap
+
 ### Changed
 
 - Organized default metrics
--- a/lib/counter.js
+++ b/lib/counter.js
@@ -18,7 +18,7 @@
 'use strict';
 
 const util = require('util');
-const { isObject, getLabels, nowTimestamp, LabelMap } = require('./util');
+const { isObject, getLabels, nowTimestamp } = require('./util');
 const { Metric } = require('./metric');
 const Exemplar = require('./exemplar');
 
@@ -34,6 +34,10 @@ class Counter extends Metric {
 			this.inc = this.incWithExemplar;
 		} else {
 			this.inc = this.incWithoutExemplar;
+		}
+
+		if (this.labelNames.length === 0) {
+			this.store.set({}, 0);
 		}
 	}
 
@@ -102,7 +106,8 @@ class Counter extends Metric {
 	 * @returns {void}
 	 */
 	reset() {
-		this.store = new LabelMap(this.labelNames);
+		this.store.clear();
+
 		if (this.labelNames.length === 0) {
 			this.store.set({}, 0);
 		}
--- a/lib/gauge.js
+++ b/lib/gauge.js
@@ -19,7 +19,7 @@
 
 const util = require('util');
 
-const { getLabels, isObject, LabelMap } = require('./util');
+const { getLabels, isObject } = require('./util');
 const { Metric } = require('./metric');
 
 class Gauge extends Metric {
@@ -38,17 +38,6 @@ class Gauge extends Metric {
 		value = getValueArg(labels, value);
 		labels = getLabelArg(labels);
 		set(this, labels, value);
-	}
-
-	/**
-	 * Reset gauge.
-	 * @returns {void}
-	 */
-	reset() {
-		this.store = new LabelMap(this.labelNames);
-		if (this.labelNames.length === 0) {
-			this.store.set({}, 0);
-		}
 	}
 
 	/**
--- a/lib/metric.js
+++ b/lib/metric.js
@@ -15,7 +15,7 @@
 'use strict';
 
 const Registry = require('./registry');
-const { isObject } = require('./util');
+const { isObject, LabelMap } = require('./util');
 const { validateMetricName, validateLabelName } = require('./validation');
 
 /**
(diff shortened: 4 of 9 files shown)


CHANGELOG.md
```diff
@@ -7,12 +7,26 @@ project adheres to [Semantic Versioning](http://semver.org/).
...
+- Gauges no longer initialize to zero, even if they have no labels
+- `reset()` changes that may require modifications to custom Metrics:
+  - it is no longer called in the middle of the constructor
+  - it now defaults to clearing the labelMap
+
+### Changed
+
+### Added
+
+## [0.16.0] - 2026-08-??
```
jdmarshall: This got picked up from another PR that needs to land for v0.16.0

lib/metric.js
jdmarshall, line 64: So my IDE is telling me we don't use these values anymore, and with the LabelMap call right below this (which also does this sort) and the deprecations in util.js, it's looking more and more like we can just drop these. Which would also be a little more memory for #800

lib/counter.js
```diff
@@ -35,6 +35,10 @@ class Counter extends Metric {
 		} else {
 			this.inc = this.incWithoutExemplar;
 		}
+
```
krajorama, line 38: I think we should discuss counters separately. Also doesn't match the PR title.

In client_golang, whether to expose a counter with a 0 value right away depends on which way you initialize it. If it doesn't have any labels or it only has constant labels (labels set at constructor with name and value), than it's exposed as 0. If it has any dynamic label values, then it's not exposed until you give it a label value(s) and an observed value.

If I understand the code correctly, this client only has dynamic labels , meaning you cannot set a value for a label at the constructor, you always have to define it when measuring? Which means this code is correct, since the only constant labels case is no labels.

lib/gauge.js
```diff
@@ -40,17 +40,6 @@ class Gauge extends Metric {
 		set(this, labels, value);
 	}
 
-	/**
-	 * Reset gauge.
-	 * @returns {void}
-	 */
-	reset() {
-		this.store = new LabelMap(this.labelNames);
-		if (this.labelNames.length === 0) {
```
krajorama, line 49: To be inline with client_golang, this should remain. Same argument: In client_golang, whether to expose a ~counter~ gauge with a 0 value right away depends on which way you initialize it. If it doesn't have any labels or it only has constant labels (labels set at constructor with name and value), than it's exposed as 0. If it has any dynamic label values, then it's not exposed until you give it a label value(s) and an observed value.

If I understand the code correctly, this client only has dynamic labels , meaning you cannot set a value for a label at the constructor, you always have to define it when measuring? Which means this code is correct, since the only constant labels case is no labels.

krajorama (changes requested):
I think it makes sense to keep the current behavior to be in line with what client_golang and client_java does. It could be the basis of introducing constant labels later. See comment inside.

lib/counter.js
jdmarshall: This code is exactly the same as previous versions, as you'll note that I did not change any Counter unit tests.

What's different is that the base Metrics constructor used to call `reset()`, which is an abstract method, in the middle of the constructor, which is a bad practice. Without calling reset however, someone has to assert the default value, which is what is happening here. 

You've felt the need to comment on every 'large' PR I've done, which is a pattern that usually means 'complaint'. So this PR includes refactoring work to make #812 more manageable when I get back to it. As will #814. Changes need to be made, so changes will me made. Not as separate PRs where they go to die, but amortized across other work.

lib/gauge.js
jdmarshall: I cannot find where that particular logic is in the go client. Can you point me to it? If so then the Issue is wontfix.

jdmarshall: @krajorama pointer to this logic in the Go client?

krajorama: The difference comes down to how the metric is collected for exposure (this is similar to the get metrics message here). A constant label counter Counter is derived from Collector that simply returns itself when [Collect](https://github.com/prometheus/client_golang/blob/89c60c0554699c74fa3d4eec472b98a3108905e3/prometheus/collector.go#L118)ed;  whereas a dynamic labels CounterVec(tor) is derived from MetricVec, which is a metricMap, where the [Collect](https://github.com/prometheus/client_golang/blob/89c60c0554699c74fa3d4eec472b98a3108905e3/prometheus/vec.go#L333) method only returns items from the dynamic map.

test/gaugeTest.js
```diff
@@ -138,12 +138,13 @@ describe.each([
...
 
-			it('should init to 0', async () => {
+			it('should not init to 0', async () => {
 				instance = new Gauge({
 					name: 'init_gauge',
 					help: 'somehelp',
 				});
-				await expectValue(0);
+
+				await expect(instance.get()).resolves.toBeEmpty;
```
krajorama, line 147: claude claims this is no-op, should be something like
```
expect((await instance.get()).values).toEqual([]);
```

test/gaugeTest.js
```diff
@@ -297,7 +299,7 @@ describe.each([
 			expect((await instance.get()).values[0].value).toEqual(12);
 
 			instance.reset();
-			expect((await instance.get()).values[0].value).toEqual(0);
+			expect(await instance.get()).toBeEmpty;
```
krajorama, line 302: claude claims this is no-op, should be something like

expect((await instance.get()).values).toEqual([]);

krajorama (approved):
Approved as agreed, acknowledging it's a breaking change. Will have to be explained in the release notes.

Added some Claude findings as comments, could be follow up as well.

PR #815, opened 2026-08-21
perf(histogram): render the export list straight from the store iterator

milcho0604:
The Array.from cleanup you suggested after #804. Only one call site qualified. LabelMap.values() hands back the raw Map iterator, so the histogram export can map and reduce over it directly and both intermediate arrays go away. The other Array.from uses in lib feed the public API, which is typed as real arrays, or sit on cold paths, so I left those alone.

Measurements are in my earlier comment at https://github.com/prometheus/client_js/pull/804#issuecomment-5306535878. It helps 24 and 26 at high series counts and can cost a few percent on 22 under default heap settings. You called that acceptable for a major release with a note, so the CHANGELOG entry says exactly that.

I verified the rendered output is byte identical to main on 22, 24 and 26, and the suite passes on all three.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Organized default metrics
+- perf: Histogram rendering builds its export list straight from the store iterator instead of an intermediate array. Faster at high series counts on Node 24 and 26, can be slightly slower on Node 22
 
 ### Added
 
--- a/benchmarks/histogram.js
+++ b/benchmarks/histogram.js
@@ -14,7 +14,11 @@
 
 'use strict';
 
-const { getLabelNames, labelCombinationFactory } = require('./utils/labels');
+const {
+	getLabelNames,
+	getLabelCombinations,
+	labelCombinationFactory,
+} = require('./utils/labels');
 
 module.exports = setupHistogramSuite;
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -57,6 +57,7 @@ This release marks our first release under the Prometheus umbrella.
 - Export `MetricObject`, `MetricObjectWithValues`, `MetricValue` and `MetricValueWithName` from the TypeScript definitions
 - chore: Old label processing code marked as deprecated
 - perf: Stop rebuilding the label array for every rendered series, and skip shared label handling entirely for values that do not set it; 9-22% faster `metrics()` in the registry benchmarks
+- perf: Histogram rendering builds its export list straight from the store iterator instead of an intermediate array. Faster at high series counts on Node 24 and 26, can be slightly slower on Node 22
```
jdmarshall: This needs to move to the new section now that 0.16 is out.

benchmarks/histogram.js
```diff
@@ -117,6 +134,25 @@ function setup(labelCount) {
...
+
+		const histogram = new client.Histogram({
+			name: 'histogram',
+			help: 'histogram',
+			labelNames: getLabelNames(labelCounts.length),
+			registers: [registry],
+		});
+
+		getLabelCombinations(labelCounts).forEach(labels =>
+			histogram.observe({ ...labels }, 1),
```
jdmarshall, line 149: Do all of our histogram tests look like this?  Putting all the results into one bucket is probably not quite testing the whole flow, though maybe it doesn't matter here.

jdmarshall (approved):

PR #816, opened 2026-08-21
test: fix BroadcastChannel leakage in worker.js test suite

krajorama:
## What

Follow-up to #806. While reviewing that PR, `test/workerTest.js`'s "aggregates worker responses in thread id order" and two `shutdown()` tests turned out to only pass by accident: `jest.resetModules()` creates a fresh `WorkerRegistry` module instance per test, each with its own `ANNOUNCEMENT_CHANNEL` and per-thread `BroadcastChannel`. Those channels keep receiving messages until explicitly closed — even once nothing in JS still references them — so a previous test's listeners stayed alive and kept reacting to later tests' messages on the same channel names.

That leakage is exactly what let those three tests pass at all: their `discovery` promise, which waits for a non-primary `ANNOUNCEMENT` echo, only ever resolved via a stray listener left over from an earlier test — never from anything the test itself set up. Run any of the three in true isolation (`-t "..."`) against `main` and they fail outright (timeout or `InvalidStateError: BroadcastChannel is closed`).

## How

- Add a guarded `WorkerRegistry.resetForTesting()` (throws outside `test`/`development`, mirroring the existing guard on `cluster.js`'s `scanListeners`) that closes `ANNOUNCEMENT_CHANNEL`, the per-thread channel, and every registered worker's channel, then resets `workers`/`requests`/`historicMetrics`/`listenersAdded`.
- Call it everywhere a test loads a fresh module instance via `jest.resetModules()`, including a 30-iteration loop that previously leaked 30 channels per run and closed none.
- Replace the `discovery`/echo pattern in the two affected tests with polling the already-existing `WorkerRegistry.workerCount()` — a real signal, rather than hoping to observe an echo that a self-posted `BroadcastChannel` message never actually delivers back to its own sender.

## Test plan

- [x] `npx jest` — full suite passes (581/581 on current `main`)
- [x] `npx eslint . && npx prettier . --check && npx tsc --project .` — all clean
- [x] Previously-flaky tests pass in the full file run (30/30 across repeated runs)
- [x] Previously-flaky tests pass in true isolation too (`-t "..."`, 5/5 each) — this is the actual proof the fragility is gone, not just hidden
- [x] The "Jest did not exit... asynchronous operations" open-handle warning that showed up intermittently before this fix no longer appears

Signed-off-by: György Krajcsovits <gyorgy.krajcsovits@grafana.com>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -46,6 +46,7 @@ let requestCtr = 0; // Concurrency control
 const requests = new Map(); // Pending requests for workers' local metrics.
 const workers = new Map();
 let historicMetrics = []; // Metrics from dead workers
+let ownChannel; // This thread's own broadcast channel, set once listeners are added.
 
 class WorkerRegistry extends Registry {
 	/**
@@ -232,6 +233,36 @@ class WorkerRegistry extends Registry {
 	}
 
 	/**
+	 * Close every channel and listener opened by this module instance.
+	 *
+	 * `BroadcastChannel`s created here keep receiving messages until closed,
+	 * even once nothing in JS still references them. Test suites that call
+	 * `jest.resetModules()` between cases therefore need this to release the
+	 * previous instance's channels; otherwise its listeners keep reacting to
+	 * later tests' messages on the same channel names. Not for production use.
+	 * @returns {void}
+	 */
+	static resetForTesting() {
+		if (!['test', 'development'].includes(process.env.NODE_ENV)) {
+			throw new Error(
+				'WorkerRegistry.resetForTesting() is only available in test or development mode.',
+			);
+		}
+
+		ANNOUNCEMENT_CHANNEL.close();
+		ownChannel?.close();
+
+		for (const entry of workers.values()) {
+			entry.channel.close();
+		}
+
+		workers.clear();
+		requests.clear();
+		historicMetrics = [];
+		listenersAdded = false;
+	}
+
+	/**
 	 * Sets the registry or registries to be aggregated. Call from workers to
 	 * use a registry/registries other than the default global registry.
 	 * @param {Array<Registry>|Registry} regs Registry or registries to be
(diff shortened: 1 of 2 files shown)


lib/worker.js
```diff
@@ -231,6 +232,36 @@ class WorkerRegistry extends Registry {
 		return workers.size;
 	}
 
+	/**
```
jdmarshall, line 235: Okay, I think I'm with you so far here.  This is also definitely forward progress for an ESM migration. Those are absolutely no fun.

lib/worker.js
```diff
@@ -231,6 +232,36 @@ class WorkerRegistry extends Registry {
...
+	 * @returns {void}
+	 */
+	static resetForTesting() {
+		if (!['test', 'development'].includes(process.env.NODE_ENV)) {
+			throw new Error(
+				'WorkerRegistry.resetForTesting() is only available in test or development mode.',
+			);
+		}
+
+		ANNOUNCEMENT_CHANNEL.close();
```
jdmarshall, line 252: You close ANNOUNCEMENT_CHANNEL and then I don't see where it gets re-opened. So I'm a little concerned why the tests are passing in this case. Some same-process weirdness with BroadcastChannels perhaps?

test/workerTest.js
```diff
@@ -79,7 +76,13 @@ describe.each([
 				return { threadId, channel };
 			});
 
-			await discovery; // Let announcements arrive
+			// Wait for the primary to register all 3 fake workers, rather than
```
jdmarshall, line 79: I remember someone telling me that `delay()` was a bad idea. Why not keep the `await discovery` here?

test/workerTest.js
```diff
@@ -44,10 +44,15 @@ describe.each([
...
+			jest.resetModules();
 			const WorkerRegistry = require('../lib/worker');
 			const registry = new WorkerRegistry(regType);
-			const metrics = await registry.workerMetrics();
-			expect(metrics).toEqual('');
+			try {
+				const metrics = await registry.workerMetrics();
+				expect(metrics).toEqual('');
+			} finally {
+				WorkerRegistry.resetForTesting();
```
jdmarshall, line 54: Wouldn't we be better off setting this at the top? And why both reset() types at the same time?

test/workerTest.js
```diff
@@ -262,7 +263,13 @@ describe.each([
 					}
 				});
 
-				await discovery;
+				// Wait for the primary to register this fake worker, rather than
```
jdmarshall, line 266: Same question as above.

PR #817, opened 2026-08-21
Alphabetize default metrics

### LINKED ISSUE #810: Alphabetize default metrics
Metrics are being reported in the order they were seen. This is a peculiarity of Object.keys() that persisted when we switched to Maps for metric storage. 

In conjunction with https://github.com/prometheus/client_js/pull/689, which will allow people to disable particular default metrics, and #539, which has a side effect of reducing the degree of label reordering between cluster_metrics calls, I think it would be useful from an ergonomics standpoint if someone reordered the defaultMetrics to register the metrics in order of their metric name.

Since each of the metrics uses a common prefix for all of its own stats, this should mostly be a matter of rearranging the entries in the object in lib/defaultMetrics.js to order the prefixes alphabetically, and then inspecting the output from one of the examples to verify. 

Some defaults like heapSizeAndUsed may need a little tweak to make sure the first metric registered is the one with the collect() on it (currently that's total, which is alphabetically last).

jonsuguiyama:
Closes #810.

- Reorder the `metrics` object in `lib/defaultMetrics.js` alphabetically by each module's Prometheus metric name.
- In `heapSizeAndUsed.js`, move the shared `collect()` from the `total` gauge to the `external` gauge, since `nodejs_external_memory_bytes` now registers first alphabetically.
- Add a test verifying default metric modules register in alphabetical order.
- Update the existing heapSizeAndUsed test to read the `external` gauge first, since it now owns the shared collect.

### DIFF
--- a/lib/defaultMetrics.js
+++ b/lib/defaultMetrics.js
@@ -32,26 +32,28 @@ const heapSpacesSizeAndUsed = require('./metrics/heapSpacesSizeAndUsed');
 const version = require('./metrics/version');
 const gc = require('./metrics/gc');
 
+// Ordered alphabetically by the default Prometheus metric name(s) each
+// module registers, so metrics are reported in a stable, predictable order.
 const metrics = {
-	processCpuTotal,
-	processStartTime,
-	osMemoryHeap,
-	processOpenFileDescriptors,
-	processMaxFileDescriptors,
-	eventLoopLag,
-	eventLoopUtilization,
+	processHandles,
+	processRequests,
 
 	...(typeof process !== 'undefined' &&
 	// eslint-disable-next-line n/no-unsupported-features/node-builtins
 	typeof process.getActiveResourcesInfo === 'function'
 		? { processResources }
 		: {}),
-	processHandles,
-	processRequests,
+	eventLoopLag,
+	eventLoopUtilization,
 	heapSizeAndUsed,
+	gc,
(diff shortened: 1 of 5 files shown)


jdmarshall (commented):
I'm on my way out the door at the moment but looks promising. Thank you.

lib/defaultMetrics.js
```diff
@@ -32,26 +32,28 @@ const heapSpacesSizeAndUsed = require('./metrics/heapSpacesSizeAndUsed');
...
 		? { processResources }
 		: {}),
-	processHandles,
-	processRequests,
+	eventLoopLag,
+	eventLoopUtilization,
 	heapSizeAndUsed,
+	gc,
 	heapSpacesSizeAndUsed,
 	version,
```
jdmarshall: I do want to discuss one exception to the alphabetization rule. 

node_version isn't really a stat, it's a configuration parameter that's part of either the build or the deployment. Having it buried like this, only someone who is really looking for it will find it, and this is, IME, one of a small number of pieces of data that people need to see more often than they believe they need to see it.

If the version is the first thing you see, then it is more likely to be seen. So I think this should be the exception to the alphabetical rule and should be at the top of the output.

jdmarshall (changes requested):

jdmarshall (approved):

PR #819, opened 2026-08-22
Release Prep

jdmarshall:
We have agreed to do the release on Monday, now that we've sorted out the workflow.


krajorama (approved):
lgtm, but let's do it Monday as planned.

PR #820, opened 2026-08-22
fix(registry): default to summation for metrics without an aggregator

chuanghiduoc:
## Problem

Both the README ("Custom metrics are summed across workers by default") and the worker/cluster JSDoc ("or by summation if `aggregator` is undefined") promise a `'sum'` fallback for metrics without an explicit `aggregator`. Two recent changes broke that contract:

1. `Registry.aggregate()` throws `'undefined' is not a defined aggregator.` when any snapshot lacks the property — snapshots produced by `registerMetric()` with a plain object, or hand-written wrappers. This crashes `clusterMetrics()` / `workerMetrics()` entirely.
2. `getMetricsAsArray('sum')` silently filters those metrics out (`aggregator === metric.aggregator` is false when the property is missing), so their values are lost on the shutdown-flush path added for workers in ef1cea2.

Reproduction on current main:

```js
const { Registry } = require('prom-client');
Registry.aggregate([
  [{ help: 'h', name: 'm', type: 'gauge', values: [{ value: 5, labels: {} }] }],
]);
// Error: 'undefined' is not a defined aggregator.
```

and:

```js
const register = new Registry();
new Counter({ name: 'normal', help: 'h', registers: [register] }); // aggregator defaults to 'sum'
register.getMetricsAsArray('sum').length; // 0 if the metric object lacks .aggregator
```

Note that `new Counter({...})` sets `aggregator: 'sum'` via the Metric constructor default, so built-in metric classes mask the issue; it only surfaces with wrapper/plain-object metrics, which is exactly what the aggregation and shutdown paths produce internally.

## Fix

Restore the documented `'sum'` fallback in both places (`metric.aggregator ?? 'sum'`).

## Testing

- two new tests in `test/registerTest.js`: aggregate() sums snapshots without an aggregator; getMetricsAsArray('sum') / getMetricsAsJSON('sum') include them
- full unit suite: 29 suites / 581 tests passing

### DIFF
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -53,7 +53,8 @@ class Registry {
 			return Array.from(values);
 		} else {
 			return Array.from(
-				values.filter(metric => aggregator === metric.aggregator),
+				// metrics without an explicit aggregator default to 'sum'
+				values.filter(metric => aggregator === (metric.aggregator ?? 'sum')),
 			);
 		}
 	}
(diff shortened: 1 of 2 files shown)


jdmarshall (approved):
Hmmm.. I had code for this but then convinced myself it wasn't needed.  :/

PR #824, opened 2026-08-26
perf: remove truthy checks from default metrics

### LINKED ISSUE #823: Truthiness Reduction Program 2026
A number of PRs have reduced or removed truthiness checks over time, particularly once nullish operators were added to the code. Truthy checks force object coercion, which is both extra clock cycles and additional GC overhead. Even if there aren't many of these left on hot paths, there are for sure a few and the rest still hurt, just not as much. Also nullish and coalescing operators all date back to 14 (one is on 16 but I can't recall which) and we support >=20 so it's overdue.

If anyone would like to do a PR to remove most of the rest, I'd like to see that. Please however be aware of the existing open PRs as merge conflicts slow the review process down more than just about anything else. 

Don't forget to look for ternary operators. There are a few hiding there.

wangxpych:
Closes #823.

This removes the remaining truthy defaults and optional-object checks from the default metric collectors. Configuration defaults now use nullish coalescing, while optional registries, `perf_hooks`, memory usage results, and active spans use explicit `undefined` checks.

The change is limited to `lib/metrics` so it does not overlap the core metric, histogram, registry utility, or worker files currently touched by open pull requests. Supported input behavior is unchanged, and the changelog records the performance cleanup.

Verification:

- `npm test`
  - ESLint
  - Prettier check
  - TypeScript compilation
  - 590 Jest tests with coverage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 - Organized default metrics
 - perf: Histogram rendering builds its export list straight from the store iterator instead of an intermediate array. Faster at high series counts on Node 24 and 26, can be slightly slower on Node 22
 - fix: Correct content type exported for cluster and worker mode.
+- perf: Remove truthy conditionals from default metric collectors
 
 ### Added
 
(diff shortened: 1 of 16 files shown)


jdmarshall (changes requested):
Rebase to settle merge conflict in the CHANGELOG

jdmarshall (approved):
Good to go

PR #826, opened 2026-08-26
Fix incorrect scraping with contentType === OpenMetrics

### LINKED ISSUE #827: counter, worker mode output Prometheus metrics even when when set to OpenMetrics
This problem was reported by @gslandtreter in #740 but missed the cutoff for 0.16 due to logistics problems.

In the meantime that code has changed substantially while working on other features, so the PR doesn't quite work as provided. A new one is being written.

jdmarshall:
The historical data does not need to be modified, mercifully, because it gets modified during any aggregate() call, which should then also cover anyone trying to support both endpoints at once.

Fixes #827 

Note: This should have shipped with 0.16.0 and got lost in the shuffle.  The author who caught this did not get back to me in time. This implementation is compatible with the memory leak changes that were pushed and hopefully with the additional changes that are in PR at the moment.

### DIFF
--- a/test/clusterTest.js
+++ b/test/clusterTest.js
@@ -97,7 +97,23 @@ describe.each([
 		it('works properly if there are no cluster workers', async () => {
 			const ar = new AggregatorRegistry(regType);
 			const metrics = await ar.clusterMetrics();
-			expect(metrics.trim()).toEqual('');
+
+			if (regType === Registry.OPENMETRICS_CONTENT_TYPE) {
+				expect(metrics).toContain('# EOF\n');
+			} else {
+				expect(metrics.trim()).toEqual('');
+			}
+		});
+
+		it('formats in the correct content type', async () => {
+			const ar = new AggregatorRegistry(regType);
+			const metrics = await ar.clusterMetrics();
+
+			if (regType === Registry.OPENMETRICS_CONTENT_TYPE) {
+				expect(metrics).toContain('# EOF\n');
+			} else {
+				expect(metrics.trim()).toEqual('');
+			}
 		});
 
 		it("listeners don't accumulate", () => {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 - Organized default metrics
 - perf: Histogram rendering builds its export list straight from the store iterator instead of an intermediate array. Faster at high series counts on Node 24 and 26, can be slightly slower on Node 22
+- fix: Correct content type exported for cluster and worker mode.
 
 ### Added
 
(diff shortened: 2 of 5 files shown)


test/clusterTest.js
```diff
@@ -97,7 +98,23 @@ describe.each([
...
-			expect(metrics.trim()).toEqual('');
+
+			if (regType === Registry.OPENMETRICS_CONTENT_TYPE) {
+				expect(metrics).toContain('# EOF\n');
+			} else {
+				expect(metrics.trim()).toEqual('');
+			}
+		});
+
+		it('formats in the correct content type', async () => {
```
krajorama, line 108: it's the same test, could we just rename and have one test?

test/clusterTest.js
```diff
@@ -97,7 +98,23 @@ describe.each([
 		it('works properly if there are no cluster workers', async () => {
 			const ar = new AggregatorRegistry(regType);
 			const metrics = await ar.clusterMetrics();
-			expect(metrics.trim()).toEqual('');
+
+			if (regType === Registry.OPENMETRICS_CONTENT_TYPE) {
```
krajorama: nit: I generally prefer table based/data driven tests over ifs. But not blocker, maybe next PR or when we'll have a third one.

krajorama, line 101: nit: I usually prefer the expected result to be part of `describe` and not to use `if` in the test

test/clusterTest.js
```diff
@@ -18,6 +18,7 @@ const cluster = require('cluster');
 const process = require('process');
 const Registry = require('../lib/cluster');
 const { setTimeout: delay } = require('timers/promises');
+const AggregatorRegistry = require('../lib/cluster');
```
krajorama: tests pass without this

```suggestion
```

test/clusterTest.js
krajorama, line 108: nit: the `_total` change isn't tested here, only for the worker.

krajorama (approved):
I've got some nits, but the fix is correct

test/clusterTest.js
jdmarshall: And I prefer not to write tests in loops if the loops aren’t universally identical, but here we are :)

These should be two separate test files that you can diff against each other to reason about. Actually most of these tests don’t even vary by version. They’re mostly useful to detect bugs in the teardown logic, as you and I have already witnessed.

test/clusterTest.js
jdmarshall: My IDE trying to be helpful.

test/clusterTest.js
jdmarshall: Same conditions different assertions aren’t the same tests. Requirements can change separately, and OM v2 support will change these tests.

jdmarshall: We do have other tests that are asserting that. This is a functional test. 

Trying to recall why I don't just look for EOF in both, and I think the reason is that within the test harness, because of the way we handle the listener 'leaks', there are problems with the main thread listening to its own REQ calls.  So instead of EOF we just get "" if there are no workers registered.

Issue #576, opened 2023-06-29
Support for native histograms


jml:
Native histograms (aka sparse histograms, aka high-res histograms) are on their way to being production ready. It would be wonderful if the Node prom-client supported them.

Some relevant links:

- [Prometheus milestone for full native histogram support](https://github.com/prometheus/prometheus/milestone/10)
- [PromCon 2022 talk introducing native histograms](https://promcon.io/2022-munich/talks/native-histograms-in-prometheus/)
- [Go client implementation of histograms, including support for native histograms](https://github.com/prometheus/client_golang/blob/main/prometheus/histogram.go)

equinsuocha:
I think it worth discussing again, as native histograms are considered a stable feature, so why not implement it?
Some ground work on [protobuf](https://github.com/siimon/prom-client/pull/108) metric exposure has been done quite a while ago in addition to the native histogram spec and reference implementations.

nicolas-fidel-wmx:
Hi @jdmarshall
With the changeover to the official Prometheus team, do you have any news regarding the integration of native histograms?

PR #829, opened 2026-08-27
Add option to keep OpenMetrics counter names stable without `_total`

krajorama:
TL;DR: add option to not break user's dashboards/arlerts when switching to OpenMetrics 1.0. At the price of demoting metric type of counters to "unknown" in metadata, but not adding `_total` suffix. Aligns with client_golang, forward compatible with OpenMetrics 2.0 which does not require suffix, just [encourages](https://prometheus.io/docs/specs/om/open_metrics_spec_2_0/#counter) it.

## Summary
- OpenMetrics requires Counter samples to carry a `_total` suffix. Today, switching a registry to `OPENMETRICS_CONTENT_TYPE` silently renames any Counter whose name doesn't already end in `_total` (e.g. `foo` becomes `foo_total` on the wire), which is a breaking change for consumers who scrape by name across the format switch.
- Adds an opt-in `counterWithoutTotalSuffix: 'unknown'` option to `Registry.setContentType()` and the `Registry` constructor. When set, such Counters keep their original name and are reported with type `unknown` instead of being renamed. Default remains `'append'` (today's behavior), so this is fully backwards compatible.
- As a side effect, removes a latent bug where `Registry.metrics()` permanently mutated the registered Counter's `.name` field in place; the naming decision is now computed locally per render.

## Test plan
- [x] `npm test` (596/596 passing)
- [x] `npm run lint`
- [x] `npx prettier --check .`
- [x] Added table-driven cases to `test/registerTest.js` covering the new option (valid/invalid mode, counter with/without pre-existing `_total` suffix)

Signed-off-by: György Krajcsovits <gyorgy.krajcsovits@grafana.com>

### DIFF
--- a/lib/registry.js
+++ b/lib/registry.js
@@ -26,7 +26,7 @@ class Registry {
 		return 'application/openmetrics-text; version=1.0.0; charset=utf-8';
 	}
 
-	constructor(regContentType = Registry.PROMETHEUS_CONTENT_TYPE) {
+	constructor(regContentType = Registry.PROMETHEUS_CONTENT_TYPE, options = {}) {
 		this._metrics = new Map();
 		this._collectors = [];
 		this._defaultLabels = {};
@@ -37,6 +37,9 @@ class Registry {
 			throw new TypeError(`Content type ${regContentType} is unsupported`);
 		}
 		this._contentType = regContentType;
+		this._counterWithoutTotalSuffix = validateCounterWithoutTotalSuffix(
+			options.counterWithoutTotalSuffix,
+		);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


lib/registry.js
```diff
@@ -26,7 +26,7 @@ class Registry {
 		return 'application/openmetrics-text; version=1.0.0; charset=utf-8';
 	}
 
-	constructor(regContentType = Registry.PROMETHEUS_CONTENT_TYPE) {
+	constructor(regContentType = Registry.PROMETHEUS_CONTENT_TYPE, options = {}) {
```
jdmarshall: This needs a typescript signature change.

lib/registry.js
```diff
@@ -65,25 +68,36 @@ class Registry {
...
-		const name = escapeString(metric.name);
+		const isOpenMetrics =
+			this.contentType === Registry.OPENMETRICS_CONTENT_TYPE;
+		const isCounter = metric.type === 'counter';
+		// A counter that isn't already named with a `_total` suffix would
+		// otherwise get silently renamed on the wire when exposed as
+		// OpenMetrics. Demoting it to `unknown` (opt-in via
+		// counterWithoutTotalSuffix: 'unknown') keeps the name stable across
+		// content types instead.
+		const demoteToUnknown =
```
jdmarshall: I've read this five times and I can't understand it.  The average developer can handle conditional branches 3 deep, and above average developers are expected to handle 4. This is 8 by my count.

lib/registry.js
```diff
@@ -65,25 +68,36 @@ class Registry {
...
+			isOpenMetrics &&
+			isCounter &&
+			!metric.name.endsWith('_total') &&
+			this._counterWithoutTotalSuffix === 'unknown';
+		const appendTotalSuffix = isOpenMetrics && isCounter && !demoteToUnknown;
+
+		const name = escapeString(
+			appendTotalSuffix ? standardizeCounterName(metric.name) : metric.name,
```
jdmarshall: 'standardizeCounterName' is misnamed. Its only purpose is to handle OpenMetrics renaming. 

I recommend you rename the method, make it a member method on the object (private? or leave exposed for testing?), and push a couple of these conditionals down into it, so that it handles some of the conditionals instead of them being handled here.  

Perhaps the method should take the metric itself, and look up the suffix settings.

krajorama:
If we get Protobuf that doesn't do the adding of suffixes thankfully via maybe #576 , then this can be closed as I'll have a workaround.

PR #830, opened 2026-08-27
Update benchmark baseline to latest release.

jdmarshall:
This enables some benchmarks for functions that did not exist in v15.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
 		"eslint-plugin-n": "^18.3.0",
 		"eslint-plugin-prettier": "^5.0.1",
 		"express": "^5.1.0",
-		"faceoff": "^1.3.0",
+		"faceoff": "^1.4.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -41,7 +41,7 @@
 		"eslint-plugin-n": "^17.20.0",
 		"eslint-plugin-prettier": "^5.0.1",
 		"express": "^5.1.0",
-		"faceoff": "^1.3.0",
```
jdmarshall, line 44: 1.3 has a math bug that reports regressions a hair under 5% (1.00/1.05 instead of 95%), which gets confusing due to rounding errors, that some tests show in the regression while others do not. This was fixed as adding a feature to report a specific range as 'slow' instead of defaulting to 5% (or 4.76%)

zbjornson (approved):

PR #835, opened 2026-08-28
fix: preserve zero-valued Counter exemplars

anisayakmitra-in:
## Summary

- preserve a Counter exemplar value of


lib/counter.js
```diff
@@ -93,7 +93,7 @@ class Counter extends Metric {
 		entry.exemplar ??= new Exemplar();
 		entry.exemplar.validateExemplarLabelSet(exemplarLabels);
 		entry.exemplar.labelSet = exemplarLabels;
-		entry.exemplar.value = value ? value : 1;
```
jdmarshall, line 96: See, this is why truthiness sucks. I've probably seen this bug a handful of times while working on other stuff and keep not getting around to it. Please file a bug to go with this PR.

jdmarshall (approved):

PR #839, opened 2026-09-01
Clear up remaining truthiness issues.

jdmarshall:
This issue has gotten great contributions from the community but a few have been overlooked, particularly in the tests.

Clearing up the rest so that code tools stop complaining.

This should be a slight performance improvement over a handful of code paths.

### DIFF
--- a/lib/metric.js
+++ b/lib/metric.js
@@ -37,10 +37,10 @@ class Metric {
 			defaults,
 			config,
 		);
-		if (!this.registers) {
-			// in case config.registers is `undefined`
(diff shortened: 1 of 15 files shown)


cjihrig (approved):

lib/metric.js
krajorama, line 60: Should this be also changed?

jdmarshall: I noticed that one. The use of nullish values in constructors shows up in the tests and so I suspect some people must be using it in their code. But passing 0 in could cause some chicanery.

PR #849, opened 2026-09-04
feat: add getClusterMetricsAsJSON to cluster registries (#470)

### LINKED ISSUE #470: Please consider adding a getClusterMetricsAsJSON funtion

amasen02:
## Summary

- Adds \getClusterMetricsAsJSON(aggregator?: string)\ to \AggregatorRegistry\ and \ClusterRegistry\ to expose aggregated cluster metrics as parsed JSON objects, mirroring \Registry#getMetricsAsJSON()\.
- Refactors \#gather()\ and collection coordination into \#aggregateMetrics(transform)\ so that both \clusterMetrics()\ and \getClusterMetricsAsJSON()\ share identical collection and worker lifecycle logic without duplicating code.
- Ensures \shutdown()\ continues to cleanly track and await in-flight aggregated metrics requests through completion before exiting.
- Updates TypeScript definitions in \index.d.ts\ for both \ClusterRegistry\ and \AggregatorRegistry\.
- Adds comprehensive unit test coverage in \	est/clusterTest.js\ verifying empty cluster state, multi-worker JSON aggregation ordering, and aggregator filtering (\sum\/\omit\), as well as compile-time type verification in \	est/typescript.ts\.
- Updates \README.md\ documentation with usage example.

Closes #470.

## Verification

- \
pm run lint\ (clean, 0 warnings/errors)
- \
pm run check-prettier\ (all files formatted)
- \
pm run compile-typescript\ (clean compilation)
- \
pm run test-unit -- --coverage\ (31 suites passed, 603 tests passed, 34 snapshots passed)

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -15,8 +15,8 @@
 'use strict';
 
 /**
- * Extends the Registry class with a `clusterMetrics` method that returns
- * aggregated metrics for all workers.
+ * Extends the Registry class with `clusterMetrics` and `getClusterMetricsAsJSON`
+ * methods that return aggregated metrics for all workers.
  *
  * In cluster workers, listens for and responds to requests for metrics by the
  * cluster master.
@@ -60,12 +60,11 @@ class AggregatorRegistry extends Registry {
 	}
 
 	/**
-	 * Gets aggregated metrics for all workers. The optional callback and
-	 * returned Promise resolve with the same value; either may be used.
-	 * @returns {Promise<string>} Promise that resolves with the aggregated
-	 *   metrics.
+	 * Executes an aggregated collection across all workers and transforms the resulting Registry.
+	 * @param {Function} transform - mapper function called with the aggregated Registry
+	 * @returns {Promise<any>}
 	 */
-	async clusterMetrics() {
+	async #aggregateMetrics(transform) {
 		const requestId = requestCtr++;
 		const orderedWorkers = [...workers.values()]
 			.filter(worker => worker.isConnected())
@@ -91,7 +90,7 @@ class AggregatorRegistry extends Registry {
 		const request = {
 			responseHandlers,
 			promise: waitFor(
-				this.#gather(requestId, metricSnapshot, responsePromises),
+				this.#gather(requestId, metricSnapshot, responsePromises, transform),
 				5_000,
 			),
 		};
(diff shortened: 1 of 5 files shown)


lib/cluster.js
```diff
@@ -128,15 +127,39 @@ class AggregatorRegistry extends Registry {
...
+	async #gather(requestId, historical, promises, transform) {
 		const responses = await Promise.all(promises);
 		const metrics = responses.flatMap(response => response.metrics);
-		return Registry.aggregate(
+		const registry = Registry.aggregate(
 			[historical, ...metrics],
 			this.contentType,
-		).metrics();
+		);
+		return transform(registry);
```
jdmarshall: One of the reasons I ended up contributing to Prometheus over OpenTelemetry is that it doesn't use inordinate levels of indirection for everything that make the code 1) slow as hell and 2) very very difficult to follow.

This indirection is unnecessary for a private function call. And a Rule of 3 violation. Just return the populated Registry and call the appropriate function.

jdmarshall (changes requested):

lib/cluster.js
amasen02: Thanks for the feedback @jdmarshall! Completely agree — updated in commit 4d4424b to remove the 	ransform parameter and indirection entirely. #gather now simply aggregates and returns the populated Registry directly.

lib/cluster.js
```diff
@@ -128,15 +127,32 @@ class AggregatorRegistry extends Registry {
...
+	}
+
+	/**
+	 * Gets aggregated metrics for all workers. The optional callback and
+	 * returned Promise resolve with the same value; either may be used.
+	 * @returns {Promise<string>} Promise that resolves with the aggregated
+	 *   metrics.
+	 */
+	async clusterMetrics() {
+		return this.#collect(registry => registry.metrics());
```
jdmarshall: You moved the Rule of Three violation one call up the tree, which is actually worse here because of the need for the thenable, which helps me understand how you were intending this before but it's still effectively the same problem.

```
  const registry = await this.#collect();
  return registry.metrics();
```

jdmarshall (approved):

test/clusterTest.js
```diff
@@ -285,7 +426,7 @@ describe.each([
 
 				await Promise.all([promise, shutdown]);
 
-				expect(results).toEqual([1, 2]);
+				expect(results.sort()).toEqual([1, 2]);
```
jdmarshall: This only proves that both fire and not in what order.  If this needs to be changed then it's not ensuring that pending work is being retired prior to shutdown.  

So something else needs to be asserted here instead. What we don't want is overlapping `aggregate` calls. So that's probably the thing to check instead.

jdmarshall (changes requested):

test/clusterTest.js
amasen02: Makes total sense. Replaced the promise-settlement ordering check with an explicit spy on Registry.aggregate in commit ad4ee59:
1. Asserts shutdown does not resolve prematurely while the request is in-flight.
2. Asserts Registry.aggregate is invoked and completes prior to shutdown resolving.
3. Asserts Registry.aggregate is called exactly once (verifying no overlapping or redundant calls).

All 31 test suites (603 tests), lint, prettier, and TypeScript compilation pass cleanly.

PR #850, opened 2026-09-04
Record cluster worker scrape failures

nightjard:
## Summary

Add an internal histogram named
`prom_client_cluster_worker_scrape_failures` to record failed cluster metric
collections.

Each observation contains the number of workers that failed to return metrics:

- For timeouts, it records the number of outstanding worker responses.
- For worker-reported errors, it records the known worker failure count.
- Failed collections continue to reject without returning partial metrics.
- New observations are exposed on the next successful `clusterMetrics()` call.

A histogram preserves information from consecutive failures instead of only
retaining the most recent value, as a gauge would.

## Testing

Added tests covering:

- Multiple workers timing out.
- Worker-reported collection errors.
- Histogram exposure after a successful recovery scrape.
- Prometheus and OpenMetrics output formats.

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -23,6 +23,7 @@
  */
 
 const { debuglog } = require('node:util');
+const Histogram = require('./histogram');
 const Registry = require('./registry');
 const { waitFor } = require('./util');
 
@@ -40,6 +41,13 @@ const ANNOUNCEMENT = '@prometheus-io/client:announcement';
 const GET_METRICS_REQ = '@prometheus-io/client:getMetricsReq';
 const GET_METRICS_RES = '@prometheus-io/client:getMetricsRes';
 const GOODBYE = '@prometheus-io/client:goodbye';
+
+const clusterWorkerScrapeFailures = new Histogram({
+	name: 'prom_client_cluster_worker_scrape_failures',
+	help: 'Number of workers that failed to return metrics during a failed cluster scrape.',
+	buckets: [1, 2, 4, 8, 16, 32],
+	registers: [],
+});
 
 let registries = [Registry.globalRegistry];
 let listenersAdded = false;
@@ -90,6 +98,7 @@ class AggregatorRegistry extends Registry {
 		const responsePromises = [this.#selfMetrics(), ...workerMetrics];
 		const request = {
 			responseHandlers,
+			workerFailures: 0,
 			promise: waitFor(
 				this.#gather(requestId, metricSnapshot, responsePromises),
 				5_000,
@@ -105,6 +114,16 @@ class AggregatorRegistry extends Registry {
 
 			return await request.promise;
 		} catch (err) {
+			let failedWorkers = request.workerFailures;
+
+			if (err.message === 'Timeout') {
+				failedWorkers += request.responseHandlers.size;
+			}
+
+			if (failedWorkers > 0) {
+				clusterWorkerScrapeFailures.observe(failedWorkers);
+			}
+
 			if (err.message === 'Timeout') {
 				throw new Error(
 					`Operation timed out. ${request.responseHandlers.size} outstanding responses.`,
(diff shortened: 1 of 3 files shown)


lib/cluster.js
```diff
@@ -105,6 +114,16 @@ class AggregatorRegistry extends Registry {
...
 			return await request.promise;
 		} catch (err) {
+			let failedWorkers = request.workerFailures;
+
+			if (err.message === 'Timeout') {
+				failedWorkers += request.responseHandlers.size;
+			}
+
+			if (failedWorkers > 0) {
+				clusterWorkerScrapeFailures.observe(failedWorkers);
```
jdmarshall, line 124: I think observing 0's when there are no workers is probably unlikely to happen but is also extra information about the class of failure we may be dealing with. As written this will never catch unclassified errors such as an NPE in the try block.

For instance, regressions in the Node API are not unknown. I work on a project that doesn't work on node 24.7-24.16, for instance, due to breaking changes in `child_process.fork()`

lib/cluster.js
```diff
@@ -118,8 +137,13 @@ class AggregatorRegistry extends Registry {
 	}
 
 	async #selfMetrics() {
+		const metrics = await Promise.all(
+			registries.map(r => r.getMetricsAsJSON()),
+		);
+		metrics.push([await clusterWorkerScrapeFailures.get()]);
```
jdmarshall, line 143: Are you sure you're not double-reporting this? That should be in the default registry.

jdmarshall (changes requested):
This same change should be made in worker.js as well, please, to retain feature parity.

lib/cluster.js
```diff
@@ -88,11 +98,14 @@ class AggregatorRegistry extends Registry {
 		);
 
 		const responsePromises = [this.#selfMetrics(), ...workerMetrics];
+		const timeoutError = new Error('Timeout');
```
jdmarshall, line 101: This can get a bit weird with stack traces. I think you're better off just letting waitFor create the error rather than passing one around.

lib/cluster.js
```diff
@@ -118,8 +146,22 @@ class AggregatorRegistry extends Registry {
 	}
 
 	async #selfMetrics() {
+		const metrics = await Promise.all(
+			registries.map(r => r.getMetricsAsJSON()),
+		);
+		// Custom registries may not contain the coordinator's internal metric.
+		if (
+			!registries.some(
```
jdmarshall, line 154: I think if it's being added to the registry you don't need this check.

lib/worker.js
```diff
@@ -78,17 +88,12 @@ class WorkerRegistry extends Registry {
 		);
 
 		if (orderedWorkers.length === 0) {
-			if (historicMetrics.length === 0) {
-				debug('No data found for requestId', requestId);
-				return '';
-			} else {
-				debug('No workers found for requestId', requestId);
-			}
+			debug('No workers found for requestId', requestId);
```
jdmarshall, line 91: Why did this change?

lib/worker.js
```diff
@@ -117,7 +126,22 @@ class WorkerRegistry extends Registry {
...
+			const timedOut = err === timeoutError;
+			let failedWorkers = request.workerFailures;
+
+			if (timedOut) {
+				failedWorkers += request.responseHandlers.size;
+			}
+
+			workerScrapeFailures.observe(failedWorkers);
+
+			// Sending can throw before request.promise is awaited.
```
jdmarshall, line 138: This is an AI hallucination?

You're rejecting and then squelching the error handler after rejection? I don't think we need any of this.

lib/worker.js
```diff
@@ -129,6 +153,23 @@ class WorkerRegistry extends Registry {
 		}
 	}
 
+	async #selfMetrics() {
```
jdmarshall, line 156: Also not needed. This thread is listening for the broadcast channel that sends the metrics. So again double reporting.

jdmarshall (changes requested):

lib/worker.js
```diff
@@ -153,23 +150,6 @@ class WorkerRegistry extends Registry {
 		}
 	}
 
-	async #selfMetrics() {
```
jdmarshall, line 156: I will say, I'm slowly coming around to the idea that this discrepancy between the two implementations will continue to present a problem going forward, and perhaps the implementation should change to not have the primary also register itself as a worker listener. But we don't need to keep going back and forth on this for this PR.

jdmarshall:
Unfortunately my time is tied up this evening but this looks pretty close. Thanks for the PR. Hopefully we can land it later this week.

jdmarshall (approved):

